PR #14, opened 2023-03-09
Remove org Components from Authentication section

martzoukos:
These three links point to nowhere when clicked:
image
and since they're repeated a bit lower in the menu, under "Organization Management", my guess is that they're left over there.

This PR removes them.


perkinsjr (approved):

perkinsjr:
Thanks had this on the todo.

PR #48, opened 2023-05-13
Move support items to navbar icons

crutchcorn:
This PR moves the following support sidebar items:

- Discord
- Twitter

To the top navbar in the form of icons.

It also removes the support email from the sidebar, which can be added to the footer in the near future if we like this idea.


perkinsjr (approved):
LGTM 

In the future we will have a new support flow on https://clerk.com/support so we can just direct there.

PR #127, opened 2023-06-24
fix: add import of createClerkClient to "Retrieve Full User" example

mattmazzola:
### Issue

The code snippet for the "Retrieve Full User" example was using `createClerkClient` without importing the function. This makes the snippet incomplete.

Up until that point in the guide the user is only familiar with these two import locations: "@clerk/remix" and "@clerk/remix/ssr.server", neither of which export `createClerkClient`.

**This means at this point in the guide the user is blocked and would not know how to import `createClerkClient`**

### Solution

I saw from this [upgrade documentation](https://clerk.com/docs/upgrade-guides/upgrade-remix-from-1.x-to-2.x) that appropriate  import is: `import { createClerkClient } from '@clerk/remix/api.server'`

I added imports to the snippet

### DIFF
--- a/src/pages/quickstarts/full-stack/remix.mdx
+++ b/src/pages/quickstarts/full-stack/remix.mdx
@@ -386,6 +386,10 @@ export const loader: LoaderFunction = async (args) => {
 #### Retrieve Full User
 
 ```tsx filename="routes/example.tsx" copy
+import { LoaderFunction, redirect } from "@remix-run/node";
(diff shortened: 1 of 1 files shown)


perkinsjr:
Hey Matt,

Appreciate all the PR's just as an FYI this documentation isn't for public consumption as we are still working on them and expect they should release in July. A lot of rough corners and missing details.

Thanks for the PR

James

PR #148, opened 2023-07-27
feat(docs): refactor repo to only contain docs content

BRKalow:
Updates the repository to only contain the documentation content. Going forward, we will be fetching the content from `clerk-marketing` to render our documentation pages.


alexisintech (approved):
send it....😸💖

PR #152, opened 2023-08-03
feat(scripts): adds lint:check-links script

BRKalow:
Adds a `check-links` script that will validate all absolute links against the the docs files in the `docs/` directory. Can be run with `npm run lint:check-links`.

I've also cleaned up our `package.json` a bit, and added a note about validating changes in our contributing guide.

Currently out of scope of this link checker are:
* External links
* Anchor link validation (e.g. validate that a header exists w/ the right slug)
* Links to paths outside of the `/docs` root path on clerk.com
* Links passed to components as props (cards)


pnpm-lock.yaml
BRKalow, line 1: Shifting this repo to `npm` for now to align with `clerkinc/javascript`. If we want to use a different package manager, I'd like us to make that decision as a team.

PR #161, opened 2023-08-09
feat: add reload user data section to useUser reference

BRKalow:
Makes some minor updates to the `useUser()` reference docs, and adds a section on reloading user data.

### DIFF
--- a/docs/references/react/use-user.mdx
+++ b/docs/references/react/use-user.mdx
@@ -3,15 +3,15 @@ title: useUser()
 description: The `useUser` hook is used to get the current user object and to update the user's data on the client side.
 ---
 
-import { Callout } from "nextra-theme-docs";
-import { Tables } from "@/components/Table";
-import { CodeBlockTabs } from "@/components/CodeBlockTabs";
-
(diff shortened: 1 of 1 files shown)


docs/references/react/use-user.mdx
```diff
@@ -103,6 +103,61 @@ export default function Home() {
 ```
 </CodeBlockTabs>
 
+### Reload user data
+
+If you need to retrieve the latest user data after updating the user elsewhere, use the `user.reload()` method.
+
+<CodeBlockTabs options={['Next.js', 'React']}>
```
alexisintech, line 110: Should this be in <Tabs> and then for the Next.js tab, have a <CodeBlockTabs> for pages vs app router?

BRKalow: So the snippet wouldn't actually change in either case, just the filename. I'm not positive it's worth it to create the extra layers there if the snippet is unchanged. What do you think?

alexisintech: oh cool ! agreed ^~^

PR #171, opened 2023-08-11
docs: add correct import source for App Router code examples

raptisj:
In the image below the correct import source is "next/navigation".

Screen Shot 2023-08-11 at 12 11 01 PM


alexisintech (approved):
Great catch! 😸💖

PR #174, opened 2023-08-14
feat: add allowlist to /references/backend

alexisintech:


BRKalow (commented):
One question about list formatting in a table, otherwise this looks good!

PR #180, opened 2023-08-15
docs: map remaining pages, remove some empty pages

BRKalow:
Maps remaining pages from our old system, and removes a number of empty pages. Some of these pages will likely end up being removed in the near future, but for now we'll continue serving the content until we have suitable replacements.


alexisintech (approved):
great catches!! 😸💖

PR #185, opened 2023-08-16
feat: add security to nav

alexisintech:
https://github.com/clerkinc/clerk-docs/assets/98043211/0918a1e0-421f-4393-92b4-4050bd4cfa42


docs/manifest.json
```diff
@@ -238,7 +237,7 @@
       ["Sign Out", "/custom-flows/sign-out"],
       ["User Impersonation", "/custom-flows/user-impersonation"],
       ["Invitations", "/custom-flows/invitations"],
-      ["Embeedded Magic Links", "/custom-flows/embedded-magic-links"],
+      ["Embedded Magic Links", "/custom-flows/embedded-magic-links"],
```
BRKalow, line 240: Good catch 👀

docs/manifest.json
```diff
@@ -344,6 +343,21 @@
...
+      "icon": "lock-closed",
+      "root": "security"
+    },
+    [
+      ["Overview", "/security/overview"],
+      ["Vulnerability disclosure policy", "/security/vulnerability-disclosure-policy"],
+      ["XSS leak protection", "/security/xss-leak-protection"],
+      ["CSRF protection", "/security/csrf-protection"],
+      ["Fixation protection", "/security/fixation-protection"],
+      ["Password protection and rules", "/security/password-protection"]
```
BRKalow, line 358:
```suggestion
      ["Password protection and rules", "/security/password-protection-and-rules"]
```

alexisintech: actually, I changed the file name bc there are links within our docs that point to /docs/security/password-protection 
did you want to me to change the file name back? and then hunt down those links? it's possible they were in sanity and may need redirects - I don't remember

BRKalow: Oh, missed that 🤦

We can rename as you have done here! I don't think there was a reason to change it so better to stick with the old one. Can you update the spreadsheet as well to make sure we don't redirect to the new (old) location?

BRKalow (approved):
✨

PR #193, opened 2023-08-17
chore: remove AP references

BRKalow:
This isn't technically launched yet, so changing the wording until it is.


alexisintech (commented):
Also, I remember briefly hearing something about it from Shehzad.
Should we remove the account portals page that's currently in the nav as well? Until the feature is more fleshed out. It's currently only one single page, and I do not think it exists on the current docs.

alexisintech (approved):
😸💖

PR #199, opened 2023-08-21
remove copy that mentions steps when written inside a <Steps /> component

alexisintech:
remove copy that mentions steps when written inside a `<Steps />` component, because the steps component already numbers each step
e.g. 'step 1' when inside a `<Steps />` component

reproduce on /docs/quickstarts/setup-clerk

Screenshot 2023-08-21 at 11 37 05


BRKalow (approved):
Good catch!

PR #200, opened 2023-08-21
Revert "fix: point previews at assembly branch for now (#198)"

BRKalow:
Hopefully this is the last time. Point back at production for previews.


alexisintech (approved):
hehehe

PR #226, opened 2023-08-26
Proper usage of the word `tamper`

yasamoka:
Fixed `docs/advanced-usage/manual-jwt.mdx` for proper usage of the word `tamper` (tamper *with it* rather than tamper *it*).


alexisintech (approved):
Good catch. Thank you for your contribution! 😸💖

PR #233, opened 2023-08-29
DOCS-282 Consolidate component references

alexisintech:
/components houses information on Clerk's prebuilt components

However, we also had /authentication/components/overview which housed redundant information on the prebuilt components. We also had [/organizations](/docs/organizations/create-component) which housed redundant information under a section titled "Components" and [/users](/docs/users/user-button) housed redundant information under a section titled "Prebuilt components".

This PR

- deletes these redundant pages, and instead, references to our /components documentation in the respective Overview pages. For example, /organizations no longer has a "Components" section with the `<UserButton />` and `<UserProfile />` pages. These pages already exist under /components/organization/* . Instead, /organizations/overview references these components and links out to their respective pages.
- migrates /users/overview and /organizations/overview from sanity to mdx, and also updates the content of these pages
- updates broken links found with our linter
- abstracts out types to be housed in dedicated pages under our /javascript/types/* path
- removes unnecessary imports (tech debt)

Both the /users and /organizations sections need a lot of work. But I think is a step in the right direction.

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -52,8 +52,8 @@
       ["Email & SMS options", "/authentication/configuration/email-options"],
       ["Allowlist / Blocklist", "/authentication/configuration/allowlist"],
       "# Prebuilt Components",
-      ["<SignUp/>", "/components/authentication/sign-up"],
       ["<SignIn/>", "/components/authentication/sign-in"],
+      ["<SignUp/>", "/components/authentication/sign-up"],
       "# Social Connections",
       ["Overview", "/authentication/social-connections/overview"],
       [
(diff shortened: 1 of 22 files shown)


docs/manifest.json
```diff
@@ -51,10 +51,6 @@
       ["Session options", "/authentication/configuration/session-options"],
       ["Email & SMS options", "/authentication/configuration/email-options"],
       ["Allowlist / Blocklist", "/authentication/configuration/allowlist"],
-      "# Prebuilt Components",
-      ["Overview", "/authentication/components/overview"],
-      ["<SignUp/>", "/authentication/components/sign-up"],
-      ["<SignIn/>", "/authentication/components/sign-in"],
```
BRKalow, line 57: I like the deduping here, but I'm wondering if we're reducing the usefulness of these sections by removing links to the component references 🤔  What do you think about continuing to have links to the reference pages?

BRKalow (approved):
legit

PR #245, opened 2023-09-01
update image extension; fix grammar on heading

alexisintech:
This PR

- fixes an incorrect image extension (the image was optimized using squoosh which changed the extension from png to jpg)
- fixes the grammar on the heading "Allow users to sign in passwordlessly"


astrotars (approved):
👍

PR #289, opened 2023-09-13
Migrate 'Same-origin' from Sanity to MDX

alexisintech:
/docs/backend-requests/making/same-origin

This PR

- adds links to respective references
- updates imports in code examples

https://github.com/clerkinc/clerk-docs/assets/98043211/17b78850-0fe5-442b-8ab1-23e62474f1c8


docs/backend-requests/making/same-origin.mdx
```diff
@@ -1,5 +1,97 @@
 ---
-sanity_slug: /docs/request-authentication/same-origin
+title: Same-origin requests
+description: Learn about various ways to make same-origin requests to the backend using the Clerk SDK.
```
BRKalow, line 3: Okay it looks like we use `Clerk SDK` quite a bit. It sounds kind of awkward to me though...what do you think is better?

* Clerk SDK
* Clerk's SDK
* Clerk SDKs
* Clerk's SDKs

docs/backend-requests/making/same-origin.mdx
```diff
@@ -1,5 +1,97 @@
...
+  };
+
+  return useSWR(url, fetcher);
+}
+```
+</CodeBlockTabs>
+
+When you re-focus a page or switch between tabs, SWR automatically revalidates data. This can be useful to immediately synchronize to the latest state in scenarios like stale mobile tabs, or laptops that went to sleep.
+
+## Tanstack Query (FKA React Query)
```
BRKalow, line 60:
```suggestion
## Tanstack Query (React Query)
```

docs/backend-requests/making/same-origin.mdx
```diff
@@ -1,5 +1,97 @@
...
+
+If you are using [Tanstack Query](https://tanstack.com/query/latest), formally known as React Query, in your application, you need to use a query function that throws errors. Since the native Fetch API does not, you can add it in.
```
BRKalow, line 62: Based on the other similar page:

```suggestion
If you are using [Tanstack Query](https://tanstack.com/query/v4/docs/react/overview), formally known as React Query, in your application, you need to use a query function that throws errors. Since the native Fetch API does not, you can add it in.
```

BRKalow: Should we align this with the other page in general? Let's also take Lennart's feedback into consideration here 🙏

docs/backend-requests/making/same-origin.mdx
```diff
@@ -1,5 +1,97 @@
...
+  if (!res.ok) {
+    throw new Error('Network response error')
+  }
+  return res.json()
+});
+```
+
+## Background Fetching
+
+For applications that are fetching content in the background, when a tab is no longer focused, you will need to include an Authorization header along with your request.
```
BRKalow, line 82:
```suggestion
For applications that are fetching content in the background, like when a tab is no longer focused, you will need to include an Authorization header along with your request.
```

docs/backend-requests/making/same-origin.mdx
alexisintech: We could cut SDK altogether and just say "Learn about various ways to make same-origin requests to the backend." or "Learn about various ways to make same-origin requests to the backend when using Clerk SDKs."

BRKalow: I like "Clerk SDKs" 👍

BRKalow (approved):

PR #290, opened 2023-09-13
Add note about provider de-optimizing static pages in nextjs apps

jescalan:
The clerk provider will de-optimize static/ISR pages due to its use of the `headers()` function. This was not clear to users and we have had several reports that this was an issue. The docs explicitly recommend wrapping the provider around the entire app, which is not exactly the right guidance if you have pages in your app that do not need to be behind auth and which you would like to deploy as static pages rather than serverless functions. This PR slightly adjusts the wording around `ClerkProvider` usage and adds a note about its de-optimizing behavior.


docs/quickstarts/nextjs.mdx
```diff
@@ -42,7 +42,13 @@ CLERK_SECRET_KEY={{secret}}
...
+
+<Callout type="info">
+  The `<ClerkProvider />` component needs to access [headers](https://nextjs.org/docs/app/api-reference/functions/headers) to authenticate correctly. This means anything wrapped by the provider will be opted into [dynamic rendering](https://nextjs.org/docs/app/building-your-application/rendering/server-components#dynamic-rendering) at request time. If you have static-optimized or ISR pages that you would prefer not to be opted into dynamic rendering, make sure they are <em>not</em> wrapped by `<ClerkProvider />`.
+  
+  This is easiest to accomplish by ensuring that `<ClerkProvider />` is added further down the tree to wrap page groups that explicitly need authentication instead of having the provider wrap your application root as recommended above. For example, if you have a set of landing/marketing pages, and an app that requires login that live within the same project, wrapping just the app layout group with the provider and keeping the marketing pages in a separate layout group would allow the marketing pages to continue to be static-optimized.
```
BRKalow, line 50: Tweaked the wording a bit:
```suggestion
  This is easiest to accomplish by ensuring that `<ClerkProvider />` is added further down the tree to wrap route groups that explicitly need authentication instead of having the provider wrap your application root as recommended above. For example, if you have a set of landing/marketing pages and a dashboard that requires login within the same project, create separate `(marketing)` and `(dashboard)` route groups. Rendering `<ClerkProvider />` in the `(dashboard)/layout.jsx` layout file will ensure that only those routes are opted into dynamic rendering, allowing the marketing routes to be statically optimized.
```

jescalan: Great call, will apply this now 🙏

alexisintech (changes requested):

alexisintech (approved):
😸💖

PR #293, opened 2023-09-14
Migrate 'JWT Templates' from Sanity to MDX

alexisintech:
/docs/backend-requests/making/jwt-templates

This PR

- adds links to respective references
- updates an image

https://github.com/clerkinc/clerk-docs/assets/98043211/c8892da1-cbe2-49cd-967d-9c6b6adf034b


docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -1,5 +1,273 @@
...
+{
+  "full_name": "{{user.last_name}} {{user.first_name}}"
+}
+```
+
+Interpolated shortcode substituted values will always be casted to strings. For example, if the user does not have a last name associated, the above full name value would be "null John".
+
+## Creating a template
+
+A template consists of the following (4) properties:
```
BRKalow, line 167:
```suggestion
A template consists of the following four properties:
```

BRKalow (approved):
Just one nit!

PR #298, opened 2023-09-16
Corrected type 'string' to boolean in UserButton props

99Yash:


royanger (commented):
This edit is good. docs/components/user/user-button.mdx needs the same edit.

royanger (approved):

jescalan (approved):

PR #313, opened 2023-09-19
Migrate saml-connections from Sanity to MDX

alexisintech:
/docs/custom-flows/saml-connections

This PR

- adds links to references

https://github.com/clerkinc/clerk-docs/assets/98043211/447f6af9-ac51-4133-a773-1df6d1d7090c


BrandonRomano (approved):
LGTM

PR #318, opened 2023-09-20
Migrate embedded-magic-links from Sanity to MDX

alexisintech:
/docs/custom-flows/embedded-magic-links

This PR

- adds links to references

https://github.com/clerkinc/clerk-docs/assets/98043211/a9de45ab-4c4e-411b-9c2a-54989ddbfef1


BrandonRomano (approved):
LGTM 🟢

PR #320, opened 2023-09-20
Migrate error-handling from Sanity to MDX

alexisintech:
/docs/custom-flows/error-handling

This PR

- add links to references

https://github.com/clerkinc/clerk-docs/assets/98043211/81ef6487-9a8a-424c-a098-d1301d24cc7b

### DIFF
--- a/docs/custom-flows/error-handling.mdx
+++ b/docs/custom-flows/error-handling.mdx
@@ -1,5 +1,100 @@
 ---
-sanity_slug: /docs/authentication/error-handling
+title: Error handling
+description: Provide your users with useful information about the errors being returned from Sign In and Sign Up requests.
 ---
 
-# This is a stub
+# Error handling
+
+When using the `signUp()` or `signIn()` functions, proper error handling can help you provide your users with useful feedback.
+
+## Example
+
+A user is attempting to sign up with your application, but they are attempting to use a password that has been found in an online data breach. This will return a 422 error code with the message: "Password has been found in an online data breach. For account safety, please use a different password."
+
+<CodeBlockTabs options={["React", "JavaScript"]}>
+```jsx
+import { useState } from "react";
+import { useSignUp } from "@clerk/clerk-react";
+
+function Example() {
+  const [error, setError] = useState(null);
+
+  const { signUp } = useSignUp();
+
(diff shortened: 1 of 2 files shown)


docs/custom-flows/error-handling.mdx
```diff
@@ -1,5 +1,100 @@
 ---
-sanity_slug: /docs/authentication/error-handling
+title: Error handling
```
BrandonRomano, line 2:
```suggestion
title: Error Handling
```

Do we use title case?  Unsure of what rules we follow here; but it appears that the [currently live page](https://clerk.com/docs/custom-flows/error-handling) uses title case.

docs/custom-flows/error-handling.mdx
```diff
@@ -1,5 +1,100 @@
...
+description: Provide your users with useful information about the errors being returned from Sign In and Sign Up requests.
```
BrandonRomano, line 3: I assume this is an intentional change; but just want to call it out to firm up my understanding.  It looks like we used to render this description on the page.  Assume this now is only used for the <title> meta tags?

CleanShot 2023-09-21 at 12 24 40@2x

docs/manifest.json
```diff
@@ -267,7 +267,7 @@
         "Multi-session applications",
         "/custom-flows/multi-session-applications"
       ],
-      ["Error Handling", "/custom-flows/error-handling"]
+      ["Error handling", "/custom-flows/error-handling"]
```
BrandonRomano, line 270: (if we go with title case as pointed out in the comment above, this probably should be adjusted here)

BrandonRomano (approved):
This looks good; I am unsure of the title casing rules that we adhere to.  Calling it out as it's something I noticed, but no need to adjust if this was intentional.

docs/custom-flows/error-handling.mdx
alexisintech: we use sentence-case!

docs/custom-flows/error-handling.mdx
alexisintech: Sanity descriptions are not used in MDX. They are used instead in our SEO descriptions.

PR #329, opened 2023-09-21
Migrate saml-at-clerk from Sanity to MDX

alexisintech:
/docs/authentication/saml-at-clerk

This PR

- adds saml-at-clerk to nav
- adds links to references
- adds images and optimized to webp using Squoosh

https://github.com/clerkinc/clerk-docs/assets/98043211/a44a989b-6a96-4b81-b3b9-ccc29c2c5d8f


docs/authentication/saml-at-clerk.mdx
```diff
@@ -1,3 +1,144 @@
...
+  width={3024}
+  height={1652}
+  src="/docs/images/authentication/enterprise-connection-settings-1.webp"
+  alt="The configuration settings for the Enterprise connection that was created."
+/>
+
+Scroll down to the **Service provider details** section. You will need to add these two fields to your IdP’s application:
+
+- SP Entity ID - This is a unique identifier for your SAML connection that your IdP application needs. Different providers use different terminology so consult our glossary table to locate the right term for your case.
+- ACS URL - This is your application’s URL that your IdP will redirect your users back to after they have authenticated in your IdP. Different providers use different terminology so consult our glossary table to locate the right term for your case.
```
royanger, line 52: "Different providers use different terminology so consult our glossary table to locate the right term for your case."
Maybe move these to a single Callout below the bullets to bring more attention?

There should be a link to the Glossary below.

docs/authentication/saml-at-clerk.mdx
```diff
@@ -1,3 +1,144 @@
...
+| IdP | Abbreviation for “Identity Provider” | | | |		
+| IdP Application | This is an application from the IdP that connects to | | | |		
+| Service Provider | This is the service that requests and receives the identity from the IdP. | | | |	
+| SP | Abbreviation for “Service Provider”. | | | |	
+
+## FAQ
+
+**I’ve enabled other strategies but they don’t work.**
+
+It is expected that once Enterprise SSO is enabled in an organization, there should be no other authentication methods applicable. This is in line with an organization’s intent to manage their users’ identity from one place.
```
royanger, line 120: "It is expected that once Enterprise SSO is enabled in an organization"

I believe this should read "It is expected that once Enterprise SSO is enabled in an application"

I'm not aware of a way to tie them to an Organization only.

royanger: I should add that with the upcoming auto-join work with Orgs that is tired to domains, this or other areas of the SAML doc might need updating. I'm not really sure how that work will affect this.

royanger (approved):

PR #334, opened 2023-09-22
Migrate hasura from Sanity to MDX

alexisintech:
/docs/integrations/databases/hasura

This PR

- adds links to references
- adds images and optimized them to webp's using Squoosh

https://github.com/clerkinc/clerk-docs/assets/98043211/e616c0be-d9f4-424a-abad-b6ffb709c80a


docs/integrations/databases/hasura.mdx
```diff
@@ -1,5 +1,166 @@
 ---
-sanity_slug: /docs/integration/hasura
+title: Hasura
+description: Learn how to integrate Clerk into your Hasura application.
 ---
 
-# This is a stub
+# Getting started
```
BrandonRomano, line 6: The H1 of these database pages should probably not just be "Getting Started", and it should include the name of the database in question.  I looked horizontally against other DB pages and they all seem to be starting with just "Getting started".

The reason for this is twofold:

- The first one is user readability perspective; we want to ensure the user is contextualized on the page so they know what they are looking at.  Right now the sidebar is the only thing that contextualizes the user on the page to know what DB they are looking at.  On mobile this sidebar is less prevalent hidden behind a click.
- Secondarily, there may be some SEO impacts if we have a bunch of pages with H1s that are just "Getting started"

I would look at how we approach our quick-start guides, as I think those H1s are well-formed. https://clerk.com/docs/quickstarts/nextjs

docs/integrations/databases/hasura.mdx
```diff
@@ -1,5 +1,166 @@
...
+
+The first step is to create a new Clerk application from your Clerk Dashboard if you haven’t done so already. You can choose whichever authentication strategy and social login providers you prefer. For more information, check out our [Set up your application](/docs/quickstarts/setup-clerk) guide.
+
+After your Clerk application has been created, navigate to the [JWT Templates](https://dashboard.clerk.com/last-active?path=jwt-templates) screen from the Clerk Dashboard. Click on the **New template** button to create a new template based on Hasura.
+
+<Images 
+  width={3024}
```
BrandonRomano, line 13: These images are too big and could be scaled down. I think the max size of container that the images are put in are less than 1000px (I think the breakpoint where the sidebar disappears is where the container is largest).

BrandonRomano (changes requested):
Few suggestions!

docs/integrations/databases/hasura.mdx
alexisintech: Yep, there's a ticket to fix this once these are all merged. Because some are already merged with these H1's, this can be done all at once.
Ticket: https://linear.app/clerk/issue/DOCS-342/fix-integrations-h1s

docs/integrations/databases/hasura.mdx
alexisintech: Sure, these are just the natural width and height on the image properties. I can cut them in half - but I'd appreciate to be educated on the why for this suggestion!

BrandonRomano: It's largely an optimization issue; we are loading images that are much larger than needed.  This leads to an increased file size that takes longer to load,  but we do not benefit from the increased image size because the browser has to scale the larger image down to fit in its container.

CleanShot 2023-09-22 at 10 48 02@2x

docs/integrations/databases/hasura.mdx
BrandonRomano: Cool cool; happy for this to be deferred, that won't hold up my approval then as long as we have a ticket to address!

docs/integrations/databases/hasura.mdx
alexisintech: Oh yes this part makes sense, but I thought Next/Image handled this 😭 there are so many images that need to be fixed 😭😭😭

alexisintech: but thank you, this makes a lot of sense I should have thought about this I learned this along time ago!! totally slipped my mind

BrandonRomano (approved):

PR #337, opened 2023-09-22
Migrate google-analytics from Sanity to MDX

alexisintech:
/docs/integrations/analytics/google-analytics

This PR

- adds links to references
- adds images and optimized them to webp's using Squoosh

https://github.com/clerkinc/clerk-docs/assets/98043211/1ea974b8-2187-4cdf-a728-0b3e9cefc6bf


jescalan (changes requested):
Both the gifs appear to be broken

Screenshot 2023-09-22 at 12 42 49 PM

jescalan (approved):

PR #339, opened 2023-09-22
Migrate deployments pages from Sanity to MDX

alexisintech:
This PR 

- migrates changing-domains from Sanity to MDX; updates the images; reorganizes the content; fixes some copy
- consolidated the import-users and export-users into one file; added reference links


docs/deployments/import-and-export-users.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+## Import users
+
+You can use the [Backend API](https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser) to create users and import them from your previous provider. 
+
+## Export users
+
+A list of users can be retrieved for use in any exports by calling the backend API found [here](https://clerk.com/docs/reference/backend-api/tag/Users#operation/GetUserList). 
+
+The [`getUserList()`](/docs/references/backend/user/get-user-list) method in the [Clerk Backend SDK](/docs/references/backend/overview) can also be used to retrieve a list of users.
```
jescalan, line 18: Do we need to add redirects for this change?

docs/deployments/changing-domains.mdx
```diff
@@ -1,5 +1,81 @@
...
-# This is a stub
+# Change domain or subdomain
+
+Learn how to change your production instance's domain or subdomain on Clerk.
+
+## Change domain
+
+Production domains can be changed in the [Clerk Dashboard](https://dashboard.clerk.com/) or via our [backend API](https://clerk.com/docs/reference/backend-api). Once you make the change to your domain, you will need to update the following:
+
+- Update the publishable key
```
royanger, line 14: This should be moved to its own section below the section for the new SSL Certs. The customer can't update this first -- the new key is generated after the DNS changes. It is just the domain base64 encoded.

This is the step that causes most of the support request relating to changing domains. Making it super clear with its own section can only help. 

Maybe something like this (I used a code block to the CallOut tags would show): 

## Update Your Publishable Key

After changing your domain, a new Publishable Key was automatically generated for your application. You will need to update your production instance with this new key and redeploy your application. 

```
<Callout>
Failing to update your Publishable will result in Clerk failing to load. 
</Callout>
```

docs/deployments/changing-domains.mdx
```diff
@@ -1,5 +1,81 @@
...
+- Update DNS records
+- Generate new SSL certificates
+- If using social connections, update redirect URLs for enabled social connections
```
royanger, line 17: Add: "You will also need to update the settings with your social connections, so the redirect URL they are using is correct."

docs/deployments/import-and-export-users.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+# Import and export users
+
+If you need some guidance or have 200k+ users, please reach out to [support](https://clerk.com/support).
+
+## Import users
+
+You can use the [Backend API](https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser) to create users and import them from your previous provider. 
+
+## Export users
```
royanger, line 14: As the top line is mentioning importing 200K+ (which is someone migrating to Clerk), it might be worth adding a line about requesting an export of users for migrating away. IE "If you wish to receive a CSV with all user data including their hashed passwords, please reach out to support."

docs/deployments/import-and-export-users.mdx
alexisintech: https://github.com/clerkinc/clerk-marketing/pull/774

royanger (approved):
This look to me.

PR #348, opened 2023-09-26
Added missing docs for chaining middleware

royanger:
The section from https://clerk.sanity.studio/desk/doc;2550b734-3774-4367-860e-a844a4edfe61 for chaining middleware wasn't ported to the new docs. Added it in. directly from Sanity.

### DIFF
--- a/docs/references/nextjs/auth-middleware.mdx
+++ b/docs/references/nextjs/auth-middleware.mdx
@@ -104,6 +104,36 @@ If you define an `afterAuth` function, it will run even if the request correspon
 
 ![Execution order of `beforeAuth`, `publicRoutes`, and `afterAuth`](/docs/images/nextjs/middleware-flow-chart.png)
 
+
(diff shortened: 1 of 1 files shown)


alexisintech:
This is the same exact example used [here](https://docs-preview-349.clerkpreview.com/docs)

Is there some way we can combine these two sections?

royanger:
> This is the same exact example used [here](https://docs-preview-349.clerkpreview.com/docs)

This link just goes to the docs root.

royanger:
> This is the same exact example used [here](https://docs-preview-349.clerkpreview.com/docs)

Go ahead and close this. I knew the chaining part used to be near the end of the page (after the image) I didn't see it there grepped the repo for 'chaining' and couple of other instead of just reading the page and realizing it was moved higher.

alexisintech:
Docs address chaining middleware! Closing this issue 😸💖

PR #350, opened 2023-09-27
feat: Replace invitationList with invitations in useOrganization

panteliselef:
We exposed invitation from useOrganization from our sdks which replaces the deprecated `invitationList`.

### DIFF
--- a/docs/references/react/use-organization.mdx
+++ b/docs/references/react/use-organization.mdx
@@ -41,7 +41,7 @@ To keep network usage to a minimum, we require developers to opt-in by specifyin
 
 ```jsx
 const { invitations } = useOrganization();
-// invitationList.data will never be populated
+// invitations.data will never be populated
 
 // Use default values to fetch invitations
 const { invitations } = useOrganization({
--- a/docs/references/react/use-organization.mdx
+++ b/docs/references/react/use-organization.mdx
@@ -12,11 +12,11 @@ const {
   isLoaded,
   organization,
   membership,
-  invitationList,
+  invitations,
   memberships,
-  membershipRequest,
+  membershipRequests,
   domains,
-} = useOrganization(useOrganizationParams);
+} = useOrganization();
 ```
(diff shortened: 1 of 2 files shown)


docs/references/react/use-organization.mdx
```diff
@@ -28,13 +28,29 @@ const {
 To keep network usage to a minimum but to also provide a smooth auto-updating experience to developers, we made the `invitationList` attribute available only when its respective parameter is present when calling the hook.
```
alexisintech, line 28: Do we still need this if `invitationList` is now deprecated?

docs/references/react/use-organization.mdx
```diff
@@ -28,13 +28,29 @@ const {
...
 
 ```jsx
-const { invitationList } = useOrganization();
-// invitationList is undefined.
+const { invitations } = useOrganization();
+// invitationList.data will never be populated
```
alexisintech, line 32: Is this comment even necessary to keep?

panteliselef: i believe yes, because it indicates that you need to opt-in in order to fetch data.

docs/references/react/use-organization.mdx
panteliselef: it is, but this needs to cover all, `invitations`, `domains` .... so i'll give it another look.

docs/references/react/use-organization.mdx
alexisintech: right, sorry I should have been more clear. should it be `invitations.data` instead of `invitationsList.data` because `invitationsList` is now deprecated in favour of `invitations`?

alexisintech (approved):

docs/references/react/use-organization.mdx
panteliselef: Oh i see, yeap i missed that one 😅

PR #364, opened 2023-09-29
Document CAA DNS records in Deployment Troubleshooting section

agis:
Presence of certain CAA DNS records on a root domain can conflict with instance deployment, if they prohibit both LetsEncrypt and Google Trust Services from issuing certificates for that domain.


panteliselef (approved):
💯

alexisintech (approved):
😸

PR #368, opened 2023-09-29
DOCS-342: fix h1's of /integrations pages for better SEO and clarity

alexisintech:
Many of the /integrations pages have `# Getting started` as their H1's

This PR updates the dedicated page's h1 and SEO title for better searchability on the site and better SEO.


BrandonRomano (approved):
Is there any way that we can update the template to incorporate the H1 on the page automatically from the page title?  This would be a better solution rather than always requiring the authors to duplicate the H1 here.

That being said, this is good for now.  Maybe create a ticket to consider doing what I suggested in the future!

PR #375, opened 2023-10-02
fix: broken create organization link in reference

BRKalow:
Update a reference to remove an extra character. Fixes a reported broken link.


alexisintech (approved):
🙈

PR #383, opened 2023-10-03
Update saml-at-clerk.mdx

microxgleek94:
a user wrote into support about the cost of SAML because it didn't specify if it was a monthly, yearly, or one time charge. this edit is to make the pricing clear on this doc page

plain tix for reference: https://app.plain.com/workspace/w_01GT53BQWV3DFW6ECTWZNQ1E9K/customer/c_01H8ABZ3WBT53FKNDSFX2R9B0K/?tab=helping&sort=oldest-first

### DIFF
--- a/docs/authentication/saml-at-clerk.mdx
+++ b/docs/authentication/saml-at-clerk.mdx
@@ -146,7 +146,7 @@ It’s going to be free during the Beta period.
 
 **How much will it cost after Beta?**
 
-It will cost $50 per connection for production Instances. Connections in development instances are and will be free, but capped to 5.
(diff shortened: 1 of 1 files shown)


royanger:
This should probably get removed from the Docs and replaced with a link to the pricing page. I can imagine we change pricing and don't think to look for stuff like this in the docs.

microxgleek94:
@royanger i agreed but saml doesn't have a price as yet on the pricing page and this is the only place where a price is mentioned - i can follow up in slack with the saml team about their thoughts about where this info should live

PR #385, opened 2023-10-04
Update nextjs.mdx : ClerkProvider Section

mpassad:
This feature needs to be clearly stated in the documentation that we have two options for the reading of the publishable key.
1. Automatic from the `.env` and directly as a prop on the `<ClerkProvider>` component


royanger:
I don't think this should be added to the quickstart. The paragraph about <ClerkProvider> links to the component specific page where this is covered along with all other props. From a quickstart perspective the code example works well.

alexisintech:
The use of this prop is talked about on the dedicated Clerk Provider page.
The quickstart guide is straightforward and has the information necessary for a quickstart.

PR #390, opened 2023-10-04
Show data fetch process from client component to protected api route when using nextjs

hsiez:
Might be useful for readers to see how a client components interacts with protected routes in nextjs


royanger (commented):
This example -- a client client component -- shouldn't be in the auth() page. This should be a DB query from an RSC and maybe also a mutation from a Server Action. 

This example could go in useAuth(), which is the hook which would be used in the client component. The example should probably use `getToken()` in case the fetch is to a non-Next backend where the Bearer token isn't passed automatically.

That said, @hsiez probably wait on a review or comment from @alexisintech  or @rachelnabors before revising. We might not want these examples in these files.

hsiez:
Thanks for explaining! Realized I overlooked some things and now this is feeling less useful. Was trying to highlight nextjs specific things when using clerk ie. "use client" and fetching from the client. This might be more suited for a stand alone example repo.

royanger:
> Thanks for explaining! Realized I overlooked some things and now this is feeling less useful. Was trying to highlight nextjs specific things when using clerk ie. "use client" and fetching from the client. This might be more suited for a stand alone example repo.

Its definitely useful -- just needs the right home!

alexisintech:
I agree with @royanger here.
We don't have an App Router example for the Next.js example on the [`useAuth()` dedicated page](https://clerk.com/docs/references/react/use-auth#use-auth).
Your contribution could serve as a helpful example there. 

If you would like to continue with this contribution, I would check out that page, look at the current Pages Router example for the Next.js example, and then repurpose that code for the App Router example that you have already drafted here.

This can be considered complete when 
- you have created an App Router example for the `useAuth()` dedicated page

hsiez:
@alexisintech @royanger Thanks for the reassurance! Moving this to the useAuth() page.

hsiez:
Notes for me:

What the example should explain:
With app router, components are server components by default. Trying to ping local api routes (/api) from a server component fails bc the server doesn't authenticate against itself. When a user signs in, the client authenticates against the server, so use client components ("use client") to get data from local api routes.

hsiez:
@alexisintech Split the code blocks so React is stand alone and the Nextjs section has a toggle for types of router. Happy to make any changes based on your preference. Thanks for your time!

alexisintech:
> @alexisintech Split the code blocks so React is stand alone and the Nextjs section has a toggle for types of router. Happy to make any changes based on your preference. Thanks for your time!

Amazing. Can you use `<Tabs />` for Next.js vs React (more info on how to do that [here](https://github.com/clerkinc/clerk-docs/blob/main/CONTRIBUTING.md#tabs-))
And then inside the Next.js tab, the `<CodeBlockTabs />` for App vs. Pages Router is perfect - I would simply switch the two so that App Router is the leading (first) option.

hsiez:
@alexisintech Great suggestions! The changes are up. Shall we trigger a deploy preview?

docs/references/react/use-auth.mdx
```diff
@@ -9,86 +9,90 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
-    const { getToken, isLoaded, isSignedIn } = useAuth();
-
-    // Fetch data from a Next.js API route handler
-    useEffect(() => {
-      if (!isLoaded || !isSignedIn) {
-        // You can handle the loading or signed state separately
-        return;
+<Tabs items={['Next.js', 'React']}>
+  <Tab>
+    ### Next.js
```
alexisintech, line 14: Not necessary, the user will be under the "Next.js" tab.

```suggestion
```

docs/references/react/use-auth.mdx
```diff
@@ -9,86 +9,90 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
+    <CodeBlockTabs options={["App Router", "Page Router"]}>
```
alexisintech, line 15:
```suggestion
    <CodeBlockTabs options={["App Router", "Pages Router"]}>
```

docs/references/react/use-auth.mdx
```diff
@@ -9,86 +9,90 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
+        };
+
+        return <div>...</div>;
+      }
+      ```
+    </CodeBlockTabs>
+  </Tab>
+
+  <Tab>
+    ### React
```
alexisintech, line 71:
```suggestion
```

alexisintech (changes requested):

docs/references/react/use-auth.mdx
```diff
@@ -9,51 +9,90 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
+        const fetchDataFromExternalResource = async () => {
+          const token = await getToken();
+          // fetch your data
+          return data;
+        }
+
+        return <div>...</div>;
+      }
+      ```
+    </CodeBlockTabs>
```
alexisintech, line 93: No need for a codeblocktabs here.. this code example is already in a `<Tab />`

```suggestion
      ```tsx filename="external-data-page.tsx"
      import { useAuth } from '@clerk/clerk-react';

      export default function ExternalDataPage() {
        const { getToken, isLoaded, isSignedIn } = useAuth();

        if (!isLoaded || !isSignedIn) {
          // You can handle the loading or signed state separately
          return null;
        }

        const fetchDataFromExternalResource = async () => {
          const token = await getToken();
          // fetch your data
          return data;
        }

        return <div>...</div>;
      }
      ```
```

alexisintech (changes requested):

hsiez:
@royanger @alexisintech I believe I address all the change requests. Haven't figured out why the revalidate failed after branch update. Might need a rekick. Also, I see y'all have moved forward with next app router examples: https://github.com/clerkinc/clerk-docs/pull/405

Feel free to close this pr if those examples serve the same purpose. If not, i think this is ready to go.

docs/references/react/use-auth.mdx
```diff
@@ -9,51 +9,86 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
+            headers: {Authentication: `Bearer ${token}`}
+          }))
+          .then(response => response.json())
+                .then(data => {
+                  setData(data.data)
+                });
+        }, []);
+
+        return <div>...</div>;
+      }
```
alexisintech, line 43: Your example is missing React imports! Also, I would abstract the logic out. And you actually can use async/await inside a client component - the component itself just cannot be async. So I've added back in the async/await to have the fetch await for the response from getToken().

```suggestion
"use client";

import * as React from "react";
import { useAuth } from "@clerk/nextjs";

export default function ExternalDataPage() {
  const [data, setData] = React.useState<object>({});
  const { getToken, isLoaded, isSignedIn } = useAuth();
  
  // In this case, /api/example is a Next.js Route Handler
  const fetchData = async () => {
    fetch("/api/example", {
      headers: { Authorization: `Bearer ${await getToken()}` },
    })
      .then((response) => response.json())
      .then((data) => {
        console.log("data", data);
        setData(data);
      });
    return;
  };

  React.useEffect(() => {
    if (!isLoaded || !isSignedIn) {
      // You can handle the loading or signed state separately
    }
    
    fetchData();
  }, []);

  return <div>{JSON.stringify(data, null, 2)}</div>;
}
```

alexisintech (changes requested):

docs/references/react/use-auth.mdx
hsiez: @alexisintech Schooled! Thanks for taking the time to teach me something. Much appreciated. Changes have been committed.

docs/references/react/use-auth.mdx
```diff
@@ -9,51 +9,90 @@ The `useAuth()` hook is a convenient way to access the current auth state. This
...
-  ```tsx filename="pages/index.tsx"
-  import { useAuth } from '@clerk/nextjs';
-
-  export default function ExternalDataPage() {
-    const { getToken, isLoaded, isSignedIn } = useAuth();
-
+<Tabs items={['Next.js', 'React']}>
+  <Tab>
+    <CodeBlockTabs options={["App Router", "Pages Router"]}>
+      ```tsx filename="/app/pageWithData/clientComponent.tsx"
```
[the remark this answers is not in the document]
hsiez: Done!

alexisintech (approved):
Thanks so much for going back and forth with us and working on this with us!! 😸💖

hsiez:
Likewise! I appreciate y'all time (:

PR #398, opened 2023-10-06
clean up cypress testing page; clean up custom session token page

alexisintech:


docs/testing/cypress.mdx
```diff
@@ -5,21 +5,21 @@ description: Use Cypress to write end-to-end tests with Clerk.
...
 
 <Callout type="info">
-  Since Cypress does not allow third-party cookies by default, you might be interested in the experimental Cookieless Dev mode.
+  Since Cypress does not allow third-party cookies by default, you might be interested in the experimental [Cookieless Dev](/docs/upgrade-guides/url-based-session-syncing) mode.
 </Callout>
 
 ## Cypress commands
 
-Easiest way to test with Clerk is to create sign in and sign out helper thats you can re-use in all of your tests.
+The easiest way to test with Clerk is to create sign-in and sign-out helpers that you can re-use in all of your tests.
```
JacobMGEvans, line 18: I would avoid words that imply difficulty level, due to its subjectiveness. This is an example of what I mean, and this doesnt need to be changed since it was previously worded this way, its more of a callout of tone and wording. 

```suggestion
A practical method for testing with Clerk involves creating sign-in and sign-out helpers, which can be effectively employed in all of your tests.
```

docs/testing/cypress.mdx
```diff
@@ -73,7 +73,9 @@ Cypress.Commands.add(`signIn`, () => {
 
 ## Simple tests
```
JacobMGEvans, line 74: I would suggest this be `Basic tests`

docs/testing/cypress.mdx
```diff
@@ -73,7 +73,9 @@ Cypress.Commands.add(`signIn`, () => {
...
 
-Now that some simple commands have been created, let's see how to write some basic tests. 
+Now that some simple commands have been created, here are some examples on how to write some simple tests.
```
JacobMGEvans, line 76:
```suggestion
Now that some commands have been created, here are some examples on how to write some basic tests.
```

docs/testing/cypress.mdx
```diff
@@ -73,7 +73,9 @@ Cypress.Commands.add(`signIn`, () => {
...
+
+This example will be testing a simple dashboard page that displays different content depending on whether the user is signed in or not.
```
JacobMGEvans, line 78:
```suggestion
This example will be testing a basic dashboard page that displays different content depending on whether the user is signed in or not.
```

JacobMGEvans (approved):
My comments don't need to be actioned on this time, but the language of "simple" should be replaced with more approachable and accessible vernacular. In most cases "simple" can just be removed or replaced with "basic"

PR #401, opened 2023-10-06
DOCS-434 remove links to outdated FAPI reference

alexisintech:
The FAPI reference is heavily outdated. This PR removes any links to it; we can add links to the reference when it is back up to date.

Link to Slack discussion: https://clerkinc.slack.com/archives/C01QFRQNHSN/p1696009744964919


JacobMGEvans (approved):
Might need to have a conversation around versioning of the docs, to Clerks major & minor, at least major versions. 

That does increase maintenance burden and complexity of the Docs repo... So there are tradeoffs.

PR #402, opened 2023-10-06
favor use of properties over attributes

alexisintech:
Many pages used either "Properties" or "Attributes" when referring the props on an object or class.
For consistency, it's best we choose one and abide by it. This PR updates headings that mention "Attributes" to now be "Properties".


JacobMGEvans (approved):
Lots of files, only a few changes lol

BRKalow:
@alexisintech Sorry I'm coming in late here..when talking about properties passed to components, I would say the canonical term within the React community is the shorthand "props," as opposed to "properties." When we're talking about our components, can we use "props"? ([ref](https://react.dev/learn/passing-props-to-a-component))

I would say when talking about objects, types, or non-component functions "properties" is reasonable.

PR #405, opened 2023-10-09
Added Next App Router examples to Email/OTP SignUp Custom Flow

royanger:
@alexisintech -- before I do a final edit/revision do you want the examples to be labelled as 'Next App Router'?. The file path/name, the `'use client'` and the use of 'next/navigation' are minor but specific to App Router. Or just call it Next.js?

### DIFF
--- a/docs/custom-flows/email-sms-otp.mdx
+++ b/docs/custom-flows/email-sms-otp.mdx
@@ -248,9 +248,12 @@ export default function Page() {
       });
 
       // Filter the returned array to find the 'phone_code' entry
-      const phoneCodeFactor = supportedFirstFactors.find(factor => {
-        return factor.strategy === 'phone_code'
-      })
+      const isPhoneCodeFactor = (
+        factor: SignInFirstFactor
+      ): factor is PhoneCodeFactor => {
+        return factor.strategy === "phone_code";
+      };
+      const phoneCodeFactor = supportedFirstFactors?.find(isPhoneCodeFactor);
 
       if (phoneCodeFactor) {
         // Grab the phoneNumberId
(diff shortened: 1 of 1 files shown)


royanger:
On line 32 I'm getting a TS error -> Property 'phoneNumberId' does not exist on type 'SignInFirstFactor'. typescript (2339). Not sure how to solve.

alexisintech:
> @alexisintech -- before I do a final edit/revision do you want the examples to be labelled as 'Next App Router'?. The file path/name, the `'use client'` and the use of 'next/navigation' are minor but specific to App Router. Or just call it Next.js?

If the App vs Pages router are exactly the same code, but will obviously have different filenames, then let's add "Pages Router" and just change the file name.
We should have both App and Pages Router examples, always. Even if its just a filename difference.

royanger:
> > @alexisintech -- before I do a final edit/revision do you want the examples to be labelled as 'Next App Router'?. The file path/name, the `'use client'` and the use of 'next/navigation' are minor but specific to App Router. Or just call it Next.js?
> 
> If the App vs Pages router are exactly the same code, but will obviously have different filenames, then let's add "Pages Router" and just change the file name. In my opinion, we should have both App and Pages Router examples, always. Even if its just a filename difference.

Its both a different filename and some code changes. I'll look at doing a Pages Router pass of the Custom Flows stuff after the App Router stuff. Still many pieces of Custom Flows to build and test for updates.

royanger:
> On line 32 I'm getting a TS error -> Property 'phoneNumberId' does not exist on type 'SignInFirstFactor'. typescript (2339). Not sure how to solve.

@BRKalow, when you have a moment, can you look at this? I'm not sure of the best way to solve this.

alexisintech (approved):
LGTM! 😸💖

PR #407, opened 2023-10-10
Added Reply and Retry to the Webhooks docs.

royanger:
Added docs on Retry and Replay. The image for for the replay-button will probably need a red box about the little pop out menu button.


public/images/integrations/webhooks/replay-options.png
alexisintech, line 1: I would use a full sized screenshot here, so that user's get the full context of where this is located and what this screen would look like for them.

public/images/integrations/webhooks/replay-button.png
alexisintech, line 1: I would use a full sized screenshot here, so that user's get the full context of where this is located and what this screen would look like for them.

alexisintech (changes requested):

alexisintech (approved):
awesome team work!! 😸💖

PR #408, opened 2023-10-10
Added hasImage for both user and org. Also added adminDeleteEnabled to orgs

royanger:


docs/references/javascript/organization/organization.mdx
```diff
@@ -16,8 +16,10 @@ This is a class that represents an organization. It's returned from the [`create
 | `slug` | `string \| null` | The organization slug. If supplied, it must be unique for the instance. |
 | `logoUrl` | `string \| null` | The full URL for the organization logo. (deprecated in favor of `imageUrl`) |
 | `imageUrl` | `string` | Holds the organization logo or default logo. |
+| `hasImage` | `boolean` | A getter boolean to check if the organization has a uploaded image. `false` if Clerk is displaying an avatar for the organization. |
```
alexisintech, line 19: I think this prop comes up other times in the docs, we should update those instances to use this description as well.

```suggestion
| `hasImage` | `boolean` | A getter boolean to check if the organization has a uploaded image. Returns `false` if Clerk is displaying an avatar for the organization. |
```

docs/references/javascript/user/user.mdx
```diff
@@ -23,6 +23,7 @@ The ClerkJS SDK provides some helper [methods](#methods) on the `User` object to
 | `fullName` | `string \| null` | The user's full name. |
 | `username` | `string \| null` | The user's username. |
 | `imageUrl` | `string` | Holds the users profile image or avatar. |
+| `hasImage` | `boolean` | A getter boolean to check if the user has uploaded an image or one was copied from OAuth. `false` if Clerk is displaying an avatar for the user. |
```
alexisintech, line 26:
```suggestion
| `hasImage` | `boolean` | A getter boolean to check if the user has uploaded an image or one was copied from OAuth. Returns `false` if Clerk is displaying an avatar for the user. |
```

alexisintech (changes requested):

docs/references/javascript/organization/organization.mdx
```diff
@@ -16,8 +16,10 @@ This is a class that represents an organization. It's returned from the [`create
 | `slug` | `string \| null` | The organization slug. If supplied, it must be unique for the instance. |
 | `logoUrl` | `string \| null` | The full URL for the organization logo. (deprecated in favor of `imageUrl`) |
 | `imageUrl` | `string` | Holds the organization logo or default logo. |
+| `hasImage` | `boolean` | A getter boolean to check if the organization has a uploaded image. Returns `false` if Clerk is displaying an avatar for the organization. |
```
alexisintech, line 19:
```suggestion
| `hasImage` | `boolean` | A getter boolean to check if the user has uploaded an image or one was copied from OAuth. Returns `false` if Clerk is displaying an avatar for the user. |
```

alexisintech (approved):
One minor thing for inconsistency. And then all good 😸💖

docs/references/javascript/organization/organization.mdx
royanger: This is not the user -- this is the organization. The 'an' is a correct edit, but this hasImage refers to the organization's image and not the user.

alexisintech: oh good point, I was sifting through too quickly - thank you for catching that!

PR #412, opened 2023-10-12
feat: JWT Template conditional expressions

agis:
The following patch documents the newly-added capability of the JWT Template language: conditional expressions using the `||` operator.

### DIFF
--- a/docs/backend-requests/making/jwt-templates.mdx
+++ b/docs/backend-requests/making/jwt-templates.mdx
@@ -18,7 +18,7 @@ At a glance, the process for creating and using a template is the following:
 - Create a JWT template from the [Clerk Dashboard](https://dashboard.clerk.com/last-active?path=jwt-templates)
 - From your frontend, generate tokens using that template
 
-The following sections will dive into these steps and explore what you can achieve with JWT Templates. 
+The following sections will dive into these steps and explore what you can achieve with JWT Templates.
 
 ## Template basics
 
@@ -150,7 +150,7 @@ Future extensions to the template language may provide more ways to query for sp
 
 ### Interpolation
 
-Shortcodes can be interpolated inside strings. 
+Shortcodes can be interpolated inside strings.
 
 For example, you could use interpolation to build the user's full name:
 
@@ -160,7 +160,74 @@ For example, you could use interpolation to build the user's full name:
 }
 ```
 
-Interpolated shortcode substituted values will always be casted to strings. For example, if the user does not have a last name associated, the above full name value would be "null John".
+Interpolated shortcodes will always result to string values. For example, if the user does not have a last name associated, the above full name value would be "null John".
+
+### Conditional expressions
+
+Conditional expressions use the `||` operator and can be used to substitute a default fallback value for shortcodes that would otherwise result in `null` or `false` values.
+
+The format of a conditional expression is the following:
+
+```json
+{
+  "key": "{{ <operand_1> || <operand_2> || <operand_n> }}"
+}
+```
+
+As an example, given the following template (comments added for clarity):
+
+```json
+{
+  "has_contact_info": "{{user.has_verified_email || user.has_verified_phone}}",
+
+  // fallback to a string value
+  "full_name": "{{user.full_name || 'Awesome User'}}",
+
+  // fallback to a number value
+  "age": "{{user.public_metadata.age || user.unsafe_metadata.age || 30 }}"
+}
+```
+
+and given a user that:
+
+- has verified their phone number
+- has not verified their email
+- has not provided their first or last name
+- does not have any public or unsafe metadata assigned
+
+
+the output of the template after substitution would be:
+
+```json
+{
+  "has_verified_contact_info": true,
+  "full_name": "Awesome User",
+  "age": 30
+}
+```
+
+The result of a conditional expression is that of the first operand that does not evaluate to `null` or `false` (also known as "falsy"). If all operands of the expression are falsy, the last operand is returned no matter its value.
+
+Conditional expressions can also be combined with interpolation:
+```json
+{
+  "greeting": "Awesome {{user.first_name || user.last_name || 'user'}}"
+}
+```
+
+Assuming a user that has no first or last name set in their profile, the above would output:
+
+```json
+{
+  "greeting": "Awesome user"
+}
+```
+
+Following are the rules that govern conditional expressions:
+
+- the result of an expression is either the first operand that is not falsy, or the last operand (no matter its value)
+- string literals should use single quotes (`'`)
+- only strings, booleans and numbers are permitted as literal (i.e. non-shortcodes) operands
 
 ## Creating a template
 
@@ -223,7 +290,7 @@ And given the following template (comments added here for clarity):
     "aud": "https://my-site.com",
     "version": 1,
     "foo": { "bar": [1,2,3] },
-    
+
     // dynamic values
     "user_id": "{{user.id}}",
     "avatar": "{{user.image_url}}",
@@ -257,7 +324,7 @@ Using the above template, generating a token for the aforementioned user would r
         }
      },
      "invalid_shortcode": "{{user.i_dont_exist}}",
-     
+
      // default claims, included automatically
     "iat": 1639398272,
     "nbf": 1639398220,


docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
 }
 ```
 
-Interpolated shortcode substituted values will always be casted to strings. For example, if the user does not have a last name associated, the above full name value would be "null John".
+Interpolated shortcodes will always result to string values. For example, if the user does not have a last name associated, the above full name value would be "null John".
+
+### Coalescing
```
georgepsarakis, line 165: ❓ I think `coalescing` is more closely related to nullish values: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing while we seem to be taking a slightly broader approach with falsy value categorization. Perhaps going with simply `conditional expression` is less ambiguous as a term (also taking into account possible extensions in the future, e.g. empty strings).

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+
+Shortcode coalescing is a conditional expression of the form:
+
+```json
+"{{ <operand_1> || <operand_2> || <operand_n> }}"
+```
+
+For example, given the following template:
+
+```json
```
georgepsarakis, line 175: 🔧 Consider splitting into separate examples, as the scope and purpose of each one might be clearer. Also, not sure how JS comments render in our docs site. Github will render them as invalid lines I believe.

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+{
+  "has_verified_contact_info": "{{user.has_verified_email || user.has_verified_phone}}",
+
+  // fallback to a string value
+  "full_name": "{{user.first_name || 'John'}} {{user.last_name || 'Doe' }}",
+
+  // fallback to a number value
+  "age": "{{user.public_metadata.age || user.unsafe_metadata.age || 30 }}"
```
georgepsarakis, line 183: 🔧 I think you may need to provide the input first so that someone correlates input-output. Here I'd imagine that both  `user.public_metadata.age` & `user.unsafe_metadata.age` are missing thus considered `null`?

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+```
+
+For example, given the following template:
+
+```json
+{
+  "has_verified_contact_info": "{{user.has_verified_email || user.has_verified_phone}}",
+
+  // fallback to a string value
+  "full_name": "{{user.first_name || 'John'}} {{user.last_name || 'Doe' }}",
```
georgepsarakis, line 180: 🔧 This is an interpolation case, right? You can consider outlining this better in a separate example, which clearly states that these two features can co-exist! 🎉

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+
+  // fallback to a number value
+  "age": "{{user.public_metadata.age || user.unsafe_metadata.age || 30 }}"
+}
+```
+
+The output after substitution (subject to the actual user data) would look like this:
```
georgepsarakis, line 187: ❓ We could perhaps describe the state more clearly here as well and avoid the parenthesized clarification?

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+
+Coalescing can be combined with interpolation:
+
+```json
+{
+  "title": "Awesome {{user.first_name || user.last_name || 'user'}}"
+}
+```
+
+Assuming a user that has no first or last name set in their profile, the above would output:
```
georgepsarakis, line 209: ❓ Do we document which shortcodes are nullable? Because for example `user.full_name` will return an empty string instead and the condition will fail to fall back to the literal value. Perhaps this could be surprising, not sure though.

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,65 @@ For example, you could use interpolation to build the user's full name:
...
+
+```json
+{
+  "title": "Awesome user"
+}
+```
+
+The rules that sum up how coalescing works are:
+
+- string operands should use single quotes (`'`)
```
georgepsarakis, line 219: ❓ Is this a requirement though? It's definitely good to advise that though.

georgepsarakis (commented):
👏

docs/backend-requests/making/jwt-templates.mdx
agis: Comments are rendered nicely - we already use them in the docs (see [example](https://clerk.com/docs/backend-requests/making/jwt-templates)).

docs/backend-requests/making/jwt-templates.mdx
agis: This is here by mistake. It should be just `first_name`, since I do mention that interpolation can be combined with coalescing a few lines below. Fixing this to be only `first_name`.

docs/backend-requests/making/jwt-templates.mdx
agis: Addressed https://github.com/clerkinc/clerk-docs/pull/412#discussion_r1360366045

docs/backend-requests/making/jwt-templates.mdx
agis: Right, this won't work as is. Changing this to a value that may indeed provide a `null` value (first/last name will always print a string).

agis: Changed this to use `full_name` instead, which can be `null`.

docs/backend-requests/making/jwt-templates.mdx
agis: it is a requirement due to https://github.com/clerkinc/clerk_go/pull/4441/files#r1356754898

georgepsarakis: Are you sure? My understanding is that quotes are used to enclose literal values that may contain characters that would otherwise confuse the parser. So for example:
- `user.public_metadata.missing_key || hello || world` will return `hello` instead of `hello || world`
- leading/trailing whitespace will be omitted.

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,74 @@ For example, you could use interpolation to build the user's full name:
...
+
+Shortcode coalescing is a conditional expression of the form:
+
+```json
+{
+  "key": "{{ <operand_1> || <operand_2> || <operand_n> }}"
+}
+```
+
+For example, given a user that:
```
georgepsarakis, line 175: 👍

docs/backend-requests/making/jwt-templates.mdx
agis: In your example (i.e. not using any quotes to surround the string), we would get `"hello"` indeed.

However, if for example you do:

```json
"{{ user.public_metadata.missing_key || \"hello\" }}"
```

you're going to get  `"{{ user.public_metadata.missing_key || \"hello\" }}"`, not  `"hello"`. Just tried this using one of the test cases.

docs/backend-requests/making/jwt-templates.mdx
SokratisVidros: Yes, great call out!

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,74 @@ For example, you could use interpolation to build the user's full name:
 }
 ```
 
-Interpolated shortcode substituted values will always be casted to strings. For example, if the user does not have a last name associated, the above full name value would be "null John".
+Interpolated shortcodes will always result to string values. For example, if the user does not have a last name associated, the above full name value would be "null John".
```
SokratisVidros, line 163: @agis Can we make it so that the strings are always empty instead of `null`? It's better UX for our customers to get "John" than "null John". Also notice that the value is trimmed for whitespaces.

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,74 @@ For example, you could use interpolation to build the user's full name:
...
+
+```json
+{
+  "title": "Awesome user"
+}
+```
+
+The rules that sum up how coalescing works are:
+
+- string operands should use single quotes (`'`)
```
SokratisVidros, line 228: Team is this about string operants or string literals?

agis: String literals, good catch, fixing.

docs/backend-requests/making/jwt-templates.mdx
agis: Note that this paragraph is pre-existing and refers to interpolation, not conditional expressions. It just appeared changed here because I re-formatted it due to max column limits.

That said, we can certainly revisit interpolation and change the way it works. However, do note that with conditional expressions now one can do:

```json
{{ user.last_name || '' }} {{ user.first_name || ''}}
```

docs/backend-requests/making/jwt-templates.mdx
agis: Changed to "conditional expressions". PTAL

docs/backend-requests/making/jwt-templates.mdx
```diff
@@ -160,7 +160,74 @@ For example, you could use interpolation to build the user's full name:
...
+
+```json
+{
+  "has_verified_contact_info": true,
+  "full_name": "Awesome User",
+  "age": 30
+}
+```
+
+In a conditional expression, operands are separated by `||`. The result of a conditional expression is the first operand that is not `null`/`false`. If all operands of the expression are `null`/`false`, the last operand is returned no matter its return value.
```
georgepsarakis, line 209: ❓ Could repeating the operator here seem redundant as we open with this explanation (line 167)?

georgepsarakis (approved):

PR #421, opened 2023-10-13
remove callout about mdx migration (migration is finished)

alexisintech:


JacobMGEvans (approved):
Always good to cleanup this sorta thing 🎉

PR #427, opened 2023-10-16
DOCS-448 add reference link for load() method

alexisintech:
[DOCS-448](https://linear.app/clerk/issue/DOCS-448/feedback-for-referencesjavascriptoverview) expresses confusion about the `load()` method called to instantiate ClerkJS.

This PR adds copy and a reference link to guide users to the `load()` reference docs, which describe the options that can be passed to this method.


JacobMGEvans (approved):
Additional context looks good!

PR #428, opened 2023-10-16
DOCS-453 add copy and code comment to clarify on the sign out process

alexisintech:
[DOCS-453](https://linear.app/clerk/issue/DOCS-453/%F0%9F%98%A2-feedback-for-custom-flowssign-out) expresses that the docs aren't clear on how to perform a sign out and redirect to a page after the sign out.

This PR adds copy and a code comment to clarify how the signOut method is used to sign out a user, and the NextJS router is used to redirect a user to a page after sign out is complete.


JacobMGEvans (approved):
The clarification seems good to me.

PR #432, opened 2023-10-17
docs: updating route handler file convention

DarrylBrooks97:
Updating the file syntax to match NextJS's [file convention](https://nextjs.org/docs/app/building-your-application/routing/route-handlers)

### DIFF
--- a/docs/references/nextjs/route-handlers.mdx
+++ b/docs/references/nextjs/route-handlers.mdx
@@ -10,7 +10,7 @@ Clerk provides helpers to allow you to protect your Route Handlers, fetch the cu
 
 If you aren't protecting your Route Handler using Clerk's middleware, you can check if the user is logged in.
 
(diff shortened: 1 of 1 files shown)


alexisintech (approved):
Good catch! 😸💖

royanger (approved):
This is good. I would suggest just using .ts and not even including .js anymore. That's become our general convention.

alexisintech:
> This is good. I would suggest just using .ts and not even including .js anymore. That's become our general convention.

Agreed! @DarrylBrooks97 could you please remove the js file extension, and simply have the .ts or .tsx extension?

This PR can be merged once that is done :)

DarrylBrooks97:
Perfect! Thank y'all, I'm loving the product btw it's good to see full TS support 🎉

alexisintech:
Thanks so much for contributing 😸💖

PR #436, opened 2023-10-17
DOCS-431 add password option for the strategy param of signincreateparams

alexisintech:
[DOCS-431](https://linear.app/clerk/issue/DOCS-431/%F0%9F%98%A2-feedback-for-referencesjavascriptsign-insign-in) states that `password` isn't listed an option for the `strategy` param of the `SignInCreateParams`.


JacobMGEvans (approved):
That diff for strategy is atrocious, I wish it did better differing large one lines like that 😅

PR #439, opened 2023-10-18
Update route-handlers.mdx

microxgleek94:
a user wrote in that the file names on this page were incorrect. instead of `app/route.ts` it should instead be `app/api/route.ts|js` as per the next docs here: https://nextjs.org/docs/app/building-your-application/routing/route-handlers#convention


royanger (changes requested):
Hey @microxgleek94. Please update to `ts filename="app/api/route.ts`. We're standardizing on only .ts in the docs.

microxgleek94 (commented):
removed "js" from file name

royanger (approved):
Looks good!

PR #443, opened 2023-10-22
fix: update broken link on fauna integration page

astrotars:
Fauna changed the URL to their user-roles page and the old one is 404ing.

Broken Link
https://docs.fauna.com/fauna/current/cookbook/security/user-roles

Working Link
https://docs.fauna.com/fauna/current/cookbook/advanced/security/user-roles


astrotars:
ref: https://clerkinc.slack.com/archives/C03PHCND9JP/p1697883213650269

alexisintech (approved):
Thanks so much for doing this for me while I was PTO 😸💖

PR #446, opened 2023-10-26
Ruby Docs: Fix initialize param

dugjason:
Client should be initialized with `api_key`, rather than `secret_key`. Check docs -> https://github.com/clerkinc/clerk-sdk-ruby/blob/main/lib/clerk.rb#L19


LekoArts (changes requested):
Thanks for the PR! Can you please also change the copy text above that still refers to the secret key?

Also, it should still work with the secret key, right? It gets used here if set: https://github.com/clerkinc/clerk-sdk-ruby/blob/main/lib/clerk.rb#L50

PR #449, opened 2023-10-27
Fix node-fetch@2 install command on firebase migration docs

tomaspozo:
The current command does not include the `npm install` / `npm i`.


LekoArts (approved):
Thanks!

PR #457, opened 2023-10-31
DOCS-494 Restructure navigation-related links

alexisintech:
We can define "navigation-related copy" as text where a user is given instructions to navigate to a certain page. In our docs, usually this is related to navigating to a page in the Clerk Dashboard. As it stands in the docs, there is usually a link to that page in the Clerk Dashboard and that link follows the format of `https://dashboard.clerk.com/last-active?path=<PATH>`. 
For example, on the [Sign-up options](https://clerk.com/docs/authentication/configuration/sign-up-options#set-up-multifactor-authentication) page, the copy reads "Although not available as an option in the initial new application screen, you can opt to turn on multifactor authentication (MFA) from the [configuration screen](https://dashboard.clerk.com/last-active?path=/user-authentication/multi-factor) in the Clerk Dashboard." 
The issue here is that if a user is not logged into their Clerk Dashboard, this will be a dead link, and the copy simply states "configuration screen". There are no clear instructions on how to get to that page.

This PR updates "navigation-related copy" to be more clear on the steps a user needs to take to navigate to a page/specific section of a page. Follows a structure that the new [styleguide](https://github.com/clerk/clerk-docs/blob/main/STYLEGUIDE.md#42-use-carets-to-help-users-navigate-menus) sets.
For example, following up on the previous example, this PR updates the copy to say "Although not available as an option in the initial new application screen, you can opt to turn on multifactor authentication (MFA) from the configuration screen. In the Clerk Dashboard, go to **User & Authentication > [Multifactor](https://dashboard.clerk.com/last-active?path=user-authentication/multi-factor)** and toggle on the multi-factor options you would like to enable. For more information on how to set up multifactor authentication, check out our [detailed guide](/docs/custom-flows/mfa#overview)."


rachelnabors (approved):
Yes, YESS YESS YES YES

jescalan (commented):
I don't think I have the bandwidth right to really review this in a detail-oriented manner because this is a huge PR, but if Rachel has reviewed and approves, I trust their judgement 🚀

PR #458, opened 2023-10-31
fix line highlights for nextjs quickstart

alexisintech:


rachelnabors (approved):
Thanks for catching these! Can we you also something to the README about these so if people update the code samples, they know what the numbers mean and update them, too?

PR #467, opened 2023-11-06
Fix copy-paste error in Session options docs

nickvigilante:
Deleted some extra text that didn't need to be there


alexisintech (approved):
Good catch! Thank you for contributing 😸💖

PR #470, opened 2023-11-07
Update Styleguide example for how to navigate users in the Clerk Dashboard

alexisintech:


rachelnabors:
Thanks for catching this! Should we add:
* Bold the navigation
* Make the last clickable dashboard link clickable?

I know you're still working on that one—worth updating the rule!

rachelnabors (approved):

PR #477, opened 2023-11-08
DOCS-583 Fix incorrect component reference

alexisintech:
Fix copy to reflect correct component that is referenced (SignIn instead of SignUp)


rachelnabors (approved):
shhiiiiip it

PR #493, opened 2023-11-15
Update organization-switcher.mdx

jshek-clerk:
Currently, the `hidePersonal` property for the `<OrganizationSwitcher />` indicates that setting this property to `false` will **HIDE** the personal workspace entry (rather than **SHOW** this entry). Changed the wording here to accurately show that setting the prop to `false` will show the personal workspace instead of hiding it.


royanger (approved):
Changes look good.

PR #494, opened 2023-11-15
DOCS-525 add explanations for audiences and clock skew

alexisintech:
[DOCS-525](https://linear.app/clerk/issue/DOCS-525/expand-on-authmiddleware-property-descriptions) states "User testing participants wondered about what audiences and clock skews are. This is an opportunity to expand on these topics or link to appropriate pages for all obscure properties."

🔎 [Old page](https://clerk.com/docs/references/nextjs/auth-middleware#options)
🔎 [New page](baking)

### DIFF
--- a/docs/references/nextjs/auth-middleware.mdx
+++ b/docs/references/nextjs/auth-middleware.mdx
@@ -121,7 +121,8 @@ The `authMiddleware()` method accepts an optional object. The following options 
 | `audience?` | `string \| string[]` | A string or list of [audiences](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3). If passed, it is checked against the `aud` claim in the token. |
 | `authorizedParties?` | `string[]` | An allowlist of origins to verify against, to protect your application from the subdomain cookie leaking attack.<br/>For example:<br/>`['http://localhost:3000', 'https://example.com']`<br/>For more information, refer to the [reference guide](/docs/references/nodejs/token-verification#validate-the-authorized-party-of-a-session-token). |
 | `beforeAuth?` | `function` | A function called before the authentication middleware is executed. If a redirect response is returned, the middleware will respect it and redirect the user. If `false` is returned, the auth middleware will not execute and the request will be handled as if the auth middleware was not present. |
-| `clockSkewInSeconds?` | `number` | Overrides the default allowed clock skew. Clock skew specifies the allowed time difference (in seconds) between the server and client clocks when validating a token. |
+| `clockSkewInSeconds? (deprecated)` | `number` | Deprecated in favor of `clockSkewInMs`. |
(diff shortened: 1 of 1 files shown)


docs/references/nextjs/auth-middleware.mdx
```diff
@@ -118,10 +118,10 @@ The `authMiddleware()` method accepts an optional object. The following options
...
+| `audience?` | `string \| string[]` | A string or list of [audiences](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3). If passed, it is checked against the `aud` claim in the token. |
 | `authorizedParties?` | `string[]` | An allowlist of origins to verify against, to protect your application from the subdomain cookie leaking attack.<br/>For example:<br/>`['http://localhost:3000', 'https://example.com']`<br/>For more information, refer to the [reference guide](/docs/references/nodejs/token-verification#validate-the-authorized-party-of-a-session-token). |
 | `beforeAuth?` | `function` | A function called before the authentication middleware is executed. If a redirect response is returned, the middleware will respect it and redirect the user. If `false` is returned, the auth middleware will not execute and the request will be handled as if the auth middleware was not present. |
-| `clockSkewInSeconds?` | `number` | Overrides the default allowed clock skew. |
+| `clockSkewInSeconds?` | `number` | Overrides the default allowed clock skew. Clock skew specifies the allowed time difference (in seconds) between the server and client clocks when validating a token. |
```
dimkl, line 124: The `clockSkewInSeconds` is used in the token verification used in `server` so it's more like:

> specifies the allowed time difference (in seconds) between the Clerk server (generating the token) and the customer application server clocks when validating a token. 

I don't suggest the above phrase, but I would suggest we use something else than `client` since it may result in misunderstanding that clock skew is related to the end-user browser clock.

dimkl (approved):

docs/references/nextjs/auth-middleware.mdx
anagstef, line 124: This is a deprecated property because the naming was wrong because it actually accepts the value in milliseconds. So, we have introduced the `clockSkewInMs` to have the correct name for this.

anagstef, line 124:
```suggestion
| `clockSkewInMs?` | `number` | Overrides the default allowed clock skew. Clock skew specifies the allowed time difference (in milliseconds) between the server and client clocks when validating a token. |
```

alexisintech: great point! I included it

anagstef (approved):
💯

docs/references/nextjs/auth-middleware.mdx
alexisintech: Updated the copy - let me know how you feel about it !

PR #501, opened 2023-11-20
Changed the word "replays" to "replace"

adnqpnd:
Corrected a word that could potentially confuse a reader.


royanger (approved):
Change look good. Thanks!

PR #502, opened 2023-11-20
Added Next, Remix, Rect, JS, Gatsby, to SDK manifest sections

royanger:
Added quickstarts to the SDK sub menus.

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -392,6 +392,7 @@
     [
       ["Overview", "/references/nextjs/overview"],
       "# Guides",
+      ["Next.js Quickstart", "/quickstarts/nextjs"],
       ["Read session and user data", "/references/nextjs/read-session-data"],
       ["Add custom sign up and sign in pages", "/references/nextjs/custom-signup-signin-pages"],
       ["Integrate Clerk into your app with tRPC", "/references/nextjs/trpc"],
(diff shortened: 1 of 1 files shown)


royanger:
@alexisintech -- the Remix and Gatsby links now go directly to the Quickstart page. I think this is because of a lack to Overview/other page above the Quickstart link.

alexisintech:
> @alexisintech -- the Remix and Gatsby links now go directly to the Quickstart page. I think this is because of a lack to Overview/other page above the Quickstart link.

Yeah so the logic is that the context is going open up on the first page in the array of pages (in the manifest.json), so if we place the /quickstart/nextjs as the first page in the array of pages, its going to open up whatever link that is.

alexisintech:
Closing as per this discussion https://clerkinc.slack.com/archives/D05JDTA9K6H/p1700508370517079
"hey I see your thought process here on [including the quickstarts in the SDK references,](https://github.com/clerk/clerk-docs/pull/502) but I'm not sure if they should exist in the side nav - we are currently trying to tackle the problem of links in the side nav switching contexts on a user (they would in the /references section, click on the nextjs quickstart, and are taken to the /quickstarts section - away from the nextjs /references context they were just in).

instead, if you want to include the quickstarts on their overview pages ? I know remix and gatsby don't currently have overview pages, so then that would require creating one for those contexts.

or you can create a ticket in linear to address the fact that quickstarts should be linked to in their respective SDK references, then we can try to think of a solution later (possibly after the site rearchitecture, or if more appropriate content gets created, etc)"
Roy has replied "Sounds good. Go ahead and close that PR"

PR #504, opened 2023-11-20
DOCS-624 clarify on the shape of user data (add reference to the User object)

alexisintech:
[DOCS-624](https://linear.app/clerk/issue/DOCS-624/%F0%9F%98%A2-feedback-for-referencesreactuse-user) is a user feedback ticket that states "what is the shape of user? More type hints please."

This PR adds references to the `User` object.

And thank you to Ali for leaving the feedback that incited these changes 💙

### DIFF
--- a/docs/references/react/use-user.mdx
+++ b/docs/references/react/use-user.mdx
@@ -5,7 +5,7 @@ description: The useUser() hook is used to get the current user object and to up
 
 # `useUser()`
 
-The `useUser()` hook is a convenient way to access the current user data where you need it. This hook provides the user data and helper methods to manage the current active session.
+The `useUser()` hook is a convenient way to access the current [`User`](/docs/references/javascript/user/user) data where you need it. This hook provides the user data and helper methods to manage the current active session.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


royanger:
Taking a quick look at the linear, the feedback asked for type hints as well. Maybe indicate that the `@clerk/types` package can be installed and then `import type { UserResource } from '@clerk/types';` can be used. This is the type for useUser() and useClerk().

alexisintech:
> Taking a quick look at the linear, the feedback asked for type hints as well. Maybe indicate that the `@clerk/types` package can be installed and then `import type { UserResource } from '@clerk/types';` can be used. This is the type for useUser() and useClerk().

I will reach out to them to let them know that! For now, I don't think its necessary for the page - it's not something we have on every page, so if we wanted to add it, we'd have to add to every single page. And then that's unnecessary repetition. so then the question will be posed, well where should the information about `@clerk/types` live?
I've created a ticket to discuss these ideas https://linear.app/clerk/issue/DOCS-676/types-section-on-javascript-reference-pages-need-to-be-consolidated

rachelnabors (approved):
Niiiiice way to work those reference docs!

PR #517, opened 2023-11-24
Component sidebar links open in new window

rachelnabors:
Makes the prebuilt/related component links in sidebars open in new windows.

**Warning** this might lead to a confusing user experience, since the "external link" icon implies these component pages live on another site as well as open in a new window. It might:
* make users reluctant to open those links
* disorient them when they see they are still on the same site

The best solution for this is a better IA. Please observe user sessions to determine if the effect is net positive. Roll back if net harm.

image


alexisintech:
I dont think is necessary as I’m going to remove those from the sidebar in the first place and rely on overview guides to link to components - I thought we had discussed that specifically with Linear ticket DOCS-668
Screenshot 2023-11-24 at 14 19 36

alexisintech (approved):
LGTM! for now...😸💖

rachelnabors:
Added the screenshot! Thanks @alexisintech !

image

PR #523, opened 2023-11-30
fix: Rename duration in seconds to lockout_expires_in_seconds in User…

yourtallness:
… Lockout error payload

This field has been renamed, so updating the reference to reflect that.

### DIFF
--- a/docs/custom-flows/error-handling.mdx
+++ b/docs/custom-flows/error-handling.mdx
@@ -105,7 +105,6 @@ const signIn = await client.signIn.create({
 
 If you have [Account Lockout](/docs/security/user-lock-guide) enabled on your instance and the user reaches the maximum allowed attempts for password or backup code verification, you will receive an HTTP status of 403 (Forbidden) and the following error payload:
 
-
 ```json
 {
   "errors": [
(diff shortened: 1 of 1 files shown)


docs/custom-flows/error-handling.mdx
```diff
@@ -114,11 +114,11 @@ If you have [Account Lockout](/docs/security/user-lock-guide) enabled on your in
...
+        "lockout_expires_in_seconds": 1800
       }
     }
   ]
 }
 ```
 
-The remaining duration will vary [depending on your configuration](/docs/security/customize-user-lockout). You can opt to render the error message returned as-is or format the supplied `duration_in_seconds` as per your liking.
\ No newline at end of file
+The remaining duration will vary [depending on your configuration](/docs/security/customize-user-lockout). You can opt to render the error message returned as-is or format the supplied `lockout_expires_in_seconds` value as per your liking in your own custom error message.
```
rachelnabors, line 124: I still don't know what this means... Can you give example code? Why is this something that is set to a numeric value and not pulled from the settings they set in the dashboard? Right now it looks like you're telling it a different number from whatever is set in the Clerk Dash

rachelnabors (changes requested):

docs/custom-flows/error-handling.mdx
yourtallness: I re-did this section to make it more evident that `lockout_expires_in_seconds` represents the live value of how many seconds remain until the lockout expires.

Also added a code sample of how to use this value to format a timestamp to one's liking.

rachelnabors: Nice, thank you!

rachelnabors (approved):

PR #529, opened 2023-12-07
Update metadata to add 8kb limit

nexus-codingg:
the Organization Metadata page (https://clerk.com/docs/organizations/metadata) shows a warning that the metadata is limited to 8kb but the User Metadata does not (https://clerk.com/docs/users/metadata) and all metadata is limited to 8kb


rachelnabors (approved):

rachelnabors:
Thanks so much! Merging!

PR #538, opened 2023-12-08
fix: rename whitelist/blacklist to allowlist/blocklist

zythosec:
Most places in the docs reference allowlist/blocklist except for these remaining items. This PR brings them into alignment with the rest of the docs.


rachelnabors:
Thank you so much for spotting this! I just checked, and we have a few other places. Will push out another PR shortly.

rachelnabors (approved):

PR #540, opened 2023-12-09
fix broken external links on migration pages

kylemac:


kylemac:
verified the links are fixed, so will power merge this

PR #553, opened 2023-12-12
react quickstart suggestions

alexisintech:
There were a few suggestions that I felt would be easier to just throw into code rather than explain in a code review.
Let me know your thoughts, and we can edit this up to merge into Roy's original PR.


docs/quickstarts/react.mdx
```diff
@@ -16,18 +15,15 @@ description: Learn how to use Clerk to quickly and easily add secure authenticat
...
       title: "Set up a Clerk application",
       link: "https://clerk.com/docs/quickstarts/setup-clerk"
-    },
-    {
-      title: "Create a React application using Vite",
-      link: "https://vitejs.dev/guide/#scaffolding-your-first-vite-project"
     }
   ]}
 >
+- Create a new React application using Vite
```
alexisintech, line 21: I've added this here and removed it from the "beforeYouStart" section, because it's actually Step 1 of this tutorial, and not done as a prerequisite.

docs/quickstarts/react.mdx
```diff
@@ -212,13 +209,15 @@ You're ready to [deploy your app to production](/docs/deployments/overview) and
 
 ## You're authenticated!
 
-Congratulations! Your app is now only available to authenticated users! But this is just the first step. If you would like to clone or investigate this application, please check out:
```
alexisintech, line 215: So I feel like this sentence is a bit misleading. This set up is not like Nextjs where the middleware is protecting every single page ootb. In this tutorial, we have set up some content that can be seen by un-authenticated users, so technically, the application is *not* only available to authenticated users - there's still content that is served to un-authenticated users.

royanger: The app behaviour was changed based on suggestions but I missed this.

alexisintech: and do the rest of the changes look fair to you?

royanger: For sure. I didn't have time earlier today.

royanger (approved):
Looking good.

docs/quickstarts/react.mdx
```diff
@@ -16,18 +15,15 @@ description: Learn how to use Clerk to quickly and easily add secure authenticat
...
     }
   ]}
 >
+- Create a new React application using Vite
 - Install `@clerk/clerk-react`
 - Set up your environment keys
 - Import the Publishable Key
 - Wrap your React app in `<ClerkProvider />`
-- Protecting your content
+- Use Clerk's control components to protect your content
```
rachelnabors, line 26: "Control components" is an old holdover of internal jargon. Recommend just "components"

PR #567, opened 2023-12-14
fix: remove md preview section in CONTRIBUTING.md

kevinzunigacuellar:
#### Description

This PR removes the tip to use the markdown preview tool from VSCode because it is incompatible with MDX files.


LekoArts:
Hi!

In theory you could add this to your `.vscode/settings.json` file (inside the repo):

```json
{
  "files.associations": {
    "*.mdx": "markdown"
  },
}
```

Then the preview would show the markdown parts (the MDX parts wouldn't work). I'll defer to @rachelnabors @alexisintech if we want to add these instructions or do as your PR suggests.

kevinzunigacuellar:
Thank you for the prompt response @LekoArts. Yes, we could extend the preview to work in `.mdx` files but I think it will defeat the purpose of having a preview view. 

A preview, in most cases, helps contributors check if the content (text and components) is working as intended. It also helps to identify any syntax, formatting or other build errors before deploying.

Besides from the stylistic features the markdown preview offers I don't think it adds that much value for .mdx files.

rachelnabors:
I agree with @kevinzunigacuellar. A better solution could be:

* add deploy previews for all PRs
* make it possible to run the docs site locally

CONTRIBUTING.md
```diff
@@ -50,12 +50,9 @@ The structure of the PR should be:
   - If there is an issue that this PR is meant to resolve, the titles will probably be the same.
 - **Description ("Leave a comment")**: Describe what the concern was and summarize how you solved it.
 
-## Previewing your changes
+## Previewing changes
```
rachelnabors, line 53:
```suggestion
## Preview your changes
```

rachelnabors (approved):

alexisintech (approved):
i like it, picasso 😸💖

PR #557, opened 2023-12-13
add guide for basic rbac

royanger:
Added: https://docs-preview-557.clerkpreview.com/docs/guides/basic-rbac
Edited: https://docs-preview-557.clerkpreview.com/docs/backend-requests/making/custom-session-token

PR includes:

- guide for basic, non-Organization RBAC
- two reference screenshots to be turned into form screenshots for the guide
- minor update to add global type information to sessions customization

I haven't linked this yet as I'm not sure where this should. Should it be the first actual 'guide' in a new Guides section? Somewhere else in the docs? It definitely should not live with the Organization docs.

PR #571, opened 2023-12-14
review suggestions for PR-557 (RBAC)

alexisintech:
suggestions for [PR-577](https://github.com/clerk/clerk-docs/pull/557) include
- updated screenshots (larger and optimized as webp)
- adds the guides section and this page to the manifest (sidenav)
  - requires [this PR in the marketing repo](https://github.com/clerk/clerk-marketing/pull/862) to be merged as it adds the necessary icon for this sidenav addition
- adds an explanation to why a user would need to build an RBAC (control which users access certain parts of their app, using Clerk's roles and publicMetadata features)
- removes mention of `src` directory - users that use the `src` directory will understand that this is their root folder

### DIFF
--- a/docs/guides/basic-rbac.mdx
+++ b/docs/guides/basic-rbac.mdx
@@ -11,7 +11,9 @@ This guide assumes that you are using Next.js App Router. The concepts here can 
 
 ## Configure the session token
 
-Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. `unsafeMetadata` can be read and updated in the browser, and as the name states, it is unsafe to store data in. `privateMetadata` can not be read in the browser. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard, making it the safest and best choice for this use case.
+Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
+
+`unsafeMetadata` can be read and updated in the browser, and as the name states, it is unsafe to store data in. `privateMetadata` can not be read in the browser. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard, making it the safest and best choice for this use case.
 
 To begin building a basic RBAC system, you need to make `publicMetadata` available to applications directly from the session token. This will make the token more convenient to work with and will reduce the need for additional network requests to load the data. 
 
--- a/docs/guides/basic-rbac.mdx
+++ b/docs/guides/basic-rbac.mdx
@@ -5,7 +5,7 @@ description: Learn how to leverage Clerk's publicMetadata to implement your own 
 
 # Implement basic Role Based Access Control (RBAC) with metadata 
 
-To control which users can access certain parts of your application, you can leverage Clerk's [roles](/docs/organizations/roles-permissions#roles) feature. Although Clerk offers roles as part of the feature set for [organizations](/docs/organizations/overview), enabling organizations is *not* required to use roles. This guide will cover setting up a basic Role Based Access Control (RBAC) system for products that don't need organizations.
+To control which users can access certain parts of your application, you can leverage Clerk's [roles](/docs/organizations/roles-permissions#roles) feature. Although Clerk offers a roles feature as part of the feature set for [organizations](/docs/organizations/overview), not every app implements organizations. This guide will cover a workaround for setting up a basic Role Based Access Control (RBAC) system for products that don't use Clerk's organizations or roles. 
 
 This guide assumes that you are using Next.js App Router. The concepts here can be adapted to Next.js Pages Router and Remix.
 
@@ -15,7 +15,7 @@ Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a t
 
 `unsafeMetadata` can be read and updated in the browser, and as the name states, it is unsafe to store data in. `privateMetadata` can not be read in the browser. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard, making it the safest and best choice for this use case.
 
-To begin building a basic RBAC system, you need to make `publicMetadata` available to applications directly from the session token. This will make the token more convenient to work with and will reduce the need for additional network requests to load the data. 
+To build a basic RBAC system, first you need to make `publicMetadata` available to the application directly from the session token. With `publicMetadata` attached directly to the user's session, a fetch and network request isn't required every time you need to access the data.
 
 In the Clerk Dashboard, navigate to [**Sessions**](https://dashboard.clerk.com/last-active?path=sessions). In the **Customize session token** section, select **Edit**. In the modal that opens, enter the following JSON. If you have already customized your session token, you may need to merge this with what you currently have.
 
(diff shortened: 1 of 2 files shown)


docs/guides/basic-rbac.mdx
```diff
@@ -1,44 +1,44 @@
...
-description: Learn how to leverage Clerk's publicMetadata to implement your own basic Role Based Access Controls
+title: Implement basic Role Based Access Control with metadata
+description: Learn how to leverage Clerk's publicMetadata to implement your own basic Role Based Access Controls.
 ---
 
-# Implement basic Role Based Access Control (RBAC) with Metadata
+# Implement basic Role Based Access Control (RBAC) with metadata 
 
-Clerk provides [User Metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. This guide will cover using Public Metadata to build a based Role Based Access Control (RBAC) system for your application. 
+To control which users can access certain parts of your application, you can leverage Clerk's [roles](/docs/organizations/roles-permissions#roles) feature. Although Clerk offers roles as part of the feature set for [organizations](/docs/organizations/overview), enabling organizations is *not* required to use roles. This guide will cover setting up a basic Role Based Access Control (RBAC) system for products that don't need organizations.
```
alexisintech, line 8: this can be re-written if it's not very accurate or clear, I kind of wrote this off the top of my head. but the guide definitely needs an explanation to why a user would need to build an RBAC (in order to control which users access certain parts of their app, using Clerk's roles and publicMetadata features)

docs/guides/basic-rbac.mdx
```diff
@@ -1,44 +1,44 @@
...
-<Callout type="info">
-Clerk offers Roles as part of the feature set for [Organizations](/docs/organizations/overview).  This guide will cover setting up a basic RBAC system for products that don't need Organizations.
-</Callout>
+This guide assumes that you are using Next.js App Router. The concepts here can be adapted to Next.js Pages Router and Remix.
 
-This guide assumes that you are using Next App Router. The concepts here can be adapted to Next Pages Router and Remix.
+## Configure the session token
 
-## Configuring the Session
+Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. `unsafeMetadata` can be read and updated in the browser, and as the name states, it is unsafe to store data in. `privateMetadata` can not be read in the browser. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard, making it the safest and best choice for this use case.
```
alexisintech, line 14: We could even expand on the *why* of why we need the metadata. something like:

```suggestion
Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.

`unsafeMetadata` can be read and updated in the browser, and as the name states, it is unsafe to store data in. `privateMetadata` can not be read in the browser. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard, making it the safest and best choice for this use case.
```

docs/guides/basic-rbac.mdx
```diff
@@ -1,44 +1,44 @@
...
 
-To build basic RBAC, you need to make `publicMetadata` available to applications directly from the token. This will make it more convenient to work with and avoid additional network requests to load the data. `publicMetadata` can be read by the browser and can only be updated server-side or in the Clerk Dashboard. `unsafeMetadata` can be read and updated in the browser, and `privateMetadata` can not be read in the browser. See [User metadata](/docs/users/metadata#user-metadata) for more information.
+To begin building a basic RBAC system, you need to make `publicMetadata` available to applications directly from the session token. This will make the token more convenient to work with and will reduce the need for additional network requests to load the data.
```
alexisintech, line 16: What do you think about this update?

```suggestion
To begin building a basic RBAC system, you need to make `publicMetadata` available to the application directly from the session token. By attaching the `publicMetadata` directly to the user's session, a fetch and network request isn't needed for every time you need to access the data.
```

docs/guides/basic-rbac.mdx
```diff
@@ -107,59 +115,65 @@ declare global {
...
 
 <Callout type="info">
-You can modify the behaviour of the helper function to meet your needs. Maybe it will return the roles that the user has, or you could create a `protectByRole()` and have that function handle the redirect.
+  You can modify the behavior of the helper function to meet your needs. Maybe it will return the roles that the user has, or you could create a `protectByRole()` and have that function handle the redirect.
 </Callout>
 
 ## Add admin tools to find users and add roles
 
-You can change roles in the Clerk Dashboard by editing the `publicMetadata` directly. You can also build a UI in your application to do this. You can leverage the `checkRole()` function you added along with server actions to build basic tools to find users and then add roles.
+You can leverage the `checkRole()` function you added along with server actions to build basic tools for finding users and adding roles.
```
alexisintech, line 161:
```suggestion
You can leverage the `checkRole()` function you added along with server actions to build basic tools for finding users and managing roles.
```

alexisintech:
I think we can also add something to the roles and perms docs, or the "managing roles" docs that mention something like
> If you are on a free plan and would like to use roles to control which users can access certain parts of your application, check out Clerk's guide on building a basic Role Based Access Control system.

docs/guides/basic-rbac.mdx
royanger: I don't mind the latest suggestion. I would like to keep the info about `publicMetadata` being available in the browser but only update-able server-side. Its a common questions when talking about this stuff.

docs/guides/basic-rbac.mdx
royanger: "Although Clerk offers roles as part of the feature set for [organizations](/docs/organizations/overview), enabling organizations is *not* required to use roles."

I think this is a little confusing 'There are roles in orgs...but you don't need orgs for roles.' I get what you mean, but I think clarifying 'virtual' roles as covered in this guide and the roles from Orgs a little more clearly would be perfect.

docs/guides/basic-rbac.mdx
royanger: Looks good.

royanger (approved):

docs/guides/basic-rbac.mdx
rachelnabors: I think this is confusing as well. We have a feature called "roles," which this technique doesn't share. Something like : "Although Clerk offers a roles feature as part of [organizations](/docs/organizations/overview), not every app implements organizations. This guide will cover a workaround for setting up a basic Role Based Access Control (RBAC) system for products that don't use Clerk's organizations or roles."

docs/guides/basic-rbac.mdx
rachelnabors: I'm going to go full International Tone:

```suggestion
To build a basic RBAC system, first you need to make `publicMetadata` available to the application directly from the session token. With `publicMetadata` attached directly to the user's session, a fetch and network request isn't required every time you need to access the data.
```

rachelnabors:
> I think we can also add something to the roles and perms docs, or the "managing roles" docs that mention something like
> 
> > If you are on a free plan and would like to use roles to control which users can access certain parts of your application, check out Clerk's guide on building a basic Role Based Access Control system.

We don't want to promote this way of implementing roles, so I would not make this link. cc @kylemac

rachelnabors (changes requested):
One comment. With that update, we're good to merge!

PR #581, opened 2023-12-24
Add callout to `OrganizationSwitcher`

thiskevinwang:
### What 

This calls out a critical piece (`OrganizationSwitcher`) in the docs — AFAICT, _"basically roles n permissions will not work unless you also use `OrganizationSwitcher`"_

### Why

I was attempting to use the new roles/permissions functionality, and was very confused by what the docs were laying out, and what I was seeing in my dev app. 

And I only chanced upon this via https://organizations-demo.clerk.app/.

### DIFF
--- a/docs/organizations/verify-user-permissions.mdx
+++ b/docs/organizations/verify-user-permissions.mdx
@@ -3,6 +3,12 @@ title: Verify the active user’s permissions
 description: A collection of utility functions and components in order to allow developers to perform authorization checks.
 ---
 
+<Callout type="info">
+  The following authorization checks are predicated on an Organization being selected. Without this, they will likely
+  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher)
+  is currently the only way to change this context.
+</Callout>
+
 # Verify the active user’s permissions
 
 Clerk provides two kinds of helpers for authorization:
--- a/docs/organizations/verify-user-permissions.mdx
+++ b/docs/organizations/verify-user-permissions.mdx
@@ -5,8 +5,7 @@ description: A collection of utility functions and components in order to allow 
 
 <Callout type="info">
   The following authorization checks are predicated on an Organization being selected. Without this, they will likely
-  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher)
-  is currently the only way to change this context.
+  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher) component and `useClerk().setActive({ organization: <orgId> })` method are two available options to handle this.
 </Callout>
 
 # Verify the active user’s permissions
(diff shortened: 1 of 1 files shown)


royanger:
Hey @thiskevinwang -- I haven't looked but is there no way to do this with `useOrganizationList()` and `useOrganization()` as well? A quick look makes it appear like they are used in the `<OrganizationSwitcher />` (https://github.com/clerk/javascript/blob/ca67ee79be638eb902695cd0f599f68201beff3d/packages/clerk-js/src/ui/components/OrganizationSwitcher/UserMembershipList.tsx#L1) but I didn't dig in too deep. If it can be done with the hooks too confirmation would be great for those building custom components.

docs/organizations/verify-user-permissions.mdx
```diff
@@ -3,6 +3,12 @@ title: Verify the active user’s permissions
 description: A collection of utility functions and components in order to allow developers to perform authorization checks.
 ---
 
+<Callout type="info">
+  The following authorization checks are predicated on an Organization being selected. Without this, they will likely
+  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher)
+  is currently the only way to change this context.
+</Callout>
```
thiskevinwang, line 10:
```suggestion
<Callout type="info">
  The following authorization checks are predicated on an Organization being selected. Without this, they will likely
  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher) component and `useClerk().setActive({ organization: <orgId> })` method are two available options to handle this.
</Callout>
```

docs/organizations/verify-user-permissions.mdx
```diff
@@ -3,6 +3,11 @@ title: Verify the active user’s permissions
 description: A collection of utility functions and components in order to allow developers to perform authorization checks.
 ---
 
+<Callout type="info">
+  The following authorization checks are predicated on an Organization being selected. Without this, they will likely
+  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher) component and `useClerk().setActive({ organization: <orgId> })` method are two available options to handle this.
+</Callout>
+
```
thiskevinwang, line 10:
```suggestion
<Callout type="info">
  The following authorization checks are predicated on an Organization being selected. Without this, they will likely
  always evaluate to false by default. Our built-in [`<OrganizationSwitcher/>`](docs/components/organization/organization-switcher) component is available as a styled option to handle this. For a headless option, our `useClerk()` and `useOrganizationList()` hooks return a `setActive({ organization: <orgId> })` method which can be used in custom components.
</Callout>

```

royanger (approved):
`setActive()` is available on `useOrganizationList()` and `useClerk()` but our docs cover using for organizations in the `useOrganizationList()` page so I changed that for consistency.

I added a few other changes to the language to try and be consistent with the Style Guide for docs.

alexisintech (approved):
Moved the callout to be placed under the H1, let me know how you feel about it and we can amend the commit if necessary :)

PR #598, opened 2024-01-08
"force organizations" guide suggestions

alexisintech:
A lengthy review/update to the Force Organizations guide that is better done in a separate PR rather than cumbersome github code review suggestions.

- reorganizes sections/renames sections to be more clear and overall editorial pass
- adds a section to organizations/overview that explains how to enable organizations.

### DIFF
--- a/docs/guides/force-organizations.mdx
+++ b/docs/guides/force-organizations.mdx
@@ -164,7 +164,7 @@ export const CustomOrgSwitcher = () => {
 #### Set an active organization based on the URL
 
 <Callout type="warning">
-  This approach still depends one the `setActive` method which only runs on the client. Due to this limitation, during SSR it is possible that `orgId` from `auth()` returns an incorrect value that does not match the organiaztion id in the URL.
+  This approach still depends on the `setActive` method, which only runs on the client. Due to this limitation, during SSR, it is possible that `orgId` from `auth()` returns an incorrect value that does not match the organization ID in the URL.
 </Callout>
 
 In some cases, you might want to set an active organization based on the URL. For example, if you have a URL like `/organizations/org_123`, you might want to set `org_123` as the active organization.
@@ -206,7 +206,7 @@ export function SyncActiveOrganization() {
 }
 ```
 
-We can now use the helper we created above inside a Layout that is directly inside the Dynamic Route with the `orgId` param.
+Now you can place the `SyncActiveOrganization` helper that you created above in a [layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#layouts) and use the layout for all pages that require an active organization. In the example below, the `SyncActiveOrganization` component will run on every page in the `/app/[orgId]` directory.
 
 ```tsx filename="app/[orgId]/layout.tsx"
 
@@ -222,8 +222,6 @@ export default function OrganizationLayout(props: PropsWithChildren) {
   );
 }
 ```
-
-
 
 ### Limit access to users with active organizations only
 
@@ -360,7 +358,7 @@ Now that you have configured your middleware to redirect users to the `/org-sele
 
 #### Limit access using an App Router layout
 
-In Next.js App Router applications, instead of using `authMiddleware`, you also have the option to use a [layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#layouts) to limit access to users with active organizations only.
+In Next.js App Router applications, instead of using `authMiddleware`, you also have the option to use a layout to limit access to users with active organizations only.
 
 In the example below, a component called `RequiredActiveOrgLayout` is created. This component will be used as a layout for all pages that require an active organization. If the user has an active organization, the route the user is visting will be rendered. If the user does not have an active organization, the organization selection page will be rendered.
 
--- a/docs/guides/force-organizations.mdx
+++ b/docs/guides/force-organizations.mdx
@@ -21,7 +21,7 @@ description: Learn how to hide personal accounts and force organizations in your
   - Limit access to users with active organizations only
 </TutorialHero>
 
-# Force organization-only accounts in your app
+# Hide personal accounts and force organizations
 
 In this guide, you will learn how to hide a user's "personal account" in order to appear as if they only have an organization account. This is useful for applications that are built for organizations only, such as B2B applications.
 
--- a/docs/guides/force-organizations.mdx
+++ b/docs/guides/force-organizations.mdx
@@ -1,86 +1,130 @@
 ---
-title: Avoid personal account and force organizations in your app
-description: Learn how to avoid personal account and force organizations in your app
+title: Hide personal accounts and force organizations
+description: Learn how to hide personal accounts and force organizations in your Clerk application.
 ---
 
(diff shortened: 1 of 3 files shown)


docs/guides/force-organizations.mdx
```diff
@@ -117,29 +161,37 @@ export const CustomSwitcher = () => {
 };
 
 ```
-    </Tab>
 
-    <Tab>
-```tsx filename="app/[orgId]/sync-active-organization.tsx"
+#### Set an active organization based on the URL
```
royanger, line 165: This is a great section @panteliselef and the code example is solid. However there is nothing instructing the developer on how to use this component or where it should be mounted.

I think this should be present as an alternative to CustomOrgSwitcher. Show how to mount this in 'pp/(with-active-organization)/layout.tsx' and explain its tole versus a component to let the user select an org. That, or cover its use in some way.

docs/guides/force-organizations.mdx
```diff
@@ -148,134 +200,116 @@ export function SyncActiveOrganization() {
...
+        </p>
+        <OrganizationList hidePersonal={true} />
+      </section>
+    )
+  }
+```
 
-export default function RequiredActiveOrgLayout(props: PropsWithChildren) {
-  const { orgId } = auth()
+In Next.js Pages Router applications, Clerk's [`authMiddleware`](/docs/references/nextjs/auth-middleware) helper can be used to limit the entire application to only user's with an active organization.
```
royanger, line 250: `authMiddleware` can play the same role for App and Pages Router. Its agnostic in that sense. We should either do App Router as the only option (especially as near the top the guide says "This guide will be written for Next.js applications using App Router"), or provide App and Pages Router examples for /org-selection

@panteliselef -- I would contend this should all be switched to `clerkMiddleware` and we merge this next week when v5 releases. Thoughts?

docs/guides/force-organizations.mdx
```diff
@@ -148,134 +200,116 @@ export function SyncActiveOrganization() {
...
 
-### Limit access for users with active organizations only
+In Next.js App Router applications, you can use a [layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#layouts) to limit the entire application to only users with active organizations.
 
-<Callout type="info">
-  **How users can end up in this state ?**
-  - User just created an account, not by accepting an invitation, and hasn’t created an organization yet
-  - User has joined or created multiple organizations and leaves or deletes the active organization. 
-</Callout>
+In the example below, a component called `RequiredActiveOrgLayout` is created. This component will be used as a layout for all pages that require an active organization. If the user has an active organization, the children will be rendered. If the user does not have an active organization, the organization selection page will be rendered.
```
royanger, line 218:
```suggestion
In the example below, a component called `RequiredActiveOrgLayout` is created. This component will be used as a layout for all pages. If the user has an active organization, the route the user is visting will be rendered. If the user does not have an active organization, the organization selection page will be rendered.
```

docs/guides/force-organizations.mdx
```diff
@@ -148,134 +200,116 @@ export function SyncActiveOrganization() {
...
 
-Based on your usecase you can either decide to limit users in the entire app or a specific part of it. For example a B2B application might need to limit users with active organization for the entire app, whereas a B2B2C application would only limit the `/dashboard` path to users with active organization.
+```tsx filename="app/(with-active-organization)/layout.tsx"
```
royanger, line 220: As this section is for 'the whole application' this should probably be modified to be in the root layout. That or something more like "app/(application)/layout.tsx"

Also see comment below.

docs/guides/force-organizations.mdx
```diff
@@ -148,134 +200,116 @@ export function SyncActiveOrganization() {
...
-});
-```
+    ```
+  </Tab>
+</Tabs>
 
-```tsx filename="pages/org-selection.tsx"
-import { OrganizationList } from "@clerk/nextjs"
-import { useRouter } from "next/router";
+#### Limit parts of an application to only users with active organizations
```
royanger, line 307: There is really limited different between 'whole application' and 'part of an application'

Whole application:
- check in app/layout.tsx
- or check in middleware
```
if(auth.userId && !auth.orgId && req.nextUrl.pathname !== "/org-selection"){
      const searchParams = new URLSearchParams({
        redirectUrl: req.url
      })
 
      const orgSelection = new URL(`/org-selection?${searchParams.toString()}`, req.url);
 
      return NextResponse.redirect(orgSelection)
```


Part of application
- check in app/dashboard/layout.tsx
- or check in middleware
```
if(auth.userId && !auth.orgId && <check if path is /dashboard> && req.nextUrl.pathname !== "/dashboard/org-selection"){
      const searchParams = new URLSearchParams({
        redirectUrl: req.url
      })
 
      const orgSelection = new URL(`/dashboard/org-selection?${searchParams.toString()}`, req.url);
 
      return NextResponse.redirect(orgSelection)
```

The guide could just cover these two options in one 'protect your application' general section.

royanger (changes requested):

docs/guides/force-organizations.mdx
panteliselef: There are a lot of combos that can be used. The reasoning behind not having an snipper with middleware for App Router is that I see people wanted to do "page" or "layout" protection. For sure going with a middleware will be more performant. 

In a slack thread we discussed that we wanted to showcase a layout protection for app router and authMiddleware for pages router.

Regarding `clerkMiddleware` we can come back an update the guide

docs/guides/force-organizations.mdx
alexisintech: I forgot to mention this as a comment last night when I posted this PR! Thank you Roy. And I am giving this a +1 that we should show where this `SyncActiveOrganization` component can be used/mounted.

docs/guides/force-organizations.mdx
alexisintech: I also thought about this and believe that because it can be used in both router-types, that we should showcase both. So we can change the text to be something like this:

```suggestion
Clerk's [`authMiddleware`](/docs/references/nextjs/auth-middleware) helper can be used to limit the entire application to only user's with an active organization. It can be used in both App Router and Pages Router applications.

... (show both App Router and Pages Router examples)

In Next.js App Router applications, instead of using `authMiddleware`, you also have the option to use a [layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#layouts) to limit the entire application to only users with active organizations.
```

alexisintech: @panteliselef would you be able to throw together the App Router example for using `authMiddleware` ?

docs/guides/force-organizations.mdx
alexisintech: I've updated the organization to better reflect this information. Let me know what you think!

docs/guides/force-organizations.mdx
royanger: This feels good to me.

JacobMGEvans (approved):

royanger (approved):

PR #599, opened 2024-01-08
DOCS-5143 fix misspelling of endpoint

alexisintech:
[DOCS-5143](https://linear.app/clerk/issue/DOCS-5143/feedback-for-userssync-datasync-clerk-data-to-your-backend-with) points out an inconsistency of the endpoint being `/api/webhook` but is listed as `/api/webhooks` in one of the endpoints references.

And thank you to pierreouannes for leaving the feedback that incited these changes 💙


royanger:
Changed to /webhooks as that makes more sense.

royanger (approved):

PR #601, opened 2024-01-08
fix code snippet

alexisintech:
`apiRoutes` and `publicRoutes` both had incorrect formatting on text that should be code snippets


jescalan (approved):

royanger:
I noticed this with another PR -- are you using a linter for Docs or a prettier config for docs? There are many formatting changes. Just wondering if this is something I should be adopting as well.

PR #628, opened 2024-01-12
chore: Add Bun.js to add @clerk package to different projects.

isiagi:
## Add Bun.js to add @clerk package to different projects.
For example: bun add @clerk/remix

## Why add Bun Js

Bun's Node.js compatible package manager is significantly faster than npm, yarn, and pnpm.

### DIFF
--- a/docs/quickstarts/expo.mdx
+++ b/docs/quickstarts/expo.mdx
@@ -17,7 +17,7 @@ Learn how to use Clerk to quickly and easily add secure authentication and user 
 
 Once you have an Expo application ready, you need to install Clerk's Expo SDK.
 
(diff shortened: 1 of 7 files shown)


isiagi:
@jescalan, please review this PR.

LekoArts:
Hi, thanks for the PR!

Right now across our documentation we use npm, yarn, and pnpm. These are widely used and relevant package managers whereas bun is relatively new and niche. To keep things consistent we'd need to update all our docs to add this new command which doesn't seem a good use of one's time right now.

We can revisit adding bun to the install commands some time later when it is more mature.

Thanks for the PR though!

PR #632, opened 2024-01-12
Add localization object changes

anagstef:


JacobMGEvans (changes requested):
We should reevaluate how we are presenting this to the user. Probably more as a technical reference of the JSON file or something more copy-able and/or easier to reference... Any suggestions are welcome.

LekoArts:
I agree with Jacob 👍 I'll make some changes to the PR

LekoArts:
I think this here is better: https://docs-preview-632.clerkpreview.com/docs/upgrade-guides/upgrading-from-v4-to-v5#clerk-localizations

anagstef:
I agree it was in a premature state. Sorry for that!

@LekoArts thank you, Lennart! Great job! I will keep this in a draft state for a little bit, because maybe I'll need to make some last changes on the list.

LekoArts (approved):
I'll go ahead and merge this to make things easier with the migration guide. If we need to add new stuff, please open a new PR or add things directly 👍

PR #640, opened 2024-01-16
Update headings for JS SDK docs

alexisintech:
Updates headings for JS SDK to be more clearly tied to each method.
For example:
```
## create()

### create() params

### create() returns
```

instead of 

```
## create()

## params

## returns
```

Also adds missing information, such as return information for methods


JacobMGEvans (approved):
A lot of changes but seem to be mostly find and replace! 

Looks great!

PR #641, opened 2024-01-16
Add more @clerk/localizations value changes on the upgrade guide

anagstef:
Add some localization values changes from UserProfile


LekoArts (approved):
Thanks!

PR #644, opened 2024-01-16
backend sdk core2 updates

alexisintech:
- introduces the "types" context to the sidenav
  - introduces new `PaginatedResourceResponse` page
  - introduces the Backend `User` object

for each method on the backend sub-api's:
- introduces a function expression to show the typing
- introduces an example section that shows a specific example of how to call the method to the backend sub-api, and shows what a response would look like
- adds references to the appropriate objects that are in return responses (e.g. Session object links to [JS Session object reference](https://clerk.com/docs/references/javascript/session#session), User object links to the [new Backend User object page](https://docs-preview-644.clerkpreview.com/docs/references/backend/types/backend-user), etc)
- adds a section "BAPI endpoint" that links the user to the BAPI docs for that specific method's endpoint

> note that the lint checker will fail because the base branch (release/v5) is missing a doc for the new clerk-middleware


JacobMGEvans (approved):

docs/references/backend/types/clerk-backend-api-response.mdx
```diff
@@ -0,0 +1,38 @@
...
+type ClerkBackendApiResponse<T> =
+  | {
+      data: T;
+      errors: null;
+      totalCount?: number;
+    }
+  | {
+      data: null;
+      errors: Array<ClerkAPIError>;
+      totalCount?: never;
```
JacobMGEvans, line 20: Is this actually comeback with `undefined` or `null` or is the property just not sent? 

`never` which indicates the values that will never occur, which is not true in this case, it will occur on a valid/error free response.

alexisintech: oh woops let me fix lOL

JacobMGEvans (approved):
Nice, basically the same information as the table anyway lol

royanger (changes requested):
- the new page wasn't added to the manifest.json / isn't linked at all
- I think that `verifySession()` is now deprecated as well
- a generic page for returning, typing and rendering errors might be worth considering. eg, `const { data, errors } = await clerkClient.allowlistIdentifiers.createAllowlistIdentifier({ identifier: 'test@example.com', notify: false });` and checking for the error and console logging or rendering it.

alexisintech:
> * the new page wasn't added to the manifest.json / isn't linked at all
> * I think that `verifySession()` is now deprecated as well
> * a generic page for returning, typing and rendering errors might be worth considering. eg, `const { data, errors } = await clerkClient.allowlistIdentifiers.createAllowlistIdentifier({ identifier: 'test@example.com', notify: false });` and checking for the error and console logging or rendering it.

there's many pages to still be done! this was a test review to make sure that this format is good to move forward with for the rest of the pages (hence why its missing the addition to the manifest.. the pr isn't finished yet!)
also, not sure where you see `verifySession()` ?

alexisintech:
@JacobMGEvans check out the newest commit and lmk what u think about the showing the return type !

docs/references/backend/allowlist/create-allowlist-identifier.mdx
```diff
@@ -18,22 +18,25 @@ function createAllowlistIdentifier: (params: AllowlistIdentifierCreateParams) =>
...
   notify: false,
 });
 
-console.log(data);
-/*
- {
+console.log(response);
+/* response:
+{
+  data: _AllowlistIdentifier {
```
JacobMGEvans, line 32: I think showing the return type like this might confuse some people.

docs/references/backend/allowlist/create-allowlist-identifier.mdx
```diff
@@ -18,22 +18,25 @@ function createAllowlistIdentifier: (params: AllowlistIdentifierCreateParams) =>
...
+const response = await clerkClient.allowlistIdentifiers.createAllowlistIdentifier({
   identifier: 'test@example.com',
   notify: false,
 });
 
-console.log(data);
-/*
- {
+console.log(response);
+/* response:
```
JacobMGEvans, line 30: Maybe having the Type here on the same like as the `response:` comment 🤷🏻

jescalan (approved):

dimkl (changes requested):
This `{ data, errors, ... }` format will be reverted. I will ping you back when is applied. Let's keep the current ( return model or throw error )

royanger:
> This `{ data, errors, ... }` format will be reverted. I will ping you back when is applied. Let's keep the current ( return model or throw error )

That's a shame. I really liked the `{ data, errors }` approach.

JacobMGEvans:
> > This `{ data, errors, ... }` format will be reverted. I will ping you back when is applied. Let's keep the current ( return model or throw error )
> 
> That's a shame. I really liked the `{ data, errors }` approach.

Same

dimkl:
@JacobMGEvans and @royanger it's not a goodbye, it's a see you soon (in v6 probably)

royanger:
> > * the new page wasn't added to the manifest.json / isn't linked at all
> > * I think that `verifySession()` is now deprecated as well
> > * a generic page for returning, typing and rendering errors might be worth considering. eg, `const { data, errors } = await clerkClient.allowlistIdentifiers.createAllowlistIdentifier({ identifier: 'test@example.com', notify: false });` and checking for the error and console logging or rendering it.
> 
> there's many pages to still be done! this was a test review to make sure that this format is good to move forward with for the rest of the pages (hence why its missing the addition to the manifest.. the pr isn't finished yet!) also, not sure where you see `verifySession()` ?

This seemed like 'update the whole section' when I looked. `verifySession()` isn't in the PR, but it is in the overall BAPI section.

royanger:
> @JacobMGEvans and @royanger it's not a goodbye, it's a see you soon (in v6 probably)

SoonTM!

PR #654, opened 2024-01-23
Fix spelling error

JohnVicke:
Found a small spelling error while browsing the docs. Should refer to the `PEM` public key. not ~~PRM~~.


LekoArts (approved):
Thanks!

PR #655, opened 2024-01-24
fix session creation for email/sms otp

kenizaya:


alexisintech (approved):
Great find! Thank you for your contribution 😸💖

PR #656, opened 2024-01-24
Fix broken image for Google block email subaddresses

mzhong9723:
I think this was the issue but lmk if not

image broken here: https://clerk.com/docs/authentication/social-connections/google#block-email-subaddresses


royanger (approved):
Looks good

PR #657, opened 2024-01-25
chore: Add section about publishing status to Google OAuth page

yourtallness:
Quite a few customers productionized with the Google OAuth method set to "Testing" publishing status in Google Cloud Platform.

While they may have onboarded a lot of users in this mode, it is not intended for production usage and as Google starts enforcing limitations, many end users experience the inability to connect.

Therefore we need to stipulate that customers switch their Google OAuth app to "In production" prior to launching.


kylemac (approved):
thanks @yourtallness!

PR #659, opened 2024-01-26
Remove unnecessary check from read-session-data doc

wong2:
`userId` has been checked by the above `if` statement


LekoArts (approved):
Thanks!

PR #660, opened 2024-01-26
Add callout for unescaping special characters in email and SMS templates

mzhong9723:
Updating docs to address: https://linear.app/clerk/issue/CORE-783/apostrophes-in-app-names-are-coming-through-in-email-subject-lines-and


alexisintech (approved):
😸💖

PR #672, opened 2024-01-31
Update sign-in.mdx

wlai-lp:
dark is string value, not a variable


alexisintech (approved):
great catch! thank you for the contribution 😸💖

PR #677, opened 2024-02-01
feat: Override types for roles and permissions

panteliselef:


alexisintech:
I thought we updated the roles to be org:admin, org:member ?
also, why does this section need to be added to the "verify user permissions" page specifically? what is the purpose of this new section?

royanger:
Great info @panteliselef. That said, I think the whole "How to add types for roles and permissions" section should move to /docs/organizations/create-roles-permissions Creating custom roles and creating the global interface for those custom roles and permissions feels directly related.

panteliselef:
@alexisintech @royanger The only reason I added this section to "Verify ..." instread of "Create .." is because "Create" refers to things you can do from the dashboard. I'm not opinionated about it and I trust your instinct, about what is the best fit here.

royanger:
> @alexisintech @royanger The only reason I added this section to "Verify ..." instread of "Create .." is because "Create" refers to things you can do from the dashboard. I'm not opinionated about it and I trust your instinct, about what is the best fit here.

I thought about that too @panteliselef. But thinking a little more I actually agree with you think putting this in 'Verify' is the play. That's the main doc page for working with roles and perms programmatically. 

I would add links to the types info from 'Create' and any page in the 'Building Custom Flows' section that would need this type into.

I know you moved the section to the 'Roles and Permissions' page @alexisintech, but that doesn't feel right. Adding info on the global interface there feels too early and like it can be missed.

alexisintech:
> > @alexisintech @royanger The only reason I added this section to "Verify ..." instread of "Create .." is because "Create" refers to things you can do from the dashboard. I'm not opinionated about it and I trust your instinct, about what is the best fit here.
> 
> I thought about that too @panteliselef. But thinking a little more I actually agree with you think putting this in 'Verify' is the play. That's the main doc page for working with roles and perms programmatically.
> 
> I would add links to the types info from 'Create' and any page in the 'Building Custom Flows' section that would need this type into.
> 
> I know you moved the section to the 'Roles and Permissions' page @alexisintech, but that doesn't feel right. Adding info on the global interface there feels too early and like it can be missed.

okay I can get on board with this actually, it makes sense. let's place it there for now and ship

royanger:
> > > @alexisintech @royanger The only reason I added this section to "Verify ..." instread of "Create .." is because "Create" refers to things you can do from the dashboard. I'm not opinionated about it and I trust your instinct, about what is the best fit here.
> > 
> > 
> > I thought about that too @panteliselef. But thinking a little more I actually agree with you think putting this in 'Verify' is the play. That's the main doc page for working with roles and perms programmatically.
> > I would add links to the types info from 'Create' and any page in the 'Building Custom Flows' section that would need this type into.
> > I know you moved the section to the 'Roles and Permissions' page @alexisintech, but that doesn't feel right. Adding info on the global interface there feels too early and like it can be missed.
> 
> okay I can get on board with this actually, it makes sense. let's place it there for now and ship

Team work makes the dream work! @panteliselef, can't wait to have this live. Thanks for this.

panteliselef:
@alexisintech the changes seems ready. I cannot approve as the PR was created by me. Feel free to approve and merge

panteliselef:
Thank you both again for improving the original content ❤️

royanger:
> @alexisintech the changes seems ready. I cannot approve as the PR was created by me. Feel free to approve and merge

I think the section you added needs to be moved back to the Verify page.

panteliselef:
@alexisintech seems to handled that.

alexisintech (approved):
thanks for the collaboration on this everyone!! 😸💖

PR #680, opened 2024-02-03
fix: remove undefined `user` in Apollo example

alanhoskins:
Use the destructured `getToken` and remove the undefined `user` in the Apollo example for Hasura


alexisintech (approved):
ooo, great catch! thank you 😸💖

PR #682, opened 2024-02-05
wip: new ia proposal

kylemac:
### Getting Started
- [ ] fix the Quickstarts Overview page (loads blank page atm) 
- [ ] make `Project Templates` page in quickstarts / getting started
- [ ] allow users to `Request a Quickstart` (ideally via productlane, but maybe not)
- [ ] should we change the path from `/docs/quickstarts` to `docs/getting-started`? probably not
- [ ] triage all other TODOs

### API Reference
- [ ] Do we need the `Overview` landing page? or can we just have a drop down. The landing page is a good moment to describe the difference between the two, but it creates a little bit of an extra hop. Lets consider whether that hop is useful or not

### UI Components
- [ ] The current implementation just shoves the old "Component Reference" section in to this new `UI Components` top-level type, but long term we want to entirely rethink this section and make something that is more interactive and feels more like a hands-on playground for our components and their properties

### Core Concepts
- [ ] `Client Handshake` feels like the odd one out under Core Concepts
- [ ] Is `Integrating Clerk` the best label?
- [ ] `Users` page is lifted from previous IA. Needs a rethink / overhaul
- [ ] `Organizations` page is lifted from previous IA. Needs a rethink / overhaul
- [ ] We need to write the `Session Management` page
- [ ] We need to write the `Security at Clerk` page
- [ ] If we wanna keep it, the `Client Handshake` needs a review because it's pulled from a PR that has not landed at the time of this writing
- [ ] I'm feeling like this is a weird place to put `System Limits` as well. My next instinct is to put it in `Guides` although it's very un-guidey. That said, I'm leaning towards moving it out of Core Concepts

### Guides
- [ ] There's a lot to do here! Overall, I'm not super happy with the content organization (the headers and the leaf nodes)
- [ ] I also just used `#TODO` as the link for most the entries, even though there's many where there is already existing content. i'll need to take another pass add plug those holes with existing stuff and see what we're left with


alexisintech:
Getting Started Notes:
i love the ideas for adding project templates and the request a quickstart!!
idk about changing the path from quickstarts to getting-started as they will still be called quickstarts (right?), so we may as well keep their paths - less redirects to deal with.

API Reference Notes:
imo, I don't think we need an API Reference landing page. we don't currently have one discussing the differences and docs doesn't get complaints about/relating to that.

Core Concepts Notes:
- I like integrating Clerk, or maybe even "Integrating Clerk into your application" to be more clear to users?
- Client Handshake could go under what used to be Advanced Concepts - so wherever that is getting moved to.
  - What needs a review about it? Sokratis and I finished going over it yesterday (02/05/2024)
- I believe the same could go for System Limits (place under advanced concepts)

PR #683, opened 2024-02-06
DOCS-653 update copy to clarify that allowlists allow/block sign-ups

alexisintech:
[DOCS-653](https://linear.app/clerk/issue/DOCS-653/clarify-that-allow-and-block-lists-block-sign-up-and-sign-ins) reports user complaints on the copy not being clear.

Roy stated "The part you highlighted @nabors does say 'access' but other areas, like the previous sentence, explicitly say "sign-up". This seems to be the confusion with customers."

This PR updates the copy accordingly


royanger (approved):
Looks good.

PR #684, opened 2024-02-06
update trpc guide to specify its for nextjs pages router

alexisintech:
We get complaints about it not being clear that tRPC guide is for pages router only. this is a quick fix until we update the guide & add the app router version


royanger (approved):
Looks good

PR #693, opened 2024-02-08
Added a section for webhook development and troubleshooting

arti-bol:
Many developers do not know about how to start building against webhook. Added a small section to help them build with Clerk. Hope it helps.

### DIFF
--- a/docs/integrations/webhooks.mdx
+++ b/docs/integrations/webhooks.mdx
@@ -165,6 +165,9 @@ From the Replay Messages menu, you will have 3 options.
   alt="The Webhooks page of the Clerk Dashboard. The Replay menu, for replaying failling or missing webhook messages, is open and showing three options."
 />
 
+## Development and Troubleshooting
+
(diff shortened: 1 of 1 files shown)


royanger:
I don't feel like this is the right approach for adding this kind of data. These solutions are more pure troubleshooting. 

I have plans to add more information for TypeScript and covering setting up ngrok/localhost and I think that's the info we want in the docs. 

When it comes to 3rd party tools, we already have an Inngest guide and docs and the benefit of Inngest is that it is actually a devtool and is something that works for dev and production long term.

I'll leave this up to you @alexisintech

alexisintech:
Hmm so I'm going to trust your instinct on this @royanger because I am just not familiar enough with webhooks.
@S3Prototype might have an opinion on this, let's hear from him and then close this out.

alexisintech:
Thanks so much for the contribution idea!
Unfortunately, if we were to add guides around integrating tooling with Clerk, we would need a document that is much more comprehensive and it would require a lot more information. _**We also need to vet companies that we refer customers to in our docs, to ensure the quality of the service and the interaction with Clerk.**_ For right now, we have a guide using Inngest for production development, and Roy is working on a guide to cover ngrok/localhost. I'm going to go ahead and close this PR.

arti-bol:
@alexisintech  - no problem. i trust your team's instinct.

PR #696, opened 2024-02-08
fix: Wrong props for SignOutButton in React quickstart

panteliselef:


royanger (approved):
Thanks for the update!

PR #724, opened 2024-02-22
(core1) remove theme usage; fix mountSignUp

alexisintech:
- `openSignUp` should be `mountSignUp`
- remove use of themes


royanger (approved):
Looks good.

PR #731, opened 2024-02-23
DOCS-5336 fix order of tabs in code example

alexisintech:
[DOCS-5336](https://linear.app/clerk/issue/DOCS-5336/fix-order-of-tabs-for-code-example-httpsclerkcomdocsusersmetadata)


S3Prototype (approved):
Nice! Can you also fix the one on line 336? It's the example under Using the backend API section.

I think the order should be Next, Node, Go, Ruby, Curl

alexisintech:
> Nice! Can you also fix the one on line 336? It's the example under Using the backend API section.
> 
> I think the order should be Next, Node, Go, Ruby, Curl

great eye!! done :)

PR #736, opened 2024-02-26
Added cookie doc to deployments section

royanger:


docs/deployments/clerk-cookies.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: Clerk Cookies
+description: 
+---
+
+# Clerk Cookies
+
+## Strictly Necessary Cookies Application Cookies
```
zythosec, line 8:
```suggestion
## Strictly Necessary Application Cookies
```

docs/deployments/clerk-cookies.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+Clerk will set cookies when your users interact with your Clerk powered application in order to function. These cookies are necessary for the website to function and cannot be switched off in our systems. They are usually only set in response to actions made by the user which amount to a request for services such as signing-up or signing-in. The user could set their browser to block or alert about these cookies, but that will cause some parts of the site to not work. These cookies do not store any personally identifiable information by default. You can modify the cookie but modifying the [session token](/docs/backend-requests/resources/session-tokens) in your [Clerk Dashboard](https://dashboard.clerk.com/last-active?path=jwt-templates).
```
zythosec, line 10:
```suggestion
Clerk will set cookies when your users interact with your Clerk powered application in order to function. These cookies are necessary for the website to function and cannot be switched off in our systems. They are usually only set in response to actions made by the user which amount to a request for services such as signing-up or signing-in. The user could set their browser to block or alert about these cookies, but that will cause some parts of the site to not work. These cookies do not store any personally identifiable information by default. You can modify the cookie by modifying the [session token](/docs/backend-requests/resources/session-tokens) in your [Clerk Dashboard](https://dashboard.clerk.com/last-active?path=jwt-templates). 
```

docs/deployments/clerk-cookies.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+| Cookie Subgroup | Cookies | More Information |
+| --- | --- | --- |
+| .clerk.com | `__session`, `__client_uat`  | first party |
+| .clerk.com, .dashboard.clerk.com (cloudflare) | `_cfuvid`  | third party<br />[Cloudflare Cookies](https://developers.cloudflare.com/fundamentals/reference/policies-compliances/cloudflare-cookies/) |
+
+
+<Callout level="warn">We encourage you to obtain legal advice on using the information here to craft your privacy policy.</Callout>
```
S3Prototype, line 18:
```suggestion
<Callout level="warn">You should seek legal advice before using this information to craft your privacy policy.</Callout>
```
Just wanted to make this more direct

JacobMGEvans (approved):
Nice. I think this is good, I will think about doing a Custom Cookie Management Blog or something in lieu of a guide, depending on needs.

PR #738, opened 2024-02-26
add a note about the beta and linking to upgrade guide to quickstarts

jescalan:
Want to make extra sure that if someone lands on this page, they know they are reading a beta guide, and where to find other upgrade info.


alexisintech (approved):
i like it picasso

PR #739, opened 2024-02-27
(core 1) update JS references for sign-in and sign-up

alexisintech:
copied updates from core 2 JS updates, because as core 1 stands, the code examples were highlighted incorrectly and this page was kinda JANK.

### DIFF
--- a/docs/references/javascript/clerk/sign-in.mdx
+++ b/docs/references/javascript/clerk/sign-in.mdx
@@ -20,11 +20,11 @@ The following methods available on an instance of the [`Clerk`](/docs/references
 
 Render the `<SignIn />` component to an HTML `<div>` element.
 
-### `mountSignIn()` properties
-
 ```typescript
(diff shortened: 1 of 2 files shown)


docs/references/javascript/clerk/sign-in.mdx
```diff
@@ -112,111 +119,138 @@ function unmountSignIn(node: HTMLDivElement): void;
 
   const signInComponent = document.querySelector<HTMLDivElement>('#sign-in')!;
 
-  const clerk = new Clerk('pk_[publishable_key]');
-
+  // Initialize Clerk with your Clerk publishable key
+  const clerk = new Clerk(`{{pub_key}}`);
```
JacobMGEvans, line 123: Why is it double curly braces template literal?

alexisintech: because that's how injectkeys works - it injects a value that needs to be a string. doesn't have to be backticks, we can change to quotes

PR #744, opened 2024-02-27
add getUserOAuthAccessToken to backend sdk

alexisintech:
adds missing operation to backend sdk

### DIFF
--- a/docs/references/backend/types/paginated-resource-response.mdx
+++ b/docs/references/backend/types/paginated-resource-response.mdx
@@ -1,0 +1,21 @@
+---
+title: PaginatedResourceResponse
+description: An interface that describes the response of a method that returns a paginated list of resources.
+---
+
+# `PaginatedResourceResponse`
+
+An interface that describes the response of a method that returns a paginated list of resources.
+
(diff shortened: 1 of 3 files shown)


docs/references/backend/types/paginated-resource-response.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+# `PaginatedResourceResponse`
+
+An interface that describes the response of a method that returns a paginated list of resources.
+
+## Properties
+
+| Properties | Type | Description |
+| --- | --- | --- |
+| `data` | `T[]` |  An array that contains the fetched data. |
```
JacobMGEvans, line 14: Do we provide a generic function that the user can pass their own object to define the types from their fetched data? 

otherwise it should be `unknown[]` or `any[]`

JacobMGEvans: Nevermind I see the example below

JacobMGEvans (approved):
Resolved a comment/question that was answered by the code example. Looks good. 😎

PR #747, opened 2024-02-27
Add callout to React hooks about opting in to resources

alexisintech:
🔎 [useOrganization](https://docs-preview-747.clerkpreview.com/docs/references/react/use-organization)
🔎 [useOrganizationList](https://docs-preview-747.clerkpreview.com/docs/references/react/use-organization-list)


docs/references/react/use-organization-list.mdx
```diff
@@ -100,7 +104,8 @@ type OrganizationSuggestionStatus = "pending" | "accepted";
 To keep network usage to a minimum, developers are required to opt-in by specifying which resource they need to fetch and paginate through. So by default, the `userMemberships`, `userInvitations`, and `userSuggestions` attributes are not populated. You must pass `true` or an object with the desired [properties](#shared-properties) to fetch and paginate the data.
 
 ```jsx
-const { userMemberships } = useOrganizationList(); // userMemberships.data will never be populated.
+// userMemberships.data will never be populated
+const { userMemberships } = useOrganizationList();
```
JacobMGEvans, line 108: Should we not remove this entirely, what is the value of showing users something that doesnt work as expected, i.e. they destructure the value and call the hook expecting to get data in the value

alexisintech: kind of like showing what you should not do because this is what it will result in o.O

JacobMGEvans: Maybe more explicitly "show" a `console.log(userMemberships)` below it the object with `undefined` in a comment, thoughts?
```
{
something: undefined, 
... 
}
```

JacobMGEvans (approved):

PR #751, opened 2024-02-29
OVERHAUL /docs/components/customization/user-profile

alexisintech:
[DOCS-5412](https://linear.app/clerk/issue/DOCS-5412/%F0%9F%98%A2-feedback-for-componentscustomizationuser-profile) is a user feedback ticket that complains: "Unsure on how the custom page works, the code provided doesn't work".

The purpose of this PR is to not only fix the code examples to be standalone, copyable, working examples, but to update the entire guide to better explain how to use this feature.

🔎 [Preview](https://docs-preview-751.clerkpreview.com/docs/components/customization/user-profile)

- removed next.js pages, react and remix examples, as they were the exact same as next.js app router example. added this copy to the top of the guide: "For the sake of this guide, examples are written for Next.js App Router but they are supported by any React meta framework, such as Remix or Gatsby." This is how we are moving forward in the docs - less code to maintain.
- reorganized the document to follow the user's story: either using a dedicated page to host their UserProfile component, or using a UserButton that opens the UserProfile component as a modal.

### DIFF
--- a/docs/components/customization/user-profile.mdx
+++ b/docs/components/customization/user-profile.mdx
@@ -242,7 +242,7 @@ Note that the first item in the navigation sidebar cannot be a `<UserProfile.Lin
 
 <Tabs type="usage" items={["Dedicated Page", "<UserButton />"]}>
   <Tab>
-    It's important to note that the first page in the list will always be rendered under the root path that you set with the `path` prop. It's `url` prop will be ignored, so you can pass any value you want. In the following example, the root path is set to `/user-profile`, so the `<CustomPage />` is rendered under `/user-profile`.
+    It's important to note that the first page in the list will always be rendered under the root path that you set with the `path` prop. It's `url` prop will be ignored, so you can pass any value you want. In the following example, `path` is set to `/user-profile`, so the `<CustomPage />` is rendered under the `/user-profile` path.
 
     ```tsx filename="/app/user-profile/[[...user-profile]]/page.tsx"
     "use client";
--- a/docs/components/customization/user-profile.mdx
+++ b/docs/components/customization/user-profile.mdx
@@ -8,8 +8,10 @@ description: Learn how to add custom pages and include external links within the
 The [`<UserProfile />`](/docs/components/user/user-profile) component supports the addition of custom pages and use of external links in the navigation sidebar.
 
 There are two ways to render the `<UserProfile />` component:
-- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
+- As a modal
 - As a dedicated page
+
+Both can be accessed when the user selects the [`<UserButton />`](/docs/components/user/user-button), and then selects the **Manage account** option.
 
 This guide includes examples for both use cases. You can select one of the following two tabs on the code examples to see the implementation for your preferred use case:
 
--- a/docs/components/customization/user-profile.mdx
+++ b/docs/components/customization/user-profile.mdx
@@ -11,9 +11,10 @@ There are two ways to render the `<UserProfile />` component:
 - As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
 - As a dedicated page
 
-This guide includes examples for both use cases. You can select the tabs on the code examples to see the implementation for your preferred use case.
+This guide includes examples for both use cases. You can select one of the following two tabs on the code examples to see the implementation for your preferred use case:
 
-By default, the `<UserButton />` sets `userProfileMode='modal'`. If you are using the default settings, then you should use the `<UserButton />` examples. However, if you are using the  `userProfileMode='navigation'` and `userProfileUrl='/user-profile'` props on the `<UserButton />` component, then you should use the `Dedicated page` examples.
+- `<UserButton />` tab: By default, the `<UserButton />` sets `userProfileMode='modal'`. If you are using the default settings, then you should select this tab.
+- `Dedicated page` tab: If you do not want the `<UserProfile />` to open as a modal, then you should select this tab. For these examples, you need to set `userProfileMode='navigation'` and `userProfileUrl='/user-profile'` on the `<UserButton />` component.
 
 For the sake of this guide, examples are written for Next.js App Router, but they are supported by any React meta framework, such as Remix or Gatsby.
 
--- a/docs/components/customization/user-profile.mdx
+++ b/docs/components/customization/user-profile.mdx
@@ -1,113 +1,31 @@
 ---
-title: <UserProfile /> customization
+title: Add custom pages and links to the <UserProfile /> component
 description: Learn how to add custom pages and include external links within the navigation sidebar of the <UserProfile /> component.
 ---
 
-# `<UserProfile />` customization
+# Add custom pages and links to the `<UserProfile />` component
 
-The [`<UserProfile />`][userprofile-ref] component supports the addition of custom pages and use of external links in the navigation sidebar. 
+The [`<UserProfile />`](/docs/components/user/user-profile) component supports the addition of custom pages and use of external links in the navigation sidebar.
 
-## `<UserProfile.Page />`
+There are two ways to render the `<UserProfile />` component:
+- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
+- As a dedicated page
 
-Custom pages can be rendered inside the [`<UserProfile />`][userprofile-ref] component and provide a way to incorporate app-specific settings or additional functionality.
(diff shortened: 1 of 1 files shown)


docs/components/customization/user-profile.mdx
```diff
@@ -1,113 +1,31 @@
...
-## `<UserProfile.Page />`
+There are two ways to render the `<UserProfile />` component:
+- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
+- As a dedicated page
 
-Custom pages can be rendered inside the [`<UserProfile />`][userprofile-ref] component and provide a way to incorporate app-specific settings or additional functionality.
+This guide includes examples for both use cases. You can select the tabs on the code examples to see the implementation for your preferred use case.
 
-To add a custom page to the [`<UserProfile />`][userprofile-ref] component, use the `<UserProfile.Page />` component.
+By default, the `<UserButton />` sets `userProfileMode='modal'`. If you are using the default settings, then you should use the `<UserButton />` examples. However, if you are using the  `userProfileMode='navigation'` and `userProfileUrl='/user-profile'` props on the `<UserButton />` component, then you should use the `Dedicated page` examples.
```
alexisintech, line 16: Do we like this better??

```suggestion
This guide includes examples for both use cases. You can select one of the following two tabs on the code examples to see the implementation for your preferred use case:

- `<UserButton />` tab: By default, the `<UserButton />` sets `userProfileMode='modal'`. If you are using the default settings, then you should select this tab.
- `Dedicated page` tab: If you do not want the `<UserProfile />` to open as a modal, then you should select this tab. For these examples, you need to set `userProfileMode='navigation'` and `userProfileUrl='/user-profile'` on the `<UserButton />` component.
```

docs/components/customization/user-profile.mdx
```diff
@@ -1,113 +1,31 @@
...
 
-# `<UserProfile />` customization
+# Add custom pages and links to the `<UserProfile />` component
 
-The [`<UserProfile />`][userprofile-ref] component supports the addition of custom pages and use of external links in the navigation sidebar. 
+The [`<UserProfile />`](/docs/components/user/user-profile) component supports the addition of custom pages and use of external links in the navigation sidebar.
 
-## `<UserProfile.Page />`
+There are two ways to render the `<UserProfile />` component:
+- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
```
S3Prototype, line 11:
```suggestion
- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked.
```
Didn't see anything in the styleguide, but do we add periods to the ends of bullet points? I've been doing it in all my PRs

docs/components/customization/user-profile.mdx
S3Prototype: Yeah, I think this is a little easier to read

docs/components/customization/user-profile.mdx
```diff
@@ -1,113 +1,31 @@
...
+- As a dedicated page
 
-Custom pages can be rendered inside the [`<UserProfile />`][userprofile-ref] component and provide a way to incorporate app-specific settings or additional functionality.
+This guide includes examples for both use cases. You can select the tabs on the code examples to see the implementation for your preferred use case.
 
-To add a custom page to the [`<UserProfile />`][userprofile-ref] component, use the `<UserProfile.Page />` component.
+By default, the `<UserButton />` sets `userProfileMode='modal'`. If you are using the default settings, then you should use the `<UserButton />` examples. However, if you are using the  `userProfileMode='navigation'` and `userProfileUrl='/user-profile'` props on the `<UserButton />` component, then you should use the `Dedicated page` examples.
 
-### Usage
+For the sake of this guide, examples are written for Next.js App Router but they are supported by any React meta framework, such as Remix or Gatsby.
```
S3Prototype, line 18:
```suggestion
For the sake of this guide, examples are written for Next.js App Router, but they are supported by any React meta framework, such as Remix or Gatsby.
```

docs/components/customization/user-profile.mdx
```diff
@@ -116,301 +34,321 @@ All props below are required.
...
-          <UserProfile.Page label="security" />
-        </UserProfile>
-      );
+### Reorder default routes
 
-      export default UserProfilePage;
-      ```
-    </CodeBlockTabs>
-  </Tab>
+If you want to reorder the default routes (`Account` and `Security`) in the `<UserProfile />` navigation sidebar, you can use the `<UserProfile.Page />` component with the `label` prop set to `'account'` or `'security'`. This will target the existing default page and allow you to rearrange it.
```
S3Prototype, line 239:
```suggestion
If you want to reorder the default routes (`Account` and `Security`) in the `<UserProfile />` navigation sidebar, add two `<UserProfile.Page />` components. each with the `label` prop set to either `'account'` or `'security'`. The order in which you render them will be the order in which they appear.
```

My suggestion may not be right, but I didn't quite understand this when I first read it. I _think_ what's happening here is you can label each `<UserProfile.Page>` and adjust the order by swapping their position. If that's so, it might help to say that more explicitly. Would love another opinion though as well cc @jescalan @royanger

docs/components/customization/user-profile.mdx
```diff
@@ -116,301 +34,321 @@ All props below are required.
...
-        <UserProfile path="/user-profile" routing="path">
-          <UserProfile.Page label="Custom Page" url="custom" labelIcon={<CustomIcon />}>
-              <MyCustomPageContent />
-          </UserProfile.Page>
-          <UserProfile.Link label="Homepage" url="/home" labelIcon={<Icon />} />
-          <UserProfile.Page label="account" />
-          <UserProfile.Page label="security" />
-        </UserProfile>
-      );
+    It's important to note that the first page in the list will always be rendered under the root path that you set with the `path` prop. It's `url` prop will be ignored, so you can pass any value you want. In the following example, `path` is set to `/user-profile`, so the `<CustomPage />` is rendered under the `/user-profile` path.
```
S3Prototype, line 245:
```suggestion
<Callout type="info">
The first page in the list will always be rendered under the root path defined with the `path` prop. Its `url` prop will be ignored.
    
In the following example, `path` is set to `/user-profile`, so the `<CustomPage />` is rendered under the `/user-profile` path.
</Callout>
```

May not necessarily need to be a callout, but I think it would look a little better. Also shortened it a bit as it looks a little intimidating in the preview to switch to the Dedicated Page tab and see a whole new paragraph IMO

S3Prototype (commented):
These a great overhaul! I haven't approved yet because I need to test the code snippets, and I will get through that soon 👍

docs/components/customization/user-profile.mdx
JacobMGEvans: >`<UserProfile.Page />` components.
You added a period after the components, preceded by "each with"

docs/components/customization/user-profile.mdx
alexisintech: no periods.
https://www.instructionalsolutions.com/blog/bulleted-list-punctuation#:~:text=Use%20a%20period%20after%20a,bullet%20lists%20or%20all%20fragments.

docs/components/customization/user-profile.mdx
alexisintech: just updated the copy - lmk what you think!

jescalan:
"dedicated page" is the default selected tab, despite not being the default way to render userprofile. perhaps we could swap these?

docs/components/customization/user-profile.mdx
S3Prototype: @JacobMGEvans I first saw periods at the ends of lists in Clerk's docs 🙃. But glad we have clarity now

docs/components/customization/user-profile.mdx
```diff
@@ -1,416 +1,357 @@
...
-# `<UserProfile />` customization
+# Add custom pages and links to the `<UserProfile />` component
 
-The [`<UserProfile />`][userprofile-ref] component supports the addition of custom pages and use of external links in the navigation sidebar. 
+The [`<UserProfile />`](/docs/components/user/user-profile) component supports the addition of custom pages and use of external links in the navigation sidebar.
 
-## `<UserProfile.Page />`
+There are two ways to render the `<UserProfile />` component:
+- As a modal that opens when the [`<UserButton />`](/docs/components/user/user-button) is clicked
+- As a dedicated page
```
S3Prototype, line 13:
```suggestion

Both can be accessed when the user selects their avatar, then selects *Manage account**.

```

While testing the code I didn't immediately know how to see the profile page

S3Prototype (approved):
Nice! The code works for me. Approved with one last suggestion

PR #752, opened 2024-02-29
fix(sdk-go): Added instructions for caching JSON Web Keys

gkats:
When verifying session tokens manually, you can fetch the JSON Web Key for your instance once and cache it until it's rolled. Updated the verifying-sessions page with a rough outline of the SDK methods you would need to fetch the JWK and a brief outline of how you could cache it.

This commit also fixes some links to use a relative path to other docs pages and the references to the secret API key, so that they are valid Go strings.

### DIFF
--- a/docs/references/go/verifying-sessions.mdx
+++ b/docs/references/go/verifying-sessions.mdx
@@ -42,7 +42,7 @@ import (
 )
 
 func main() {
-  clerk.SetKey(`{{secret}}`)
+  clerk.SetKey("{{secret}}")
 
   mux := http.NewServeMux()
   mux.HandleFunc("/", publicRoute)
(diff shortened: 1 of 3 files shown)


docs/references/go/verifying-sessions.mdx
```diff
@@ -128,20 +121,85 @@ func protectedRoute(jwks *clerk.JSONWebKeySet) func(http.ResponseWriter, *http.R
...
-    var jwk *clerk.JSONWebKey
-    for _, k := range jwks.Keys {
-      if k.KeyID == unsafeClaims.KeyID {
-        jwk = k
-        break
-      }
-    }
+If you need to verify session tokens frequently, it might be a good idea to cache the JSON Web Key for your instance, to avoid making too many requests to the Clerk Backend API.
+
+The following snippet repeats the same example as above. However, it demonstrates a possible way to store your JSON Web Key. The example is meant as a rough guide, implementation may vary depending on your needs.
```
alexisintech, line 132:
```suggestion
If you need to verify session tokens frequently, it's recommended to cache the JSON Web Key for your instance in order to avoid making too many requests to the Clerk Backend API.

The following example includes the same code as above, with the addition of demonstrating a possible way to store your JSON Web Key. The example is meant to serve as a guide; implementation may vary depending on your needs.
```

alexisintech (approved):
Minor changes!

PR #754, opened 2024-02-29
update Go overview; update manifest for consistency in title and sidenav title

alexisintech:
The sidenav title for the page was inconsistent with the actual title of the page. The path is "other-examples" and is also inconsistent, but this can addressed in the upcoming IA. No need for adding a redirect right now.

Also, I apologize for these edits slipping through the cracks in the initial PR - but I want this doc as polished as possible since we are currently marketing it!

### DIFF
--- a/docs/references/go/overview.mdx
+++ b/docs/references/go/overview.mdx
@@ -3,9 +3,9 @@ title: Clerk Go SDK
 description: Learn how to integrate Go into your Clerk application.
 ---
 
-# Go
+# Clerk Go SDK
 
-Clerk's [Go SDK](https://github.com/clerk/clerk-sdk-go) is a wrapper over the [Clerk Backend API](https://clerk.com/docs/reference/backend-api).
+The [Clerk Go SDK](https://github.com/clerk/clerk-sdk-go) is a wrapper over the [Clerk Backend API](https://clerk.com/docs/reference/backend-api).
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


docs/references/go/overview.mdx
```diff
@@ -44,17 +50,19 @@ The following examples demonstrate both approaches.
...
+  "github.com/clerk/clerk-sdk-go/v2/$resource$"
 )
 
 // Each operation requires a context.Context as the first argument.
 ctx := context.Background()
 
-// Set the API key
-clerk.SetKey("sk_live_XXX")
+// Set the API key with your Clerk secret key
+clerk.SetKey("{{secret_key}}")
```
gkats, line 65: :question: This is not working code, but rather a placeholder/example. 

Just confirming we want to use the real keys here as well.

docs/references/go/overview.mdx
```diff
@@ -75,24 +83,28 @@ for _, resource := range list.$Resource$s {
...
+  "github.com/clerk/clerk-sdk-go/v2/$resource$"
 )
 
 // Each operation requires a context.Context as the first argument.
 ctx := context.Background()
 
 // Initialize a client with an API key
 config := &clerk.ClientConfig{}
-config.Key = "sk_live_XXX"
+config.Key = "{{secret_key}}"
```
gkats, line 107: Same here about this snippet being pseudo-code in case we don't want to show the actual keys. Feel free to resolve the discussion if we do want that. :smile:

gkats (approved):

docs/references/go/overview.mdx
alexisintech: We try our best to format all of our code examples as working code examples, because users copy and paste them - would it be too much trouble to make these working code examples? I'm not familiar with Go, so if it is too much trouble, we can keep what you had!

alexisintech: okay gotcha ! I'll remove the keys :)

PR #760, opened 2024-03-01
Pull in updates in Core 2 docs into main docs

alexisintech:
Copy and pasted a lot, but removed anything that was a v5 update. Double-checked that the information is accurate for JavaScript v4 (core 1).
These changes are mainly structural/content updates that can be applied to both docs. These changes shoulda been made on main first, and then pulled into Core 2, but wasn't thought about at the time, so we're gonna do it backwards. Get both on the same page.

also, changes "environment keys" to "environment variables" small nit update for @jescalan  :~)


docs/components/unstyled/sign-in-with-metamask.mdx
```diff
@@ -58,16 +58,16 @@ The `<SignInWithMetamaskButton>` component is used to complete a one-click, cryp
...
     );
   }
   ```
 </CodeBlockTabs>
 
-#### Custom usage
+### Custom usage
 
-In some cases you will want to use your own button, or button text. You can do that by wrapping your button up.
+In some cases, you will want to use your own button, or button text. You can do that by wrapping your button in the `<SignInWithMetamaskButton>` component.
```
jescalan, line 70: 🎁 "wrapping the button up" hahaha - good catch

docs/deployments/overview.mdx
```diff
@@ -29,7 +29,7 @@ In most cases, you will want to clone your development instance. This will copy
 
 ## API keys and environment variables
 
-A common mistake when deploying to production is forgetting to change your API keys to your production instances keys. The best way to set this up is to make use of environment variables. All modern hosting providers, such as AWS, GCP, Vercel, Heroku, and Render, make it easy to add these values. Locally, you should use a `.env` file. This way, these values are being set dynamically depending on your environment. Here's a list of Clerk variables you'll need to change:
+A common mistake when deploying to production is forgetting to change your API keys to your production instances keys. The best way to set this up is to make use of environment variables. All modern hosting providers, such as AWS, GCP, Vercel, Heroku, and Render, make it easy to add these values. Locally, you should use an `.env` file. This way, these values are being set dynamically depending on your environment. Here's a list of Clerk variables you'll need to change:
```
jescalan, line 32: are you sure this is right? i would have gone for the previous version, but maybe im bad at grammar idk

jescalan: like, if you remove `.env` and say just "use a file" vs "use an file", the former feels more accurate, and `.env` is basically an adjective in this instance so it shouldn't make an impact right

docs/guides/authjs-migration.mdx
```diff
@@ -319,7 +319,7 @@ This walkthrough will help you move user data from your existing database to Cle
 To retain the user data in your database for easy querying, check out the documentation on [data synchronization with webhooks](/docs/integrations/webhooks/sync-data).
 
 1. Clone `github.com/clerk/migration-script`
-2. Create a `.env` file in the root of the cloned repository with the `CLERK_SECRET_KEY` **of your `Production` instance**.
+2. Create an `.env` file in the root of the cloned repository with the `CLERK_SECRET_KEY` **of your `Production` instance**.
```
jescalan, line 322: Same here 😬

docs/quickstarts/react.mdx
```diff
@@ -48,7 +42,7 @@ Scaffold your new React application using [Vite](https://vitejs.dev/guide/#scaff
   ```
 
   ```bash filename="terminal"
-  yarn create vite clerk-react--template react-ts
+  yarn create vite clerk-react --template react-ts
```
jescalan, line 45: nice nice nice

docs/quickstarts/react.mdx
```diff
@@ -63,32 +57,36 @@ Scaffold your new React application using [Vite](https://vitejs.dev/guide/#scaff
...
   ```
   ```bash filename="terminal"
-  yarn add @clerk/clerk-react
+  yarn add @clerk/clerk-react@beta
   ```
-  
+
   ```bash filename="terminal"
-  pnpm add @clerk/clerk-react
+  pnpm add @clerk/clerk-react@beta
```
jescalan, line 75: These ones we shouldn't pull for sure

docs/quickstarts/remix.mdx
```diff
@@ -31,9 +31,9 @@ Once you have a Remix application ready, you need to install Clerk's Remix SDK.
   ```
 </CodeBlockTabs>
 
-### Set environment keys
+### Set environment variables
 
-Below is an example of a `.env.local` file. 
+Below is an example of an `.env.local` file.
```
jescalan, line 36: 😬

jescalan (changes requested):
One little code snippet uses the beta tag which we should fix before merge but otherwise looks great. Worth noting that there will probably be another slightly ugly rebase after this lands, but it won't be as ugly as the one from yesterday 🙏

docs/deployments/overview.mdx
alexisintech: @LekoArts had a debate on this - and we chose to go with "an" as if someone is saying "an env file" instead of a "a dotenv file". at the end of the day, its preference, and we had to stick with one

docs/quickstarts/react.mdx
alexisintech: WOOPS - removed

jescalan (approved):

PR #764, opened 2024-03-01
docs-5476 Update /javascript/clerk/sign-in and /javascript/clerk-sign-up

alexisintech:
Part of the JavaScript revamp project: JS Quickstart recently got updated with how users can use JS: npm module vs script tag. Examples need to be updated accordingly. Also, we have a lot of user feedback requesting that JavaScript docs show specific examples on how to use each method on a class, etc.

### DIFF
--- a/docs/components/authentication/sign-in.mdx
+++ b/docs/components/authentication/sign-in.mdx
@@ -37,7 +37,7 @@ Below is basic implementation of the `<SignIn />` component. You can use this as
       const SignInPage = () => (
         <SignIn path="/sign-in" routing="path" signUpUrl="/sign-up" />
       );
-      
+
       export default SignInPage;
       ```
     </CodeBlockTabs>
@@ -92,61 +92,43 @@ Below is basic implementation of the `<SignIn />` component. You can use this as
 
     <InjectKeys>
 
-    <CodeBlockTabs type="npm-script" options={['NPM Module', 'window.Clerk']}>
-      ```typescript
+    <CodeBlockTabs type="npm-script" options={['NPM module', '<script>']}>
+      ```js filename="index.js" {14}
       import Clerk from '@clerk/clerk-js';
-      import { dark } from "@clerk/themes";
-
-      // Add a <div> element with id="sign-in" to your HTML
-      document.querySelector<HTMLDivElement>('#app')!.innerHTML = `
-        <div
-          id="sign-in"
-        >
-        </div>
-      `;
-
-      const signInComponent = document.querySelector<HTMLDivElement>('#sign-in')!;
 
       // Initialize Clerk with your Clerk publishable key
-      const clerk = new Clerk(`{{pub_key}}`);
+      const clerk = new Clerk('{{pub_key}}');
       await clerk.load();
 
-      clerk.mountSignIn(signInComponent, {
-        appearance: {
-          baseTheme: dark
-        }
-      });
+      document.getElementById("app").innerHTML = `
+        <div id="sign-in"></div>
+      `;
+
+      const signInDiv =
+        document.getElementById("sign-in");
+
+      clerk.mountSignIn(signInDiv);
       ```
 
-      ```html
-      <!-- Add a <div> element with id="sign-in" to your HTML -->
+      ```html filename="index.html" {10}
+      <!-- Add a <div id="sign-in"> element to your HTML -->
       <div id="sign-in"></div>
 
       <script>
-        const script = document.createElement('script');
-        // Set your Clerk publishable key
-        script.setAttribute('data-clerk-publishable-key', `{{pub_key}}`);
-        script.async = true;
-        script.src = `https://[your-domain].clerk.accounts.dev/npm/@clerk/clerk-js@latest/dist/clerk.browser.js`;
+        window.addEventListener("load", async function () {
+          await Clerk.load();
 
-        script.addEventListener('load', async function () {
-          await window.Clerk.load();
+          const signInDiv =
+            document.getElementById('sign-in');
 
-          const signInComponent = document.querySelector('#sign-in');
-
-          window.Clerk.openSignIn(signInComponent, {
-            appearance: {
-              baseTheme: dark
-            }
-          });
+          Clerk.mountSignIn(signInDiv);
         });
-        document.body.appendChild(script);
       </script>
       ```
     </CodeBlockTabs>
 
     </InjectKeys>
-  </Tab>  
+  </Tab>
 </Tabs>
 
 ## Properties
@@ -170,4 +152,3 @@ To learn about how to customize Clerk components, see the [customization documen
 
 [components-ref]: /docs/components/overview
 [ap-ref]: /docs/account-portal/overview
-
--- a/docs/components/authentication/sign-up.mdx
+++ b/docs/components/authentication/sign-up.mdx
@@ -85,12 +85,48 @@ Below is basic implementation of the `<SignUp />` component. You can use this as
   </Tab>
 
   <Tab>
-    ```js filename="sign-up.js"
-    window.Clerk.mountSignUp(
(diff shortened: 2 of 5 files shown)


docs/references/javascript/clerk/sign-up.mdx
```diff
@@ -1,13 +1,17 @@
 ---
-title: <SignUp /> component
-description: The <SignUp /> component renders a UI for signing up users. The functionality of the <SignUp /> component are controlled by the instance settings you specify in your Clerk Dashboard. You can further customize your SignUp component by passing additional properties at the time of rendering.
+title: JavaScript <SignUp />
```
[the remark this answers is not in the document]
alexisintech: or maybe "<SignUp /> JS Reference" or "Use the <SignUp /> component with JavaScript" or something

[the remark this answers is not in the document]
S3Prototype: This is a good callout, and not an easy issue to solve. "JavaScript <SignUp>" makes sense for sure. I also like your suggestion of "Use the component with JavaScript" but I don't think it suits the page, which is a reference doc, so it's tough.

In my opinion we should lean toward something like "SignUp JavaScript Reference". The [page for the React component](https://docs-preview-764.clerkpreview.com/docs/components/authentication/sign-up) could be called "<SignUp> Component Reference", so it's more obvious from search.

I think the titles for the JS reference stuff should look less like a component, because even though they _are_ components, people think "React" when they see that word, and we should lean toward comprehensibility over accuracy in this case

docs/components/authentication/sign-in.mdx
```diff
@@ -92,61 +92,43 @@ Below is basic implementation of the `<SignIn />` component. You can use this as
 
     <InjectKeys>
 
-    <CodeBlockTabs type="npm-script" options={['NPM Module', 'window.Clerk']}>
-      ```typescript
+    <CodeBlockTabs type="npm-script" options={['NPM module', '<script>']}>
```
S3Prototype, line 95: These updates are good, but instead of having a JavaScript tab here we should probably have a callout in the intro for this page letting users know this is the React component, and they should go to `/docs/references/javascript/clerk/sign-in` to use SignUp in JavaScript.

Having the info in both places makes it easier for discrepancies to accrue over time, and makes navigating the site via docs search or google more confusing.

docs/components/authentication/sign-up.mdx
```diff
@@ -85,12 +85,48 @@ Below is basic implementation of the `<SignUp />` component. You can use this as
   </Tab>
 
   <Tab>
-    ```js filename="sign-up.js"
-    window.Clerk.mountSignUp(
-      document.getElementById("sign-up")
-    );
-    window.Clerk.openSignUp();
-    ```
+    In the example below, the [`mountSignUp()`](/docs/references/javascript/clerk/sign-up#mount-sign-up) method is used to render the `<SignUp />` component to a `<div>` element with `id="sign-up"`.
```
S3Prototype, line 88: We should avoid words like "below" because non-sighted people don't necessarily perceive this information as being above or below. I prefer "The following example".

I'd also suggest avoiding describing the code in detail like this. They can read the code to see the implementation details, and what if the specifics of the example change later but we forget to update this text? Let's just tell them what the example does: "The following example renders the `<SignUp />` component using `mountSignUp()`."

docs/components/authentication/sign-up.mdx
```diff
@@ -85,12 +85,48 @@ Below is basic implementation of the `<SignUp />` component. You can use this as
...
+
+    To learn more about the methods available for rendering and controlling the `<SignUp />` component in a JavaScript application, see the [JavaScript reference docs](/docs/references/javascript/clerk/sign-up).
```
S3Prototype, line 90: Worth mentioning they can't use these examples without following the installation steps specific to using Clerk with JavaScript: `/docs/references/javascript/overview#installation`

docs/quickstarts/javascript.mdx
```diff
@@ -36,7 +36,7 @@ To add the [ClerkJS SDK](/docs/references/javascript/overview) to your JavaScrip
 
 Select your preferred method below to get started.
 
-<Tabs items={["NPM module", "<script>"]}>
+<Tabs type="npm-script" items={["NPM module", "<script>"]}>
```
S3Prototype, line 39: This change is fine, but this page renders all the steps in the ToC, even though half of them are scoped to one tab and half are scoped to the other. Everything below the red line in this screenshot is hidden if "NPM module" is selected, but still is in the ToC:

Screenshot 2024-03-05 at 12 23 55 PM

docs/references/javascript/clerk/sign-in.mdx
```diff
@@ -1,9 +1,13 @@
...
-description: This reference describes the properties and methods related to Clerk's <SignIn /> component for a JavaScript application.
+title: JavaScript <SignIn />
+description: This JavaScript reference describes the properties and methods related to Clerk's <SignIn /> component.
 ---
 
-# `<SignIn />` component
+# JavaScript `<SignIn />`
+
+<Callout type="warning">
+  This JavaScript reference describes the properties and methods related to Clerk's `<SignIn />` component. If you are looking for the React component, see the [`<SignIn />`](/docs/components/authentication/sign-in) documentation.
```
S3Prototype, line 9:
```suggestion
  This reference describes the properties and methods related to using Clerk's `<SignIn />` component in JavaScript. To use it with frameworks such as Next.js or Expo, see [the React component reference](/docs/components/authentication/sign-in).

```

It's a really tricky needle to thread with trying to help people understand whether or not they should be on this page. My suggestion is more explicit in describing potential use cases to help the reader understand where they should be a little more easily I think. And we know [based on data](https://clerkinc.slack.com/archives/C06A9NUVC6L/p1708610328789879?thread_ts=1708605388.110319&cid=C06A9NUVC6L) that Next, React and Expo are the top 3 choices for our users

S3Prototype (approved):
Approving because the changes themselves are worth having, but I recommend making followup linear issues for:
- fixing the ToC's behavior in the JS quickstart
- considering removing the JavaScript tab from the components references, and just linking users to the JS reference with a similar callout to the one you have at the top of the JS reference. You can make that change in this PR instead if you want, either way is fine

docs/references/javascript/clerk/sign-up.mdx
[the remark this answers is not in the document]
jescalan: Yeah agreed, I don't think we use the component syntax here since it's not actually used. Something like "Sign Up Component - JavaScript SDK" feels reasonable to me 🤷‍♂️

docs/quickstarts/javascript.mdx
alexisintech: ugh how annoying - so @jescalan and I did talk about splitting the JS quickstart into two separate tutorial pages in the future, but the `<Tabs />` is our solution for now - possibly after the IA, we can think about this again. what do you think?

docs/references/javascript/clerk/sign-in.mdx
alexisintech: okay good idea! and I updated it to say "the component reference" instead of "React component reference" because it doesn't live in the React reference docs - it lives in the component reference docs.

docs/references/javascript/clerk/sign-up.mdx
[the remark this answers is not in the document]
alexisintech: https://clerkinc.slack.com/archives/CHZ1FBBEG/p1709667576078549

docs/components/authentication/sign-up.mdx
alexisintech: I've added some copy to the new commit that alerts users to follow our quickstart ! (not the js overview, as that is outdated and will be getting reworked)

PR #766, opened 2024-03-04
Update styleguide for lists

S3Prototype:
Adds a rule to the styleguide that say no periods at the ends of lists


alexisintech:
hmm, so I had sent this doc: https://www.instructionalsolutions.com/blog/bulleted-list-punctuation#:~:text=Use%20a%20period%20after%20a,bullet%20lists%20or%20all%20fragments.

basically, if the sentence is a fragment, it doesn't need ending punctuation. if it's a full sentence, then yes it needs punctuation.

PR #772, opened 2024-03-05
use the right import for clerkClient

davemo:


royanger (approved):
Approved

royanger:
@alexisintech -- I believe this should just be for core-2. I'll leave this to make sure that its merged correctly into the two versions.

PR #775, opened 2024-03-06
Add updateUserMetadata() to backend references

itelo:
While searching through the docs, I noticed that the documentation for this function is missing. I am unsure whether it is missing because we should simply use `user.update` or if there is something else to consider. However, I currently use this function in my project, and it seems to work as expected. I hope this helps, Clerk! Thanks!

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -811,6 +811,7 @@
           ["getCount()", "/references/backend/user/get-count"],
           ["createUser()", "/references/backend/user/create-user"],
           ["updateUser()", "/references/backend/user/update-user"],
+          ["updateUserMetadata()", "/references/backend/user/update-user-metadata"],
           ["deleteUser()", "/references/backend/user/delete-user"],
           ["getUserOAuthAccessToken()", "/references/backend/user/get-user-oauth-access-token"]
         ]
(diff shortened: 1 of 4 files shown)


alexisintech:
Hi, thanks for the idea for contributing! This page exists in our beta docs, so if you would like to add to our main docs, you can copy the file and add it to this PR: https://github.com/clerk/clerk-docs/blob/5b7cb41e31f6b3b4d3dd6d553edabb224b149547/docs/references/backend/user/update-user-metadata.mdx

doing so would require you to also add this file: https://github.com/clerk/clerk-docs/blob/5b7cb41e31f6b3b4d3dd6d553edabb224b149547/docs/references/backend/types/backend-user.mdx

and would also require you to update our `manifest.json` to add these two pages to our navigation sidebar. You can see how it's done on [this line](https://github.com/clerk/clerk-docs/blob/5b7cb41e31f6b3b4d3dd6d553edabb224b149547/docs/manifest.json#L839) and [this line](https://github.com/clerk/clerk-docs/blob/5b7cb41e31f6b3b4d3dd6d553edabb224b149547/docs/manifest.json#L1028-#L1031).

if this addition doesn't interest you any longer, or you would like me to take over the addition, just let me know :)

itelo:
@alexisintech, thanks for the feedback!!! I will update it later today!

alexisintech (approved):
awesome, thank you so much for the contribution 😸💖

PR #777, opened 2024-03-06
fix 404 to custom organization switcher document:  path now matches manifest.json

davemo:
This was 404'ing


alexisintech (approved):
AGH great find thanks so much!!

PR #781, opened 2024-03-06
Add H1s to all docs where they're missing

S3Prototype:
This is a companion to the clerk-marketing PR that adds all H1 conent to the SEO title of the page.

This needs to be approved first before merging that one. There is no risk of showing duplicate titles on these pages, as Clerk marketing already shows **either** the frontmatter title or the H1.

[Clerk Marketing PR here](https://github.com/clerk/clerk-marketing/pull/901)

## How to test

If you want to ensure this is the full list of pages with missing H1s, follow these instructions in the `clerk-marketing` repo:

**In `lib/docs/scripts/index-content.mts`**
1. Add `data` to the options on line 142, so it looks like this:
```
.use(remarkPluginExtractSearchRecords, { records, data })
```
**In `lib/docs/plugins/remark-plugin-extract-search-records.mts`**
1. Add `data` to the `RemarkPluginExtractSearchRecordsOptions` type:
```
type RemarkPluginExtractSearchRecordsOptions = {
  records: ExtractedRecord[];
  data?: {
    [key: string]: string;
  };
};
```
2. Add `data` to the options on line 43, so it looks like this:
```
> = ({ records, data }: RemarkPluginExtractSearchRecordsOptions) => {
```
3. Add the following code after line 91, outside of the `visit` method 
```
   if (data && !headings.some(h => h.level === 1)) {
      const { title } = data;
      console.log(title);
    }
``` 
In the console, run `yarn docs:index-content`. It will print a list of the pages that are missing an H1. Ignore the overview pages like "Component Reference" and "Welcome to Clerk Docs"

### DIFF
--- a/docs/deployments/exporting-users.mdx
+++ b/docs/deployments/exporting-users.mdx
@@ -3,9 +3,9 @@ title: Exporting Users
 description: Learn how to export user's data from your Clerk application.
 ---
 
-## Exporting your user's data from Clerk
+# Exporting your user's data from Clerk
 
-### Programmatic access to user data
+## Programmatic access to user data
 By using the [GetUserList](https://clerk.com/docs/reference/backend-api/tag/Users#operation/GetUserList) endpoint of Clerk's Backend API, you can programmatically export your user's data safely. 
 
 Additionally, the [`getUserList()`](/docs/references/backend/user/get-user-list) method in the [Clerk Backend SDK](/docs/references/backend/overview) can also be used to retrieve a list of users.
--- a/docs/deployments/exporting-users.mdx
+++ b/docs/deployments/exporting-users.mdx
@@ -3,9 +3,9 @@ title: Exporting Users
 description: Learn how to export user's data from your Clerk application.
 ---
 
-# Exporting your user's data from Clerk
+# Export your user's data from Clerk
 
-## Programmatic access to user data
+## Access user data in the backend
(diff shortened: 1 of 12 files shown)


docs/deployments/exporting-users.mdx
```diff
@@ -3,9 +3,9 @@ title: Exporting Users
 description: Learn how to export user's data from your Clerk application.
 ---
 
-## Exporting your user's data from Clerk
+# Exporting your user's data from Clerk
```
alexisintech, line 6: [avoid use of gerunds](https://github.com/clerk/clerk-docs/blob/main/STYLEGUIDE.md#24-avoid-gerunds--ing-words)

```suggestion
# Export your user's data from Clerk
```

docs/deployments/exporting-users.mdx
```diff
@@ -3,9 +3,9 @@ title: Exporting Users
...
 
-### Programmatic access to user data
+## Programmatic access to user data
```
alexisintech, line 8: prefer an active voice

```suggestion
## Access user data in the backend
```

alexisintech (commented):
Did we also want to start removing the title from the frontmatter ? We could do that for these pages

alexisintech (approved):
left a few suggestions; approved 😸💖

S3Prototype:
@alexisintech 
> Did we also want to start removing the title from the frontmatter ? We could do that for these pages

As of now, it's not necessary, as the H1 is still prioritized. In future we can just avoid adding the frontmatter.title, and if we really want to we can circle back and delete them from old docs a well

PR #784, opened 2024-03-06
DOCS-5452 adds links to reference documentation on /react/use-user page

alexisintech:
[DOCS-5452](https://linear.app/clerk/issue/DOCS-5452/%F0%9F%98%A2-feedback-for-referencesreactuse-user) is a user feedback ticket that says:
> no informaton on what arguments can be passed to user.update

This PR adds clear reference links in each section on this page.

### DIFF
--- a/docs/references/react/use-user.mdx
+++ b/docs/references/react/use-user.mdx
@@ -82,8 +82,6 @@ export default function Home() {
 
 To retrieve the latest user data after updating the user elsewhere, use the `user.reload()` method.
 
-For more information on the `reload()` method, see the [`User`](/docs/references/javascript/user/user) reference.
-
 ```tsx filename="home.tsx"
 import { useUser } from "@clerk/clerk-react";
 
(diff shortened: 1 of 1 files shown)


docs/references/react/use-user.mdx
```diff
@@ -78,6 +82,8 @@ export default function Home() {
 
 To retrieve the latest user data after updating the user elsewhere, use the `user.reload()` method.
 
+For more information on the `reload()` method, see the [`User`](/docs/references/javascript/user/user) reference.
```
S3Prototype, line 85: Reload is not documented on this page. I could only find it here: https://clerk.com/docs/references/react/use-user#reload-user-data-with-the-use-user-hook. We should add it to this reference. I've [created a Linear issue](https://linear.app/clerk/issue/DOCS-5856/add-reload-to-the-user-js-sdk-reference-page)

docs/references/react/use-user.mdx
```diff
@@ -44,6 +46,8 @@ export default function Home() {
 
 The following example demonstrates how to use the `useUser()` hook to access the `user` object, which includes the `update` method for updating the user's data.
 
+For more information on the `update()` method, see the [`User`](/docs/references/javascript/user/user) reference.
```
S3Prototype, line 49:
```suggestion
For more information on the `update()` method, see the [`User`](/docs/references/javascript/user/user#update) reference.
```
Linking to the heading

S3Prototype (approved):

docs/references/react/use-user.mdx
alexisintech: oh you know what - because reload is only available on the User resource return from useUser . its not available on the User object... hmmm

PR #796, opened 2024-03-08
update remaining usage of login/logged in in favor of sign in

alexisintech:


S3Prototype (approved):
I skipped the spotify docs because I think it's most important to use "sign in" as it relates to users being signed into Clerk. That said, it doesn't hurt to change it there as well. Thanks for catching the other changes too

PR #800, opened 2024-03-12
Add an SSR example to supabase docs

S3Prototype:
This PR adds an example that uses server components, [based on user feedback](https://linear.app/clerk/issue/DOCS-5834/add-an-ssr-example-to-supabase-integration-docs).

- [Preview](https://docs-preview-800.clerkpreview.com/docs/integrations/databases/supabase#fetch-supabase-data-in-your-code)


docs/integrations/databases/supabase.mdx
alexisintech, line 1: This page needs an H1!

docs/integrations/databases/supabase.mdx
```diff
@@ -718,6 +718,75 @@ The following steps will show you how to fetch and send content to your Supabase
 
 </Steps>
 
+### Server component example
```
alexisintech, line 721: So if a user follows the entire guide, but uses this step instead of the "Fetch Supabase data in your code", this piece of code will work?

If so, I believe we break the last step in the tutorial into two parts:
- Fetch Supabase data with a client component
- Fetch Supabase data with a server component (SSR)

S3Prototype: I added a "Server Component" tab to the Fetch Supabase data step to avoid adding a redundant step to the tutorial

S3Prototype: [Preview here](https://docs-preview-800.clerkpreview.com/docs/integrations/databases/supabase#fetch-supabase-data-in-your-code)

PR #802, opened 2024-03-12
DOCS-5894 add working examples to JS password-management reference

alexisintech:
https://docs-preview-802.clerkpreview.com/docs/references/javascript/user/password-management


docs/references/javascript/user/password-management.mdx
```diff
@@ -1,39 +1,282 @@
...
 # User password management
 
-The [`User`](/docs/references/javascript/user/user) object has methods that help you manage a user's password.
+These methods on the [`User`](/docs/references/javascript/user/user) object help you manage a user's password.
+
+The following examples assume that you have followed the [quickstart](/docs/quickstarts/javascript) in order to add Clerk to your JavaScript application.
 
 ## `updatePassword()`
 
+Updates the user's password. Note that passwords must be at least 8 characters long.
```
S3Prototype, line 14:
```suggestion
Updates the user's password. Passwords must be at least 8 characters long.
```

docs/references/javascript/user/password-management.mdx
```diff
@@ -1,39 +1,282 @@
...
+| `currentPassword` | `string` | The user's current password. |
 | `signOutOfOtherSessions?` | `boolean \| undefined` | If set to `true`, all sessions will be signed out. |
 
+### `updatePassword()` example
+
+<InjectKeys>
+
+<Tabs type="npm-script" items={['NPM module', '<script>']}>
+  <Tab>
+    <CodeBlockTabs options={["main.js", "index.html"]}>
```
S3Prototype, line 34: I think it's better to split the `main.js` snippet and `index.html` out instead of having them be two tabs on the same codeblock. I don't think it's immediately obvious that you're supposed to use both to build the full example, and a lot of our docs use code block tabs to show alternatives rather than separate but equally necessary snippets

docs/references/javascript/user/password-management.mdx
```diff
@@ -1,39 +1,282 @@
...
+
+        <script type="module" src="/main.js"></script>
+      </body>
+    </html>
+    ```
+    </CodeBlockTabs>
+  </Tab>
+
+  <Tab>
+  ```html filename="index.html"
```
S3Prototype, line 101: I know we told them to do the JS quickstart first at the top of the page, but this example can't actually work without the `<script>` from the quickstart, correct?

```
  <script
    async
    crossorigin="anonymous"
    data-clerk-publishable-key=your_publishable_key
    src="https://[your-domain].clerk.accounts.dev/npm/@clerk/clerk-js@latest/dist/clerk.browser.js"
    type="text/javascript"
  ></script>
  ```

If that's the case, we should add it to the example, so users can actually copy-paste it and have it work (though they must provide their domain and pub key of course)

alexisintech: yeahhh I thought about this a lot as I wrote the examples - I'll have to update other JS examples so stay tuned for a fix PR !

docs/references/javascript/user/password-management.mdx
alexisintech: We do use codeblocktabs to show multiple files - an example is [here](https://docs-preview-802.clerkpreview.com/docs/advanced-usage/satellite-domains#configure-your-satellite-app-with-properties)

If you believe this shouldn't be the case, we can create a ticket to fix instances in the docs that do this.

Or,
I can add some text that says something like "The tabs below demonstrate what your JavaScript and HTML files could look like when using ClerkJS's NPM module."

What do you think we should do?

S3Prototype: I think for now you can add that text, and in future I think we should avoid this pattern. Rather than trying to fix them all at once in a single ticket, let's just be mindful that when we touch these snippets we should break them up. That way it can be done over time

S3Prototype (approved):
Approving to unblock, though there's one unresolved thread still

docs/references/javascript/user/password-management.mdx
alexisintech: And why do you think we should avoid this pattern? If we add copy to clearly indicate what the tabs are for, then why not utilize our custom components to reduce visual noise/page length?

S3Prototype: It's not consistent design language. Many places in our docs, different tabs = alternative options, eg "App router" and "Pages router". Having copy to indicate things is good, but people skim the docs and expect our code snippets to be structured generally the same way throughout.

S3Prototype: Yes

PR #813, opened 2024-03-15
Update overview.mdx

lifeofpavs:
I have noticed the docs state `secret_key` should be used but actually only `api_key` works


alexisintech (approved):
Great find ! Thank you for the contribution 😸💖

PR #815, opened 2024-03-15
Aa/js create metadata

alexisintech:


docs/references/backend/user/get-user-oauth-access-token.mdx
```diff
@@ -16,7 +16,7 @@ function getUserOauthAccessToken: (userId: string, provider: `oauth_${OAuthProvi
 | Name | Type | Description |
 | --- | --- | --- |
 | `userId` | `string` | The ID of the user to retrieve. |
-| `provider` | <code>oauth_$\{[OAuthProvider](/docs/references/javascript/types/oauth-provider)\}</code> | The OAuth provider to retrieve the access token for. |
+| `provider` | <code>oauth_$\{[OAuthProvider](/docs/references/javascript/types/oauth#o-auth-provider)\}</code> | The OAuth provider to retrieve the access token for. |
```
S3Prototype, line 19:
```suggestion
| `provider` | `string` | The OAuth provider to retrieve the access token for. The value will match the following format: <code>oauth_$\{[OAuthProvider](/docs/references/javascript/types/oauth#o-auth-provider)\}</code> |
```

I suggest something like this because I think the type should still say `string`, but I like that you provided the template for how the value will look

docs/manifest.json
```diff
@@ -637,8 +637,7 @@
           ["SignUpInitialValues", "/references/javascript/types/sign-up-initial-values"],
           ["MagicLinkError", "/references/javascript/types/magic-link-error"],
           ["DeletedObject", "/references/javascript/types/deleted-object"],
-          ["OAuthProvider", "/references/javascript/types/oauth-provider"],
-          ["OAuthStrategy", "/references/javascript/types/oauth-strategy"],
+          ["OAuthProvider and OAuthStrategy", "/references/javascript/types/oauth"],
```
S3Prototype, line 640: I think combining these is probably a good idea, but I wanted to note that the page name looks crowded in the sidebar in the preview. I don't have a suggestion for a fix, and I don't think it should block this PR, but wanted to point it out

docs/references/javascript/sign-in/first-factor.mdx
```diff
@@ -21,7 +21,7 @@ Common scenarios are one-time code (OTP) or social account (SSO) verification. T
...
...: <ul><li>`email_link`: User will receive an email magic link via email. The `email_address_id` parameter can also be specified to select one of the user's known email addresses.</li><li>`email_code`: User will receive a one-time authentication code via email. At least one email address should be on file for the user. The `email_address_id` parameter can also be specified to select one of the user's known email addresses.</li><li>`phone_code`: User will receive a one-time authentication code in their phone, via SMS. At least one phone number should be on file for the user. The `phone_number_id` parameter can also be specified to select which of the user's known phone numbers the SMS will go to.</li><li>`web3_metamask_signature`: The verification will attempt to be completed using the user's web3 wallet public address. The `web3_wallet_id` parameter can also be specified to select which of the user's known web3 wallets will be used. Currently Clerk supports [Metamask](https://metamask.io/).</li><li>`oauth_<provider>`: The user will be authenticated with their social sign in account. See available [OAuth providers](/docs/references/javascript/types/oauth#o-auth-provider).</li></ul> |
```
S3Prototype, line 24: >  user will be authenticated with their social sign in account

Should this be sign in or sign-in?

docs/references/javascript/sign-up/verification.mdx
```diff
@@ -22,7 +22,7 @@ The `prepareVerification` is used to initiate the verification process for a fie
...
+| `strategy` | `'phone_code' \| 'email_code' \| 'email_link' \| 'saml' \| 'oauth_<provider>' \| 'web3_metamask_signature'` | The verification strategy to validate the user's sign-up request.<br/>The following strategies are supported:<br/><ul><li>`phone_code`: Send an SMS with a unique token to input.</li><li>`email_code`: Send an email with a unique token to input.</li><li>`email_link`: Send an email with a link which validates sign-up</li><li>`saml`: The user will be authenticated with SAML. **Experimental**</li><li>`oauth_<provider>`: The user will be authenticated with their social sign in account. See available [OAuth providers](/docs/references/javascript/types/oauth#o-auth-provider).</li><li>`web3_metamask_signature`: The verification will attempt to be completed using the user's web3 wallet public address. The `web3_wallet_id` parameter can also be specified to select which of the user's known web3 wallets will be used. Currently Clerk supports [Metamask](https://metamask.io/).</li></ul> |
```
S3Prototype, line 25: I just noticed the hash is `#o-auth-provider`. That's really weird that it got broken out like that, but it works, so 👍

docs/manifest.json
S3Prototype: I honestly wouldn't mind naming the page OAuth types since that's its title in the actual markdown and SEO. Of course it's different from how every other page in this section is named but I don't think it's different in a confusing way. I leave it up to you though as I don't think it's a huge deal

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
 | --- | --- | --- |
 | `email` | `string` | The value of the email address |
 
-### Returns
+### `createEmailAddress()` returns
 
 | Type | Description |
 | --- | --- |
-| <code>Promise\<[EmailAddress][email-ref]\></code> | The email address created for the user. |
+| <code>Promise\<[EmailAddress][email-ref]\></code> | This method returns a `Promise` which resolves to the the email address created for the user. |
```
S3Prototype, line 32:
```suggestion
| <code>Promise\<[EmailAddress][email-ref]\></code> | This method returns a `Promise` which resolves to the email address created for the user. |
```

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
+
+### `createEmailAddress()` example
+
+The following example adds `test@test.com` as an email address for the user. If the user is not signed in, the user will be prompted to sign in.
```
S3Prototype, line 36:
```suggestion
The following example adds `test@test.com` as an email address for the user. If the user is signed out, they will be prompted to sign in.
```

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
+    });
+  </script>
+```
+</CodeBlockTabs>
+
+</InjectKeys>
 
 ## `createPhoneNumber()`
 
+Adds a phone number for the user. A new [`PhoneNumber`][phone-ref] will be created and associated with the user.
```
S3Prototype, line 106:
```suggestion
This example creates a new [`PhoneNumber`][phone-ref] and associates it with the user.
```

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
-Adds an external account for the user. A new [`ExternalAccount`][exacc-ref] will be created and associated with the user.
-
 ### `CreateExternalAccountParams`
 
 | Name | Type | Description |
 | --- | --- | --- |
-| `strategy` | [`OAuthStrategy`](/docs/references/javascript/types/oauth-strategy) | The strategy corresponding to the oauth provider, e.g. `oauth_facebook`, `oauth_github`, etc |
-| `redirectUrl?` | `string` | The URL to redirect back to one the oauth flow has completed successfully or unsuccessfully. |
+| `strategy` | [`OAuthStrategy`](/docs/references/javascript/types/oauth#o-auth-strategy) | The strategy corresponding to the OAuth provider. For example: `oauth_facebook`, `oauth_github`, etc. |
+| `redirectUrl?` | `string` | The URL to redirect back to one the OAuth flow has completed successfully or unsuccessfully. |
```
S3Prototype, line 309:
```suggestion
| `redirectUrl?` | `string` | The URL to redirect back to once the OAuth flow has completed successfully or unsuccessfully. |
```

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
+| `Promise<ExternalAccount>` | This method returns a `Promise` which resolves to the [`ExternalAccount`][exacc-ref] created for the user. |
+
+### `createExternalAccount()` example
+
+After calling `createExternalAccount`, the initial `state` of the returned `ExternalAccount` will be `unverified`. To initiate the connection with the external provider, redirect the user to the `externalAccount.verification.externalVerificationRedirectURL` contained in the result of `createExternalAccount`.
+
+Upon return, inspect within the `user.externalAccounts` the entry that corresponds to the requested strategy:
+
+- If the connection was successful, then `externalAccount.verification.status` should be `verified`.
+- If the connection was not successful, then the `externalAccount.verification.status` should not be `verified` and the `externalAccount.verification.error` will contain the error encountered so that you can present corresponding feedback to the user.
```
S3Prototype, line 325:
```suggestion
- If the connection was not successful, then the `externalAccount.verification.status` will be `unverified` and the `externalAccount.verification.error` will contain the error encountered, which you can present to the user.
```

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
-Upon return, one can inspect within the `user.externalAccounts` the entry that corresponds to the requested strategy:
+| `Promise<ExternalAccount>` | This method returns a `Promise` which resolves to the [`ExternalAccount`][exacc-ref] created for the user. |
+
+### `createExternalAccount()` example
+
+After calling `createExternalAccount`, the initial `state` of the returned `ExternalAccount` will be `unverified`. To initiate the connection with the external provider, redirect the user to the `externalAccount.verification.externalVerificationRedirectURL` contained in the result of `createExternalAccount`.
+
+Upon return, inspect within the `user.externalAccounts` the entry that corresponds to the requested strategy:
+
+- If the connection was successful, then `externalAccount.verification.status` should be `verified`.
```
S3Prototype, line 324:
```suggestion
- If the connection succeeded, then `externalAccount.verification.status` will be `verified`.
```

I prefer "will" because it's more definite. "Should" seems to imply that the connection can succeed and the status can still be "unverified". If that's the case, we definitely gotta document that, but if not we should say "will".

docs/references/javascript/user/create-metadata.mdx
```diff
@@ -1,100 +1,463 @@
...
+- If the connection was not successful, then the `externalAccount.verification.status` should not be `verified` and the `externalAccount.verification.error` will contain the error encountered so that you can present corresponding feedback to the user.
+
+The following example demonstrates how to add a Notion account as an external account for the user. When the user clicks the "Add Notion as a social connection" button, the user will be redirected to Notion to connect their account. After connecting their account, they will be redirected back to the `/` route of your application and the status of the connection will be displayed.
```
S3Prototype, line 327:
```suggestion
The following example demonstrates how to add a Notion account as an external account for the user. When the user selects the "Add Notion as a social connection" button, the user will be redirected to Notion to connect their account. After connecting their account, they will be redirected to the `/` route of your application, and the status of the connection will be displayed.
```

S3Prototype (approved):
Approved with suggestions

docs/references/backend/user/get-user-oauth-access-token.mdx
alexisintech: here's the function definition with the type 😳

Screenshot 2024-03-18 at 17 48 17

what do u think

docs/references/javascript/sign-in/first-factor.mdx
alexisintech: sign-in ! good catch 👁️

docs/references/javascript/user/create-metadata.mdx
alexisintech: I think the status will be `failed` if the verification fails. the status can also be `unverified`.

Screenshot 2024-03-18 at 17 54 58

docs/references/backend/user/get-user-oauth-access-token.mdx
S3Prototype: Oh, interesting! This is a [template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html), so my suggestion is wrong.

docs/references/javascript/user/create-metadata.mdx
S3Prototype: Got it. Your original wording is the most accurate way to describe it I think. Although I prefer "will not be" over "should not be", as it leaves less room for confusion. I assume it's not expected behavior that an unsuccessful connection will sometimes be `verified`. If that is expected, we'll need to explain why.

docs/references/backend/user/get-user-oauth-access-token.mdx
alexisintech: I have learned something new about types!

PR #816, opened 2024-03-15
DOCS-5924 add working examples to JS TOTP methods reference

alexisintech:
🔎 [TOTP methods](https://docs-preview-816.clerkpreview.com/docs/references/javascript/user/totp)


docs/manifest.json
```diff
@@ -522,12 +522,12 @@
         { "title": "User" },
         [
           ["User", "/references/javascript/user/user"],
```
S3Prototype, line 524: What do you think of calling this "User object" in the sidebar, to at least make it a little less weird to open the "User" dropdown and just see "User" again?

S3Prototype (approved):
This is a good example.

Not sure if we do this everywhere in the reference docs but I think it's really smart to put the example in its own section at the bottom, especially since the example puts the methods together 👍

docs/references/javascript/user/totp.mdx
```diff
@@ -95,3 +103,269 @@ Generates a fresh new set of backup codes for the user. Note that if this method
 | `codes` | `string[]` | The generated set of backup codes |
 | `createdAt` | `Date` | Creation date of the TOTP secret |
 | `updatedAt` | `Date` | Update timestamp of the TOTP secret |
+
+## TOTP example
+
+The following example demonstrates how to use the TOTP methods to create and verify a TOTP secret, disable TOTP, and create backup codes for a user. Upon any error, the example will display the error message to the user.
```
S3Prototype, line 109:
```suggestion
The following example demonstrates how to use the TOTP methods to create and verify a TOTP secret, disable TOTP, and create backup codes for a user. Upon any error, the example will display the error message to the user.

<Callout type="Info">
This example requires you to enable **Authenticator application** and **Backup codes** as MFA strategies in your Clerk dashboard.
</Callout>
```

Didn't realize I had to do this

docs/manifest.json
alexisintech: great minds think alike ! I was thinking this for all the classes - I agree and I'll update accordingly

docs/references/javascript/user/totp.mdx
alexisintech: Yes, it's in a callout at the top. I can make the callout at the top more clear, or I can add the callout here as well? We need the callout at the top for sure, or we could add the callout to each method, but its for sure the first step users should do before trying to use any of the methods.

alexisintech:
> This is a good example.
> 
> Not sure if we do this everywhere in the reference docs but I think it's really smart to put the example in its own section at the bottom, especially since the example puts the methods together 👍

I felt it was kind of necessary for this page, because the createTOTP and verifyTOTP go together - and then I was like, hell lets throw it all in one instance and reduce visual noise on the page, and give users a true working example. 
It can be tricky to do that for other pages, as they might be require calling methods that aren't introduced on that exact page - for example, getting into the organization docs, you'd need to build a full out custom flow for each page of methods at that point. and then the user is getting sent to other pages of the docs trying to figure out where some of the methods came from, because they weren't introduced on that page - yeah it gets tricky. I was thinking of how to do it, and actually wrote out entire custom flows and real working examples, but 1. don't want to confuse the readers and 2. can just point them to our custom flows at that point LOL no need to dupe code

docs/references/javascript/user/totp.mdx
S3Prototype: Ah, I see. Yeah, I totally missed it at the top, but I think I would've read it if it was above the intro paragraph, so maybe that's the solution.

Is an authenticator app required for using TOTP? If so, we should mention it specifically in the callout

docs/references/javascript/user/totp.mdx
```diff
@@ -3,79 +3,87 @@ title: Time-based One-time Password
...
+function createTOTP(): Promise<TOTPResource>;
+```
 
-#### Returns
+### `createTOTP()` returns
 
 | Type | Description |
 | --- | --- |
-| <code>Promise\<[TOTPResource](#totp-resource)\></code> | The newly created TOTP object. |
+| <code>Promise\<[TOTPResource](#totp-resource)\></code> | This method returns a `Promise` which resolves with the newly created TOTP object. |
```
S3Prototype, line 30:
```suggestion
| <code>Promise\<[TOTPResource](#totp-resource)\></code> | This method returns a `Promise` which resolves to the newly created TOTP object. |
```

In the other PRs we've been saying "resolves to" instead of "resolves with"

S3Prototype (commented):
Looks good 👍

PR #818, opened 2024-03-16
Remove non-existent property from SignOutButton (redirectUrl)

rhijjawi:
Documentation implies that this: `<SignOutButton signOutOptions={{ redirectUrl: "/"}}>` works. It does not.


S3Prototype:
Thanks!

S3Prototype (approved):

S3Prototype:
I will merge this soon, but I'm checking with the team to see if this property was ever valid. If so I'd like to mark it as deprecated somewhere as well as removing it with this PR. Will respond soon

rhijjawi:
@S3Prototype, iirc this became a global setting.

S3Prototype (approved):
Just checked with the team @rhijjawi. To use `redirectUrl` on `SignOutButton`, install the Core 2 version of the clerk-nextjs SDK:

`npm i @clerk/nextjs@beta`

Thanks for pointing this out, though, as it doesn't work in Core 1 and should be removed from the docs 👍

PR #823, opened 2024-03-18
DOCS-5937 add comprehensive example to the JS /organization/members reference

alexisintech:
https://docs-preview-823.clerkpreview.com/docs/references/javascript/organization/members


docs/references/javascript/organization/members.mdx
```diff
@@ -1,42 +1,105 @@
...
-Retrieve the members of the currently active organization.
-
 ### `GetMembersParams`
 
 | Name | Type | Description |
 | --- | --- | --- |
 | `initialPage?` | `number` | A number that can be used to skip the first n-1 pages. For example, if `initialPage` is set to 10, it is will skip the first 9 pages and will fetch the 10th page. |
 | `pageSize?` | `number` | A number that indicates the maximum number of results that should be returned for a specific page. |
-| `role` | `string` | The [roles](/docs/organizations/roles-permissions) of memberships that will be included in the response. |
+| `role?` | <code>[`OrganizationCustomRoleKey`](#organization-custom-role-key)[]</code> | The [roles](/docs/organizations/roles-permissions) of memberships that will be included in the response. |
```
S3Prototype, line 24:
```suggestion
| `role?` | <code>[OrganizationCustomRoleKey](#organization-custom-role-key)[]</code> | The [roles](/docs/organizations/roles-permissions) of memberships that will be included in the response. |
```
The backticks made it nested in two `<code>` blocks:

image

Which makes me wonder, why do we need to use `<code>` here? Would the backticks (`) not work?

docs/references/javascript/organization/members.mdx
```diff
@@ -1,42 +1,105 @@
...
+    <div id="sign-in"></div>
+  `;
+
+  const signInDiv = document.getElementById("sign-in");
+
+  clerk.mountSignIn(signInDiv);
+}
+```
+
+```html filename="index.html"
```
S3Prototype, line 63: When I run the `<script>` example I get this error:
```
TypeError: Cannot read properties of null (reading 'getMemberships')
```
I do have an organization in my clerk app, and the user I'm signed in as is part of that organization. Is there something else I may be missing here? I checked in the browser console and my user object is defined, but Clerk.organization is null

S3Prototype: Maybe my organization is not active? But I'm not sure how to fix that

docs/references/javascript/organization/members.mdx
alexisintech: The backticks would only cover `OrganizationCustomRoleKey` and its leave its array signature `[]` out of the snippet. And we cannot use backticks for the entire thing because of the MDX link - it just doesn't work. So we use <code>

alexisintech: Good catch!

S3Prototype (approved):
Nice 👍

PR #853, opened 2024-03-27
Add note about how users cannot change domains for dev instances

S3Prototype:
Just a small callout. Based on customer feedback:

> It would be good to mention either how to add a custom domain for a development environment, or state that it's not possible. Thank you!

(https://linear.app/clerk/issue/DOCS-5442/add-a-note-about-how-users-cannot-change-domains-for-dev-instances)


alexisintech (approved):
LGTM! 😸💖

PR #867, opened 2024-04-01
[DOCS-6039] Add pagination params in invitations/get-invitation-list

dimkl:
Changes:

- Add pagination params in invitations/get-invitation-list
- Update limit description for sessions & users to be consistent with the other APIs


S3Prototype (approved):
Thanks!

alexisintech (approved):
😸💖

PR #871, opened 2024-04-01
DOCS-5112 update rate limiting doc

alexisintech:
From [this ticket](https://linear.app/clerk/issue/DOCS-5112/update-rate-limiting-page):

- [X] confirm that the JWKS endpoint is not rate limited — @kylemac confirms it is not.
- [X] because JWKS endpoint  is not rate limited, we can add it to the Backend API Requests table
- [X] add a note that currentUser() is using the getUser() BAPI endpoint so all currentUser() uses count against the 100 reqs / 10 seconds rate limit
- [X] add a note that the FAPI rate limits are per user, while the BAPI endpoints are per application

🔎 https://docs-preview-871.clerkpreview.com/docs/backend-requests/resources/rate-limits
🔎 https://docs-preview-871.clerkpreview.com/docs/references/nextjs/current-user

### DIFF
--- a/docs/backend-requests/resources/rate-limits.mdx
+++ b/docs/backend-requests/resources/rate-limits.mdx
@@ -5,15 +5,19 @@ description: Learn about rate limiting on Clerk's APIs.
 
 # Rate limits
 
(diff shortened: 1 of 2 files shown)


S3Prototype (approved):
Great addition

PR #872, opened 2024-04-01
email sms template updates

alexisintech:
- updates path / filename to more accurately represent its title
- update clerk dashboard navigation instructions
- removes unnecessary screenshots
- removes use of "click"


alexisintech:
@S3Prototype whenever you have a chance, can you let me know what other screenshots you think are probably worth removing?

S3Prototype:
@alexisintech IMO we should take the stance that every screenshot in the docs is only there because it's absolutely necessary. After giving this page a quick look, I don't see any screenshots that should be there. Do you think any of these are required for the user to understand the information?

alexisintech:
> @alexisintech IMO we should take the stance that every screenshot in the docs is only there because it's absolutely necessary. After giving this page a quick look, I don't see any screenshots that should be there. Do you think any of these are required for the user to understand the information?

I think the WYSIWG editor screenshots are understandable. We want to create a context in the docs that explains the Dashboard, how to navigate it, and the different features/settings it has. This is one of those instances, and in some cases, might require screenshots. Like these in the video, I think are valuable, but feel free to disagree and let me know !

https://github.com/clerk/clerk-docs/assets/98043211/06632dc2-0999-4dc3-9efa-e9372abfce34

alexisintech:
@S3Prototype I've posted a new commit [dc9b286](https://github.com/clerk/clerk-docs/pull/872/commits/dc9b28602a79e2052e0473c828b70576e2625540) that removes almost all screenshots and reorganizes the page - lmk if you like this solution better

S3Prototype:
I like it 👍

S3Prototype (approved):

PR #905, opened 2024-04-12
remove e2e testing from docs

alexisintech:


nexus-codingg (approved):
LGTM! this will be very helpful in reducing confusion around E2E testing until a proper solution is shipped

PR #909, opened 2024-04-16
Update basic-rbac.mdx, Correcting typos and grammar.

Ehren12:
Corrected typographical errors and grammatical structure.

### DIFF
--- a/docs/guides/basic-rbac.mdx
+++ b/docs/guides/basic-rbac.mdx
@@ -11,7 +11,7 @@ This guide assumes that you are using Next.js App Router. The concepts here can 
 
 ## Configure the session token
 
-Clerk provides [user metadata](/docs/users/metadata#user-metadata), a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
+Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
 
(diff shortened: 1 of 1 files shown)


docs/guides/basic-rbac.mdx
```diff
@@ -11,11 +11,11 @@ This guide assumes that you are using Next.js App Router. The concepts here can
 
 ## Configure the session token
 
-Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
+Clerk provides [user metadata](/docs/users/metadata#user-metadata), a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
```
alexisintech, line 14: this is already grammatically correct

```suggestion
Clerk provides [user metadata](/docs/users/metadata#user-metadata), which is a tool that can be leveraged to build flexible custom logic into your application. Metadata can be used to store information, and in this case, it can be used to store a user's role.
```

docs/guides/basic-rbac.mdx
```diff
@@ -155,7 +155,7 @@ export default function AdminDashboard() {
 ```
 
 <Callout type="info">
-  You can modify the behavior of the helper function to meet your needs. Maybe it will return the roles that the user has, or you could create a `protectByRole()` and have that function handle the redirect.
+  You can modify the behaviour of the helper function to meet your needs. Maybe it will return the roles that the user has, or you could create a `protectByRole()` and have that function handle the redirect.
```
alexisintech, line 158: we use american spelling in our docs

```suggestion
  You can modify the behavior of the helper function to meet your needs. Maybe it will return the roles that the user has, or you could create a `protectByRole()` and have that function handle the redirect.
```

alexisintech (changes requested):

PR #914, opened 2024-04-17
Added isLoading to useEffect deps.

hosenur:
Without waiting for loading to complete, the useEffect was always forwarding to `/sign-in` , added isLoading to dependency and forward only when loading is complete and userId doesn't exist.


alexisintech:
Thank you for the contribution 😸💖

PR #922, opened 2024-04-19
Core 1: Long Lived Backport Target

kylemac:
⚠️ This PR should not be merged ⚠️


manovotny:
Closing this out — it has been superseded by the frozen Core 1 archive.

Core 1 docs are now served from the `frozen-core-1` branch in `clerk/clerk`, which vendors the full Core 1 content in-repo and reads it off the local filesystem. Nothing builds from this `core-1` branch anymore: `clerk.com/docs/core-1/*` is rewritten to the frozen deployment, so there is no longer a backport target for this PR to track.

The branch itself stays put for now — its content is already duplicated inside `frozen-core-1`, so nothing is lost either way.

PR #929, opened 2024-04-22
Fix authMiddleware imports in the docs

S3Prototype:
We list `authMiddleware` as being imported from `@clerk/nextjs`. This is wrong. It's from `@clerk/nextjs/server`.


krishna-santosh:
@S3Prototype Clerk is promoting the use of clerkMiddleware in place of authMiddleware, wouldn't it be more appropriate to completely chnage the example to use clerkMiddleware instead of adding /server ?

https://clerk.com/docs/references/nextjs/auth-middleware

S3Prototype:
@krishna-santosh I believe that authmiddleware page exists for posterity, as it has a note at the top warning people to use clerkMiddleware instead

image

Is that what you're referring to?

krishna-santosh:
Yeah, You are right, but there's one example that appears right after creating a new project, it still show the old example which is causing error, I guess that should be updated, I was assuming your PR was referring to that issue, I haven't looked at the changes, my bad.

nikosdouvlis (approved):

S3Prototype:
@krishna-santosh No worries, and thanks for looking! I'll post in the proj-dashboard slack channel to let them know!

PR #956, opened 2024-04-26
Updating links to revolvapp product and docs.

izaaklauer:
The old links now redirect to the [redactor home page](https://imperavi.com/redactor/).

It's worth noting that the new links have a `/legacy/` path, and Revolvapp is now listed as "legacy" [on the imperavi home page](https://imperavi.com/#:~:text=Redactor%20Classic-,Revolvapp,-Books)


alexisintech (approved):
thank you!

PR #969, opened 2024-05-01
Update env var guide

S3Prototype:
This PR:

- Explains that our env vars work with meta-frameworks, and pure React users should stick to the ClerkProvider props ([per this discussion](https://clerkinc.slack.com/archives/C04KENP2GFL/p1712946780211689?thread_ts=1712946018.335729&cid=C04KENP2GFL))
- Removes React env var examples, as VITE env vars would only work for users deploying React with Vite

### DIFF
--- a/docs/deployments/clerk-environment-variables.mdx
+++ b/docs/deployments/clerk-environment-variables.mdx
@@ -8,13 +8,17 @@ You can use environment variables to configure how your Clerk app behaves, such 
 
 This page is a reference for all available Clerk environment variables.
 
+## Compatibility
+
+In the frontend, Clerk's environment variables work for most popular meta-frameworks, such as Next.js or Remix. If you're building a pure React app, you should use the props on the components you're using.
+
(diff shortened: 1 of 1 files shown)


docs/deployments/clerk-environment-variables.mdx
```diff
@@ -8,13 +8,17 @@ You can use environment variables to configure how your Clerk app behaves, such
 
 This page is a reference for all available Clerk environment variables.
 
+## Compatibility
+
+In the frontend, Clerk's environment variables work for most popular meta-frameworks, such as Next.js or Remix. If you're building a pure React app, you should use the props on the components you're using.
+
+For example, to force users to redirect to a specific URL after signing in, you would use the `signInForceRedirectUrl` prop on [`<ClerkProvider>`](/docs/components/clerk-provider) rather than the `CLERK_SIGN_IN_FORCE_REDIRECT_URL` environment variable.
```
alexisintech, line 15:
```suggestion
In the frontend, Clerk's environment variables work for most popular meta-frameworks, such as Next.js or Remix. 

If you're building a pure React app, you should use the props on the components you're using. For example, to force users to redirect to a specific URL after signing in, you would use the `signInForceRedirectUrl` prop on [`<ClerkProvider>`](/docs/components/clerk-provider) rather than the `CLERK_SIGN_IN_FORCE_REDIRECT_URL` environment variable.
```

alexisintech (approved):

PR #702, opened 2024-02-12
(WIP - core2) Update custom flows

alexisintech:
🔎 [/authentication/sign-up-sign-in-options](https://docs-preview-702.clerkpreview.com/docs/authentication/configuration/sign-up-sign-in-options)
🔎 [/components/user-button](https://docs-preview-702.clerkpreview.com/docs/components/user/user-button)
🔎 [/custom-flows/add-email](https://docs-preview-702.clerkpreview.com/docs/custom-flows/add-email)
🔎 [/custom-flows/add-mfa](https://docs-preview-702.clerkpreview.com/docs/custom-flows/add-mfa)
🔎 [/custom-flows/add-oauth-connection](https://docs-preview-702.clerkpreview.com/docs/custom-flows/add-oauth-connection)
🔎 [/custom-flows/email-links](https://docs-preview-702.clerkpreview.com/docs/custom-flows/email-links)
🔎 [/custom-flows/email-password-mfa](https://docs-preview-702.clerkpreview.com/docs/custom-flows/email-password-mfa)
🔎 [/custom-flows/email-password](https://docs-preview-702.clerkpreview.com/docs/custom-flows/email-password)
🔎 [/custom-flows/email-sms-otp](https://docs-preview-702.clerkpreview.com/docs/custom-flows/email-sms-otp)
🔎 [/custom-flows/oauth-connections](https://docs-preview-702.clerkpreview.com/docs/custom-flows/oauth-connections)
🔎 [/custom-flows/overview](https://docs-preview-702.clerkpreview.com/docs/custom-flows/overview)
🔎 [/custom-flows/sign-out](https://docs-preview-702.clerkpreview.com/docs/custom-flows/sign-out)

redirects to be made in marketing repo:
/organizations/creating-organizations → docs/custom-flows/creating-organizations
/organizations/custom-organization-switcher → custom-flows/custom-organization-switcher
/custom-flows/mfa → /custom-flows/email-password-mfa
/organizations/inviting-users → /custom-flows/inviting-users
/organizations/managing-roles→ /custom-flows/managing-roles
/organizations/metadata→ /custom-flows/metadata
/organizations/updating-organizations→ /custom-flows/updating-organizations
/organizations/verified-domains→ /custom-flows/verified-domains
/organizations/viewing-memberships → /custom-flows/viewing-memberships
/authentication/configuration/email-options → /authentication/configuration/email-sms-templates
/authentication/configuration/sign-in-options → /authentication/configuration/sign-up-sign-in-options
[...]

PR #971, opened 2024-05-01
Update email-password custom flows guide

S3Prototype:
This pr:

- Updates our instructions on enabling authentication options. The current steps lead to a 422 error.
- Adds a React import to the React examples to prevent errors
- Clarifies that user credentials will be stored in Clerk, but other services can still be used with Clerk auth

These are all in response to user feedback [you can find in this linear issue](https://linear.app/clerk/issue/DOCS-6138/test-and-update-our-custom-flows-email-password-docs).

### DIFF
--- a/docs/custom-flows/email-password.mdx
+++ b/docs/custom-flows/email-password.mdx
@@ -5,36 +5,32 @@ description: Clerk provides the useSignUp and useSignIn hooks, which allow you t
 
 # Custom email/password flow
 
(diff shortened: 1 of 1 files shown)


alexisintech:
this is already being worked on with roy and I in our custom flows project https://github.com/clerk/clerk-docs/pull/702

alexisintech (approved):

PR #973, opened 2024-05-01
Update repetitive headings for Backend and JS SDK docs

alexisintech:


S3Prototype:
It sucks that we can't test how the search results look before merging this. Or maybe we can? Perhaps @BRKalow knows?

Last time I tried testing how search would look with some changes, I came to the conclusion that the search index in localhost and previews is using the index we have in prod, which means triggering reindexing locally does nothing. [Discussion here](https://clerkinc.slack.com/archives/C01QFRQNHSN/p1709741454774399?thread_ts=1709673479.358929&cid=C01QFRQNHSN).

Could be wrong of course, that's just the conclusion I came to.

alexisintech:
> It sucks that we can't test how the search results look before merging this. Or maybe we can? Perhaps @BRKalow knows?
> 
> Last time I tried testing how search would look with some changes, I came to the conclusion that the search index in localhost and previews is using the index we have in prod, which means triggering reindexing locally does nothing. [Discussion here](https://clerkinc.slack.com/archives/C01QFRQNHSN/p1709741454774399?thread_ts=1709673479.358929&cid=C01QFRQNHSN).
> 
> Could be wrong of course, that's just the conclusion I came to.

yes search only works for prod currently

docs/references/backend/organization/get-organization-list.mdx
```diff
@@ -20,9 +20,9 @@ function getOrganizationList: (params: GetOrganizationListParams) => Promise<Pag
 | `query?` | `string` | A search query to filter organizations by. |
 | `orderBy?` | `name \| updated_at \| members_count` | The field to order by. Prefix with a `-` to reverse the order. Prefix with a `+` to list in ascending order. |
 
-## `getOrganizationList()` examples
+## Examples
 
-### `getOrganizationList()` with no parameters
+### Example with no parameters
```
S3Prototype, line 25: I'm realizing pages like this will clutter the search results as well, because there are multiple headings with `getOrganizationList` in them. I don't think we should address that in this PR, because it would require the more in depth work of changing each example's heading to describe what the example does rather than show which param is being passed.

I added an issue for this to triage: https://linear.app/clerk/issue/DOCS-6264/update-example-headings-to-describe-what-the-examples-are

S3Prototype (approved):

docs/references/backend/organization/get-organization-list.mdx
S3Prototype: No, we'd simplify the sentence we have under each of the headings. So for [your example](https://clerk.com/docs/references/backend/organization/get-organization-invitation-list#get-organization-invitation-list-organization-id-status), it says:

> Retrieves organization invitation list that is filtered by the status of the invitation.

We could rename the heading to something as simple as "Filter by invitation status". Just something that puts what the example does in simple terms. I think users would find that a more useful thing to see in the ToC as well

docs/references/javascript/organization-domain.mdx
```diff
@@ -76,9 +70,3 @@ function attemptAffiliationVerification(params: AttemptAffiliationVerificationPa
 | Name | Type | Description |
 | --- | --- | --- |
 | `code` | `string` | The one-time code that was sent to the user as part of this verification step. |
-
-### `attemptAffiliationVerification()` returns
```
S3Prototype, line 80: Why do we delete the "returns" headings and content here but keep them in other files?

S3Prototype (approved):
👍

PR #974, opened 2024-05-02
Elements: Update code block demo props

bradlc:
- Add `priority` prop to demos "above the fold" (https://github.com/clerk/clerk/pull/370)
- Use `rem` units for `height` styles
- Set background colours (https://github.com/clerk/clerk/pull/369)

### DIFF
--- a/docs/elements/examples/sign-in.mdx
+++ b/docs/elements/examples/sign-in.mdx
@@ -15,12 +15,13 @@ Before you build your sign-in flow, you need to configure the appropriate settin
 1. In the **Contact information** section, ensure that *only* **Password** is enabled.
 1. In the **Username** section, ensure that **Username** is enabled.
 
-<CodeBlockDemo demos={[
+<CodeBlockDemo priority demos={[
   {
     key: 'sign-in-username-password',
     demo: '/demo/elements/sign-in/username-password',
     style: {
-      height: 568
+      height: `${568 / 16}rem`,
+      backgroundColor: 'rgb(244 244 245)'
     }
   }
 ]}>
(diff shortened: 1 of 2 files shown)


docs/elements/examples/sign-in.mdx
```diff
@@ -15,12 +15,13 @@ Before you build your sign-in flow, you need to configure the appropriate settin
...
 1. In the **Username** section, ensure that **Username** is enabled.
 
-<CodeBlockDemo demos={[
+<CodeBlockDemo priority demos={[
   {
     key: 'sign-in-username-password',
     demo: '/demo/elements/sign-in/username-password',
     style: {
-      height: 568
+      height: `${568 / 16}rem`,
```
alexcarpenter, line 23: If the expectation is that these should always render in rems, can we just accept a numbered px value instead of needing to always do the conversion here?

bradlc: I think it would depend on the demo. In this case the height is rem-based because the size of everything inside the demo is rem-based.

Contrived example but if your demo was this then you'd want to set the height in `px`:

```jsx
<div style={{ height: '500px' }} />
```

bradlc: I also think it's worth keeping the `style` prop 1:1 with the standard React `style` prop

alexcarpenter: Got it, I guess I don't see when we'd use px in the demo, since we're using TW but non-blocker. Another doc block on height usage would be helpful.

alexcarpenter (approved):

docs/elements/examples/sign-in.mdx
bradlc: 💯 This will be documented in [CONTRIBUTING.md](https://github.com/clerk/clerk-docs/blob/main/CONTRIBUTING.md) in the near future

PR #976, opened 2024-05-02
Update core 1 docs link

S3Prototype:
We need to add an icon and change it from a simple link to a dropdown that says something like "Previous Docs", with "Core 1" being the only item in the list.

Icon changes must happen in a pr to clerk/clerk


bradlc:
Added the icon for this here: https://github.com/clerk/clerk/pull/376

Are you settled on "Previous Version Docs" @S3Prototype?

PR #1022, opened 2024-05-10
Update backend docs to explain error handling

S3Prototype:
[This plain issue](https://app.plain.com/workspace/w_01GT53BQWV3DFW6ECTWZNQ1E9K/thread/th_01HTJD9HCQFTGG0EEFA3N5KPAW/) has the following note:

> We need to make it clear in our docs for backend and node sdk that if there's an error, the methods won't throw, they will instead return the error.

This PR handles that.


alexisintech (approved):

jescalan (approved):
This looks great - I might just add a little code block example of a typical response pattern somewhere? like `const { data, errors } = someBackendApiCall()`

PR #1026, opened 2024-05-10
Added some info based on feedback from customer

royanger:
- added info about type exports
- moved 'type' header up a level so it is easier to find for customers
- added info about returning success codes
- added info on installing ngrok

### DIFF
--- a/docs/integrations/webhooks/sync-data.mdx
+++ b/docs/integrations/webhooks/sync-data.mdx
@@ -36,10 +36,11 @@ To test a webhook locally, you will need to expose your local server to the inte
 1. Go to the [ngrok website](https://ngrok.com/) and create an account.
 2. Once you have made it to the [ngrok dashboard](https://dashboard.ngrok.com/get-started/setup/macos), in navigation sidebar, select [**Domains**](https://dashboard.ngrok.com/cloud-edge/domains).
 3. Select **Create domain**.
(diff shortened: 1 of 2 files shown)


docs/integrations/webhooks/sync-data.mdx
```diff
@@ -126,6 +127,8 @@ Create a Route Handler that uses `svix` to verify the webhook signature and that
 
 For the sake of this guide, you will only log the payload to the console. In a real world application, you would use the payload to trigger some action. For example, you are listening for the `user.updated` event, so you could perform a database `update` or `upsert` to update the user's details.
 
+Your webhook will need to return a success code like `200` or `201` when it has finished running successfully, or an error code. If it returns an error code the Webhook will reflect that in the Dashboard log and the webhook will [retry](/docs/integrations/webhooks/overview#retry). When the webhook returns a success code there will be no retries and the Webhook will show a success status in the Dashboard.
```
alexisintech, line 130:
```suggestion
Your webhook will need to return either an error code, or a success code, like `200` or `201`. If it returns an error code, the webook will reflect that in the Dashboard log and it will [retry](/docs/integrations/webhooks/overview#retry). When the webhook returns a success code, there will be no retries and the webhook will show a success status in the Dashboard.
```

alexisintech (approved):

PR #1030, opened 2024-05-13
docs: minor clean up for remix.mdx

mattmazzola:
- Remove duplicate Meta declarations (Same values set in both function and in html)
  - Chose to keep values in function over HTML since then all meta is the same location
- Rename `routes/_index.html` to `routes/_index.tsx`
  -  Given it is an "index" route, it is likely never intended to show in the url. This means it should be prefixed with underscore. This also more closely matches with structure of default remix application
- Remove extra empty lines


alexisintech (approved):
amazing! thank you :)

PR #1034, opened 2024-05-13
update error handling for custom flows guide

alexisintech:
https://clerk.com/docs/pr/1034/custom-flows/error-handling

### DIFF
--- a/docs/references/javascript/types/clerk-api-error.mdx
+++ b/docs/references/javascript/types/clerk-api-error.mdx
@@ -1,0 +1,35 @@
+---
+description: An interface that represents an error returned by the Clerk API.
+---
+
+# `ClerkAPIError`
+
+An interface that represents an error returned by the Clerk API.
+
+```typescript
+export interface ClerkAPIError {
+  code: string;
+  message: string;
+  longMessage?: string;
+  meta?: {
+    paramName?: string;
+    sessionId?: string;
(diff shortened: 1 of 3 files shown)


docs/references/javascript/types/clerk-api-error.mdx
```diff
@@ -0,0 +1,35 @@
+---
+description: An interface that represents an error returned by the Clerk API.
+---
+
+# `ClerkAPIError`
+
+An interface that represents an error returned by the Clerk API.
+
+```typescript
```
S3Prototype, line 9: I'm not necessarily opposed to this, but it doesn't seem like we put the interface in a codeblock like this for the other type references.

I think it's smart to add this because the `meta` property is so complex, but I worry we'll forget that reasoning and in the future feel the need to add entire interfaces to codeblocks in other types, which is kind of redundant.

What do you think of adding something like this below the property table?

> The `meta` property consists of the following structure:
> ```typescript
>  meta?: {
>    paramName?: string;
>    sessionId?: string;
>   emailAddresses?: string[];
>    identifiers?: string[];
>    zxcvbn?: {
>      suggestions: {
>        code: string;
>        message: string;
>      }[];
>    };
>    permissions?: string[];
>  };
> ```

S3Prototype (approved):

PR #1047, opened 2024-05-15
DOCS-649 update references overview to match sidenav

alexisintech:
waiting on elysia, c#, and astro svg's


docs/references/overview.mdx
skullface, line 1: the `style` and `stroke-width` tags from the original SVG downloaded from Linear wasn’t playing nicely as vanilla HTML:

```html
<svg xmlns="http://www.w3.org/2000/svg" id="Layer_1" data-name="Layer 1" version="1.1" viewBox="0 0 400 400"><defs><style>.cls-1,.cls-7{fill:#d6d6d6;stroke-width:0}.cls-7{fill:#fff}</style></defs><path d="[...]" style="fill:#262626;stroke-width:0"/><path d="[...]" style="fill:#565656;stroke-width:0"/><path class="cls-7" d="[...]"/>
```

needed to be transformed to JSX/TSX syntax. In this case, I extrapolated the styles from insides the `<defs></defs>` and applied them directly to the `<path>`s.

```js
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 400 400"><path d="[...]" fill="#262626" strokeWidth="0"/><path d=""[...]" fill="#565656" strokeWidth="0"/><path fill="#fff" strokeWidth="0" d="[...]"/>
```

S3Prototype (approved):

PR #1055, opened 2024-05-17
Update add-onboarding-flow.mdx

Itachi-Uchiha581:
RedirectToSignIn is a depreciated dependency and was not being used in this example

### DIFF
--- a/docs/guides/add-onboarding-flow.mdx
+++ b/docs/guides/add-onboarding-flow.mdx
@@ -72,7 +72,6 @@ Note that the following example protects all routes except one. This is so that 
 
 ```tsx filename="src/middleware.ts"
 import { clerkMiddleware, createRouteMatcher } from "@clerk/nextjs/server";
-import { redirectToSignIn } from "@clerk/nextjs/server";
 import { NextRequest, NextResponse } from "next/server";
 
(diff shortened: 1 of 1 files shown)


docs/guides/add-onboarding-flow.mdx
```diff
@@ -72,7 +72,6 @@ Note that the following example protects all routes except one. This is so that
 
 ```tsx filename="src/middleware.ts"
 import { clerkMiddleware, createRouteMatcher } from "@clerk/nextjs/server";
-import { redirectToSignIn } from "@clerk/nextjs/server";
```
S3Prototype, line 75: Thank you! Can you remove it where it's destructured on line 81?

Itachi-Uchiha581: It can't be removed as it would result in an error at line 90

S3Prototype: Ah, I see I missed that. Doesn't that mean this example does use redirectToSignIn?

S3Prototype: Ah, I see now : 👍

S3Prototype (approved):

PR #1090, opened 2024-05-28
Correct the path to JWT-related functions

folushooladipo:
The prior docs stated that functions like decodeJwt were to be imported from @clerk/backend/tokens, though @clerk/backend/jwt is the correct address based on installing @clerk/backend@^1.2.1


S3Prototype:
Thanks! Double checking with the SDK team to make sure this is intended

folushooladipo:
> Thank you ❤️

Very glad to be of help!

Issue #1122, opened 2024-06-03
Incorrect property name for email address in Clerk webhook documentation


rupaut98:
The Clerk webhook documentation for Inngest integration contains an incorrect property name for accessing the user's email address. The property should be email_address instead of email.

The current code in the documentation:
```
const email = user.email_addresses.find(e =>
  e.id === user.primary_email_address_id
).email;
```

The corrected code should be:

```
const email = user.email_addresses.find(e =>
  e.id === user.primary_email_address_id
).email_address;
```

PR #1123, opened 2024-06-03
Fix incorrect email property name in Clerk webhook documentation

rupaut98:
This pull request fixes the incorrect property name for accessing the user's email address in the Clerk webhook documentation for Inngest integration. The property should be email_address instead of email.

Current code in the documentation:
```
const email = user.email_addresses.find(e =>
  e.id === user.primary_email_address_id
).email;
```

Corrected code:
```
const email = user.email_addresses.find(e =>
  e.id === user.primary_email_address_id
).email_address;
```

The current documentation uses .email, which does not exist in the email_addresses object structure. The correct property is .email_address, which aligns with the actual data structure returned by the Clerk API.

This fix addresses the issue described in Issue #1122 . By making this change, developers following the documentation will be able to correctly access the user's primary email address without encountering errors.

### DIFF
--- a/docs/integrations/webhooks/inngest.mdx
+++ b/docs/integrations/webhooks/inngest.mdx
@@ -128,7 +128,7 @@ const sendWelcomeEmail = inngest.createFunction(
    const { first_name } = user
    const email = user.email_addresses.find(e =>
      e.id === user.primary_email_address_id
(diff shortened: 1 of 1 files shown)


S3Prototype (approved):
Thank you!

PR #1136, opened 2024-06-05
Add some clarification on how preview app upgrades work with support

jescalan:
The text in there now makes it seem like support will create some sort of special new app for you, but this isn't the case. Wording adjusted to accurately capture the process!


S3Prototype (approved):
Thanks!

PR #1141, opened 2024-06-06
Clean up /verify-user-permissions page

alexisintech:
> [!IMPORTANT]
> 🔎 [Preview](https://clerk.com/docs/pr/1141/organizations/verify-user-permissions#verify-the-active-user-s-permissions-in-an-organization)

While [updating the `auth()` documentation to better explain the `protect()` helper](https://github.com/clerk/clerk-docs/pull/1140), I noticed that this page needed a lot of cleaning up - and we actually had a ticket for it!

This PR:
- cleans up copy that explains the code examples
- adds missing filenames to codeblocks
- updates the props for codeblocks to our new syntax 
  - `{{ filename: '' }}`
- updates the names of tabs in code blocks to be consistent throughout the doc 
  - `has()` `<Protect>` and `auth().protect()`

### DIFF
--- a/docs/organizations/verify-user-permissions.mdx
+++ b/docs/organizations/verify-user-permissions.mdx
@@ -5,9 +5,8 @@ description: A collection of utility functions and components in order to allow 
 
 # Verify the active user's permissions in an organization
 
-<Callout type="info">
-  The following authorization checks are predicated on a user having an active organization. Without this, they will likely always evaluate to false by default. Clerk's built-in [`<OrganizationSwitcher/>`](/docs/components/organization/organization-switcher) component and[`useOrganizationList().setActive({ organization: <orgId> })`](/docs/references/react/use-organization-list) method are two available options to set an organization as active for a user.
-</Callout>
+> [!IMPORTANT]
+> The following authorization checks are predicated on a user having an active organization. Without this, they will likely always evaluate to false by default. Clerk's built-in [`<OrganizationSwitcher/>`](/docs/components/organization/organization-switcher) component and[`useOrganizationList().setActive({ organization: <orgId> })`](/docs/references/react/use-organization-list) method are two available options to set an organization as active for a user.
 
 In general, you should always verify whether or not a user is authorized to access sensitive information, important content, or exclusive features. The most secure way to implement authorization is by checking the active user's [role or permissions](/docs/organizations/roles-permissions#permissions).
 
@@ -25,7 +24,7 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
 
 <Tabs items={["<Protect>", "has()"]}>
   <Tab>
-    The following example uses Clerk's `<Protect>` component to only render the form for users with the `org:team_settings:manage` permission. The example uses the `fallback` prop to render a different UI if the user is not authorized.
+    The following example uses Clerk's `<Protect>` component to only render the form for users with the correct permission. The example uses the `fallback` prop to render a different UI if the user is not authorized.
 
     ```tsx {{ filename: '/app/dashboard/settings/form.tsx' }}
       'use client';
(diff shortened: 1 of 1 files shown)


docs/organizations/verify-user-permissions.mdx
```diff
@@ -48,9 +48,11 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
   </Tab>
 
   <Tab>
-    The `useAuth()` hook can help you access what permissions a user has by using `has()`. In the example below, `has({ permission: "org:team_settings:manage" })` returns `false` if the user does not have the `org:team_settings:manage` permission.
+    Use the [`useAuth()`](/docs/references/react/use-auth) hook to access the `has()` helper in Client Components.
 
-    ```tsx filename="/app/dashboard/settings/form.tsx"
+    The following example uses `has()` to inspect a user's permissions granularly. If the user doesn't have the `org:team_settings:manage` permission, `has()` returns `false`, causing the component to return `null` instead of rendering its children.
```
S3Prototype, line 53: IMO it's not entirely necessary to describe what these examples are doing in detail like this, as they can understand all that from reading the code. I'd personally prefer a high-level description of what the example's goal is, like "The following example renders a UI **only if** the user has the necessary permissions." or something like that

docs/organizations/verify-user-permissions.mdx
```diff
@@ -48,9 +48,11 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
...
+
+    ```tsx {{ filename: '/app/dashboard/settings/form.tsx' }}
```
S3Prototype, line 55: There's nothing wrong with doing this, but also it's not necessary because we transform the old convention into the new ones behind the scenes ([discussion](https://clerkinc.slack.com/archives/C06CA0JUZ8D/p1717697264843089?thread_ts=1717697144.373819&cid=C06CA0JUZ8D))

docs/organizations/verify-user-permissions.mdx
```diff
@@ -74,139 +76,177 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
...
       }
       ```
 
   </Tab>
 
   <Tab>
-    <Callout>
-      `auth().protect()` will throw a `notFound` Next.js error if authorization checks fail.
-    </Callout>
+      <Callout type="warning">
```
S3Prototype, line 120: Does it make a difference if we use `<Callout>` rather than our new callout syntax?

docs/organizations/verify-user-permissions.mdx
```diff
@@ -74,139 +76,177 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
...
+        `auth().protect()` only works for App Router and is considered experimental.
+      </Callout>
 
-    ```tsx filename="/app/dashboard/settings/layout.tsx"
+      The following example uses [`auth().protect()`](/docs/references/nextjs/auth#protect) to protect a RSC from unauthenticated and unauthorized access.
```
S3Prototype, line 124: This is a great breakdown 👍

docs/organizations/verify-user-permissions.mdx
```diff
@@ -74,139 +76,177 @@ The examples below work for both SSR and CSR. Examples are written for Next.js A
...
+            return users.getTeams(userId)
+          }
+        ```
+      </Tab>
 
-        if(!has({permission: "org:team_settings:read"})){
-          return Response.json({error: "Forbidden"}, {status: 403})
-        }
+      <Tab>
+        <Callout type="warning">
```
S3Prototype, line 207: This callout as well – should it be updated to new callout syntax?

S3Prototype (approved):
Approved to unblock, but with comments

docs/organizations/verify-user-permissions.mdx
alexisintech: yes but its removing tech debt as we move along

docs/organizations/verify-user-permissions.mdx
alexisintech: probably should use the new syntax ! i'll remove the tech debt

S3Prototype:
@alexisintech Looks good 👍 I resolved the one open convo about the `{{filename:'x'}}` syntax

PR #1156, opened 2024-06-11
fix(CSP): fix directive syntax, ensure all necessary values are present

BRKalow:
Updates the CSP docs to ensure the provided values are formatted correctly. Also uses our `{{fapi_url}}` placeholder to inject the user's instance FAPI host.


docs/security/clerk-csp.mdx
```diff
@@ -41,4 +44,6 @@ const nextConfig = {
 };
 
 module.exports = nextConfig;
-```
\ No newline at end of file
+```
+
+Note that you will need to make sure any third-party domains where scripts or assets are loaded from are also specified.
```
octoper, line 49: Do we have a component to make the  note more noticable?

octoper (approved):

docs/security/clerk-csp.mdx
BRKalow: Yeah, can update!

PR #1166, opened 2024-06-15
Fix missing imports from Node SDK usage

wobsoriano:
In the Node SDK example on the Clerk docs site, there are sections where resources are destructured from the `clerkClient` function and imported directly. These methods are not exported, resulting in an error.

- https://clerk.com/docs/references/nodejs/overview#manual-authentication
- https://clerk.com/docs/references/nodejs/available-methods#instantiate-a-default-clerk-client-instance

This change updates those sections to just import `clerkClient` and access the resource from it.

Before:

```ts
import { sessions } from '@clerk/clerk-sdk-node';

// more code

const session = await sessions.verifySession(sessionId, clientToken);
```

After:

```ts
import { clerkClient } from '@clerk/clerk-sdk-node';

// more code

const session = await clerkClient.sessions.verifySession(sessionId, clientToken);
```

### DIFF
--- a/docs/references/nodejs/available-methods.mdx
+++ b/docs/references/nodejs/available-methods.mdx
@@ -18,14 +18,6 @@ If you would like to use the default instance of `clerkClient` provided by the S
 
   const userList = await clerkClient.users.getUserList();
   ```
-
-  If you are interested in a certain resource, you can destructure `clerkClient` to access just that resource.:
-
-  ```js
-  import { sessions } from '@clerk/clerk-sdk-node';
-
-  const sessionList = await sessions.getSessionList();
-  ```
   </Tab>
 
   <Tab>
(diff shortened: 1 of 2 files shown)


docs/references/nodejs/available-methods.mdx
```diff
@@ -38,18 +30,6 @@ If you would like to use the default instance of `clerkClient` provided by the S
...
-
-  ```js
-  const pkg = require('@clerk/clerk-sdk-node');
-  const { clients } = pkg;
-
-  clients
-    .getClient(clientId)
-    .then((client) => console.log(client))
-    .catch((error) => console.error(error));
-  ```
```
wobsoriano, line 52: We don't need this part since this import doesn't exist anyway

alexisintech (approved):
Good find! Thank you so much 😸💖

PR #1176, opened 2024-06-18
Change "width" param to behave as a min-width

izaaklauer:
### What's changing?

We are updating the `width` param to return an image larger than the requested width

### What's the current behavior?

Currently, the img.clerk.com is returning the exact width requested. We're updating that API to return a few specific widths, rather than infinitely customizable widths

### Why not respect the api as-written?

The current docs claim that `width` is a maximum width. We don't believe that to be a very useful API. Developers generally care about getting the smallest possible image that will look good at a specific width, not the largest possible (but still a bit blurry) image that will be under a specific size.

### What impact will this change have?

This won't impact any users of clerk's SDK. Users who are consuming the API directly and have an image width defined will sometimes receive larger-than-requested images, but layout will not break. Usage sampling of our API has revealed only these two categories of users thus far.

### Timing considerations

This docs update will be deployed simultaneously with the API change.


alexisintech (approved):
Thank you for the explanation in the description 😸💖

PR #1211, opened 2024-06-30
Replace dotenv with @next/env in neon guide

kevinzunigacuellar:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation

This PR removes the usage of `dotenv` from the neon guide. While a small change, it favors the use of framework-specific utility libraries and reduces unnecessary dependencies, thus simplifying the overall setup.

### Changes Made

- Replaced `dotenv` with [`@next/env`](https://www.npmjs.com/package/@next/env) utility for loading environment variables.
- Added a check for `DATABASE_URL` in `drizzle.config.ts`.

### DIFF
--- a/docs/integrations/databases/neon.mdx
+++ b/docs/integrations/databases/neon.mdx
@@ -37,7 +37,7 @@ This tutorial demonstrates how to integrate Neon Postgres with Clerk in a Next.j
 1. Navigate to the project directory and install the required dependencies:
     ```sh {{ filename: 'terminal' }}
     cd clerk-neon-example
-    npm install @neondatabase/serverless dotenv
+    npm install @neondatabase/serverless
     npm install drizzle-orm --legacy-peer-deps
     npm install -D drizzle-kit
     ```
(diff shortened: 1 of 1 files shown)


docs/integrations/databases/neon.mdx
```diff
@@ -124,14 +126,18 @@ Use the `drizzle-kit` package to generate and run database migrations.
...
     ```typescript {{ filename: 'drizzle.config.ts' }}
     import { defineConfig } from "drizzle-kit";
-    import * as dotenv from "dotenv";
+    import { loadEnvConfig } from "@next/env";
+
+    loadEnvConfig(process.cwd());
 
-    dotenv.config({ path: ".env.local" });
+    if (!process.env.DATABASE_URL) {
+      throw new Error("DATABASE_URL is not defined");
```
royanger, line 134:
```suggestion
      throw new Error("DATABASE_URL must be a Neon postgres connection string");
```
Let's use consistent messaging.

royanger (changes requested):
Thanks for the PR! This guide is suggesting to use the standard `.env.local` file to store environment variables, which Next has built-in support for. Let's remove `dotenv` and `next/env` altogether, keep in the second check for the DB URL.

kevinzunigacuellar:
> Let's remove `dotenv` and `next/env` altogether, keep in the second check for the DB URL.

Thanks for the review Roy

Unfortunately, removing `next/env` from the `drizzle.config` file will cause an error because the Drizzle ORM CLI doesn’t rely on the Next.js server to run scripts, so these variables won't be available at runtime.

While it is possible to remove `next/env` from `schema.ts`, it’s a common practice to load environment variables at this point for scripts that run outside the Next.js server context.

royanger (approved):

kevinzunigacuellar:
Thanks for the support @royanger 🎉

PR #1213, opened 2024-07-01
Update User Backend SDK docs to match BAPI docs; add clerk/javascript file to every Backend SDK method doc

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

While checking that our /backend/* reference docs were up to date ([due to this linear ticket](https://linear.app/clerk/issue/DOCS-6039/check-that-our-docsreferencesbackend-pages-match-our-backend-api)), I tracked which files are present by adding their corresponding `clerk/javascript` file in a comment on their respective page. This helped me find multiple missing methods (see the previously tagged linear ticket).

Also, adding these links will make it easy to find where these methods exist in the `clerk/javascript` repo when updating their respective pages.

For example, the source code for `/backend/allowlist/create-allowlist-identifier` exists at `clerk/javascript` file: https://github.com/clerk/javascript/blob/main/packages/backend/src/api/endpoints/AllowlistIdentifierApi.ts#L22 so I added a comment on this doc with a link to this file.

---

Also, while checking that our Backend SDK docs match our BAPI docs, I am updating one sub-api at a time. This PR serves to update the User sub-api.


jescalan (commented):
This is really great as an initial effort, but seems like a fragile long term solution. If anything changes in the backend sdk at all, the references here break, without us having any way of knowing. This seems like the type of thing that should be automated in some way, rather than put in manually.

I think it's still probably better to have this here now than nothing, but I'd love to see some more thought put in to how we can make this more of a sustainable thing 🙏

alexisintech:
abstracted `user` sub-api related changes to here https://github.com/clerk/clerk-docs/pull/1240

PR #1224, opened 2024-07-03
Supabase doc updates

bmorrisondev:
> [!IMPORTANT]
> 🔎 Preview:
>
> - [Supabase integration guide](https://clerk.com/docs/pr/1224/integrations/databases/supabase)

I walked through this document for the first time today and found a number of areas that could be expanded on or improved for clarity.

This includes:

- Adding context as to what the guide is, and is not
- Adding details on using the SQL editor moer
- Reordering the steps to create the requesting_user_id first (Step 1 does not work otherwise)
- Expanding on the RLS policies
- A small tweak based on our UI changes


jescalan (approved):
Really impressive work here 👏

It's quite an involved guide and I wonder if there's anything we can do to like, automate some of these steps or make it less of an episode to get the integration done, but after reading it through nothing is jumping out to me. It may just be what it is!

bmorrisondev:
> Really impressive work here 👏
> 
> It's quite an involved guide and I wonder if there's anything we can do to like, automate some of these steps or make it less of an episode to get the integration done, but after reading it through nothing is jumping out to me. It may just be what it is!

bmorrisondev:
My mistake on closing this! 

We _could_ build a small CLI that hooks into the database using the connection string to run the necessary SQL on the required tables. Do we have a way of determining how popular this integration is? Maybe checking how many people have a supabase jwt template?

PR #1232, opened 2024-07-05
Update Client Backend SDK docs to match BAPI docs

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

Part of the work for checking that our /backend/* reference docs were up to date ([due to this linear ticket](https://linear.app/clerk/issue/DOCS-6039/check-that-our-docsreferencesbackend-pages-match-our-backend-api))

I added each pages' corresponding `clerk/javascript` file in a comment on at the top of the page. Adding these links will make it easy to find where these methods exist in the `clerk/javascript` repo when updating their respective pages.

I updated pages accordingly to reflect the BAPI docs.

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -2232,7 +2232,7 @@
                           "href": "/docs/references/backend/types/paginated-resource-response"
                         },
                         {
-                          "title": "Backend Client object",
+                          "title": "Backend `Client` object",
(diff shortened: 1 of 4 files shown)


JacobMGEvans (approved):
The Backend Client page looks good and the table is solid, nothing jumps out at me 😄

PR #1241, opened 2024-07-08
Update explanation of "active organizations"

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1241/organizations/overview#active-organization

### Explanation:
[The linear ticket](https://linear.app/clerk/issue/DOCS-8865/clarify-organization-account-wording) reads:
> On [the organizations overview page](https://clerk.com/docs/organizations/overview#active-organization), there are multiple mentions of "personal accounts" and "organization accounts" - while I get what we're after here, we had a prospective customer get confused by this and think that there were different account types - user accounts and org accounts, and both operated independently. This isn't the case, there are only user accounts, and users belong to organizations. I think we could probably clarify the wording here to make it clear how this works.

### This PR:

- Updates the explanation of active organizations

### DIFF
--- a/docs/organizations/overview.mdx
+++ b/docs/organizations/overview.mdx
@@ -82,7 +82,7 @@ The easiest way to allow users to set an organization as active is to use Clerk'
 
 You can also use the `setActive()` method, which is available on the [`Clerk`](/docs/references/javascript/clerk/session-methods#set-active) object and is returned by the [`useOrganizationList()`](/docs/references/react/use-organization-list) hook.
 
-If you would like to hide personal workspaces and require users to set an organization as active, see the [Force organizations](/docs/guides/force-organizations) guide.
+If you would like to hide personal workspaces and require users to always have an organization set as active, see the [Force organizations](/docs/guides/force-organizations) guide.
 
 ## Create an organization
 
--- a/docs/organizations/overview.mdx
+++ b/docs/organizations/overview.mdx
@@ -74,11 +74,15 @@ Membership requests are requests from users who want to join an organization. A 
 
 ## Active organization
 
-By default, when a user signs in to a Clerk-powered application, there is no active organization set. Even if they are a member of only one organization, they must explicitly set it as active.
+When a user is a member of an organization, they can switch between their personal workspace and an organization workspace. The organization workspace that a user is currently viewing is called the **active organization**. The active organization determines which organization-specific data the user can access and which roles and permissions they have within the organization.
+
+By default, when a user initially signs in to a Clerk-powered application, they are signed in to their personal workspace and no active organization is set. Even if they are a member of only one organization, they must explicitly set it as active.
 
 The easiest way to allow users to set an organization as active is to use Clerk's [`<OrganizationSwitcher />`](/docs/components/organization/organization-switcher) component.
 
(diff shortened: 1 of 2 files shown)


panteliselef (approved):

docs/organizations/overview.mdx
```diff
@@ -74,20 +74,12 @@ Membership requests are requests from users who want to join an organization. A
 
 ## Active organization
 
-There are two types of accounts that a single user can have in a Clerk application: a personal account and an organization account. A personal account is the default account type, and it is created when a user signs up for your application. An organization account is created when a user creates or joins an organization within your application.
+By default, when a user signs in to a Clerk-powered application, there is no active organization set. Even if they are a member of only one organization, they must explicitly set it as active.
```
royanger, line 77: nit: If a user had signed in, become active in an organization and then signs in again they would still be active in that organization.

Maybe "By default, when a user initially signs up or in to a Clerk-powered application..."?

I get the confusion of the customer who initially reported this page, but I feel like the information that is being removed is good but just poorly worded and confusing. @alexisintech -- maybe we could chat Monday afternoon for 15 minutes and see what we can come up with?

jescalan: This is a good suggestion - adding "initially" does clarify a bit here. We could also shift the framing a little from "there's no active organization set" to "the user has a 'personal workspace' that is set to active, even if they are a member of one or more organizations"

jescalan (changes requested):
This is an improvement but agreed with Roy's comment here that we could still tinker with the framing a little more here to further improve while we're at it!

docs/organizations/overview.mdx
```diff
@@ -74,19 +74,15 @@ Membership requests are requests from users who want to join an organization. A
 
 ## Active organization
 
-There are two types of accounts that a single user can have in a Clerk application: a personal account and an organization account. A personal account is the default account type, and it is created when a user signs up for your application. An organization account is created when a user creates or joins an organization within your application.
+When a user is a member of an organization, they can switch between their personal workspace and an organization workspace. The organization workspace that a user is currently viewing is called the **active organization**. The active organization determines which organization-specific data the user can access and which roles and permissions they have within the organization.
```
royanger, line 77:
```suggestion
When a user is a member of an organization, they can switch between their personal workspace and an organization workspace. The organization workspace that a user is currently viewing is called the **active organization**. The active organization determines which organization-specific data the user can access and which role and related permissions they have within the organization.
```
A user can only have one role, and permissions are tied to that, so just tightening up that part a bit.

docs/organizations/overview.mdx
```diff
@@ -74,19 +74,15 @@ Membership requests are requests from users who want to join an organization. A
...
 
-Users will always have a single personal account. They can have multiple organization accounts if they are members of multiple organizations.
+By default, when a user initially signs in to a Clerk-powered application, they are signed in to their personal workspace and no active organization is set. Even if they are a member of only one organization, they must explicitly set it as active.
```
royanger, line 79:
```suggestion
By default, when a user initially signs in to a Clerk-powered application, they are signed in to their personal workspace and no active organization is set. Even if they are a member of only one organization, they must explicitly set it as active or the application can have logic to set this automatically. 
```

royanger (approved):
Just a couple of minor suggestions here, but otherwise this is much better. 🚀

PR #1245, opened 2024-07-09
Clarifying that organization name *and* slug can be updated.

izaaklauer:
Previously, this clamed that only an org name can be updated.

In fact, the org slug can also be updated! The API: 
https://github.com/clerk/javascript/blob/025ab87c660f65ad0dad8980554f1e21fc6dbce6/packages/types/src/organization.ts#L99-L102

And I have a little proof-of-concept proving that it works.


alexisintech (approved):
you are correct! we even it have it documented https://clerk.com/docs/references/javascript/organization/organization#update

thank you 😸💖

PR #1246, opened 2024-07-09
(/backend/overview) Add info for using the Backend SDK vs with other SDKs

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1246/references/backend/overview

### This PR:

- Splits the guide into two concepts: using the Backend SDK on its own vs. using it with other SDKs
- Adds `options`

### DIFF
--- a/docs/references/backend/overview.mdx
+++ b/docs/references/backend/overview.mdx
@@ -211,11 +211,13 @@ try {
 
 ## `createClerkClient({ options })`
 
-The `createClerkClient()` function requires an `options` object. The following options are available:
+The `createClerkClient()` function requires an `options` object. It is recommended to set these options as [environment variables](/docs/deployments/clerk-environment-variables#api-and-sdk-configuration), if possible, and pass them to the function. For example, you can set the `secretKey` option using the `CLERK_SECRET_KEY` environment variable, and then pass it to the function like this: `createClerkClient({ secretKey: process.env.CLERK_SECRET_KEY })`.
+
+The following options are available:
 
 | Property | Type | Description |
 | --- | --- | --- |
-| `secretKey?` | `string` | The Clerk secret key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
+| `secretKey` (required) | `string` | The Clerk secret key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
 | `jwtKey?`  | `string` | The PEM public key from the **[API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page -> Show JWT public key -> PEM Public Key** section of the Clerk Dashboard. For more information, refer to [Manual JWT verification](/docs/backend-requests/handling/manual-jwt). |
 | `publishableKey?` | `string` | The Clerk publishable key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
 | `domain?` | `string` | The domain for the application. For development, you can pass the `localhost` your application is running on. For example: `localhost:3001` |
--- a/docs/references/backend/overview.mdx
+++ b/docs/references/backend/overview.mdx
@@ -225,7 +225,7 @@ The following options are available:
 | `proxyUrl?` | `string` | The proxy URL from a multi-domain setup. |
 | `sdkMetadata?` | `{ name: string, version: string }` | Metadata about the SDK. |
 | `telemetry?` | `{ disabled: boolean, debug: boolean }` | [Telemetry](/docs/telemetry) configuration. |
-| `userAgent?` | `string` |  |
+| `userAgent?` | `string` | The User-Agent request header passed to the Clerk API. |
 | `apiUrl?` | `string` | The Clerk [Backend API](/docs/reference/backend-api) endpoint. Defaults to `'https://api.clerk.com'`. |
 | `apiVersion?` | `string` | The version passed to the Clerk API. Defaults to `'v1'`. |
 | `audience?` | `string \| string[]` | A string or list of [audiences](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3). |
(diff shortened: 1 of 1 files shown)


docs/references/backend/overview.mdx
```diff
@@ -109,3 +208,22 @@ try {
...
+## `createClerkClient({ options })`
+
+The `createClerkClient()` function requires an `options` object. The following options are available:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| `secretKey?` | `string` | The Clerk secret key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
+| `jwtKey?`  | `string` | The PEM public key from the **[API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page -> Show JWT public key -> PEM Public Key** section of the Clerk Dashboard. For more information, refer to [Manual JWT verification](/docs/backend-requests/handling/manual-jwt). |
+| `publishableKey?` | `string` | The Clerk publishable key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
+| `domain?` | `string` | The domain for the application. For development, you can pass the `localhost` your application is running on. For example: `localhost:3001` |
```
JacobMGEvans, line 221: period at the end of the description for consistency but not necessary
```suggestion
| `domain?` | `string` | The domain for the application. For development, you can pass the `localhost` your application is running on. For example: `localhost:3001`. |
```

JacobMGEvans (approved):

docs/references/backend/overview.mdx
```diff
@@ -109,3 +208,24 @@ try {
   // Error handling
 }
 ```
+
+## `createClerkClient({ options })`
+
+The `createClerkClient()` function requires an `options` object. It is recommended to set these options as [environment variables](/docs/deployments/clerk-environment-variables#api-and-sdk-configuration), if possible, and pass them to the function. For example, you can set the `secretKey` option using the `CLERK_SECRET_KEY` environment variable, and then pass it to the function like this: `createClerkClient({ secretKey: process.env.CLERK_SECRET_KEY })`.
```
alexisintech, line 214:
```suggestion
The `createClerkClient()` function requires an `options` object. It is recommended to set these options as [environment variables](/docs/deployments/clerk-environment-variables#api-and-sdk-configuration) where possible, and then pass them to the function. For example, you can set the `secretKey` option using the `CLERK_SECRET_KEY` environment variable, and then pass it to the function like this: `createClerkClient({ secretKey: process.env.CLERK_SECRET_KEY })`.
```

docs/references/backend/overview.mdx
```diff
@@ -4,97 +4,196 @@ description: The JavaScript Backend SDK exposes Clerk's backend API resources an
 
 # JavaScript Backend SDK
 
-The [JavaScript Backend SDK](https://github.com/clerk/javascript/tree/main/packages/backend) exposes Clerk's backend API resources and low-level authentication utilities **for JavaScript environments**. While the backend SDK is mainly used as a building block for Clerk's higher-level SDKs, it can also be used on its own for advanced flows and custom integrations.
+Clerk's JavaScript Backend SDK exposes [Clerk's Backend API](https://clerk.com/docs/reference/backend-api) resources and low-level authentication utilities **for JavaScript environments**.
 
-Clerk SDKs expose an instance of the Backend SDK for use in server environments.
+For example, if you wanted to get a list of all users in your application, instead of creating a fetch to Clerk's [`https://api.clerk.com/v1/users`](https://clerk.com/docs/reference/backend-api/tag/Users#operation/GetUserList) endpoint, you can use the [`users.getUserList()`](/docs/references/backend/user/get-user-list) method provided by the JavaScript Backend SDK.
```
dimkl, line 9: @alexisintech just to be clear the `users.getUserList()` is not exposed directly from the `@clerk/backend` just like you mention in the docs below.

docs/references/backend/overview.mdx
```diff
@@ -109,3 +208,24 @@ try {
...
+
+The following options are available:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| `secretKey` (required) | `string` | The Clerk secret key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
+| `jwtKey?`  | `string` | The PEM public key from the **[API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page -> Show JWT public key -> PEM Public Key** section of the Clerk Dashboard. For more information, refer to [Manual JWT verification](/docs/backend-requests/handling/manual-jwt). |
+| `publishableKey?` | `string` | The Clerk publishable key from the [API Keys](https://dashboard.clerk.com/last-active?path=api-keys) page in the Clerk Dashboard. |
+| `domain?` | `string` | The domain for the application. For development, you can pass the `localhost` your application is running on. For example: `localhost:3001`. |
```
dimkl, line 223: @panteliselef Could you verify if the `domain` options is only used int multi-domain setup?

dimkl (approved):

docs/references/backend/overview.mdx
alexisintech: it reads "All resource operations are mounted as sub-APIs on the `clerkClient` object. For example, if you would like to get a list of all of your application's users, you can use the `getUserList()` method on the `users` sub-API. You can find the full list of available sub-APIs and their methods in the sidebar."

the paragraph begins with stating resources are mounted as sub-APIs on `clerkClient` so users.getUserList() is mounted on `clerkClient`

is this the section you meant?

docs/references/backend/overview.mdx
panteliselef: Correct, Only in a multi-domain setup, and only for the satellite apps, not the one that is used as the primary.

And i believe this is only needed for handshake to work, so for people that use `@clerk/backend` in order simply to access our Backend API it will not make a difference if they use `domain` or even `proxyUrl`.

PR #1266, opened 2024-07-15
Convert the Remix SPA mode page into a guide for apps without Clerk

anagstef:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1266/references/remix/spa-mode
> - https://clerk.com/docs/pr/1266/quickstarts/remix

### Explanation:
The Remix SPA mode docs page was showcasing how to convert an existing Clerk-Remix project into a SPA mode. Probably, this won't be the case for most people.

### This PR: 

- Creates a guide similar to the Remix quickstart, that will help devs install Clerk into a Remix that uses the SPA mode.

### DIFF
--- a/docs/references/remix/spa-mode.mdx
+++ b/docs/references/remix/spa-mode.mdx
@@ -5,8 +5,24 @@ description: Clerk supports Remix SPA mode out-of-the-box.
 
 # Remix SPA Mode
 
+<TutorialHero
+  framework="remix"
+  beforeYouStart={[
+    {
+      title: "Set up a Clerk application",
+      link: "https://clerk.com/docs/quickstarts/setup-clerk",
+      icon: "clerk",
+    }
+  ]}
+>
+- Install `@clerk/remix`
+- Configure ClerkApp with your Publishable key
+- Update your paths through ClerkApp options
+- Protecting your pages
+</TutorialHero>
+
 <Callout type="info">
-This guide explains how to use Clerk with [Remix in SPA mode](https://remix.run/docs/en/main/guides/spa-mode). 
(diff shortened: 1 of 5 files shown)


docs/references/remix/spa-mode.mdx
```diff
@@ -43,6 +66,52 @@ function App() {
   );
 }
 
+export default ClerkApp(App, {
+  publishableKey: 'pk_test_XYZ',
```
royanger, line 70: Does this not read the `CLERK_PUBLISHABLE_KEY` environment variable the way the SSR version does?

royanger (approved):

panteliselef (approved):

docs/references/remix/spa-mode.mdx
anagstef: @royanger Currently, no. I will research in a later release, if we can leverage Vite's `import.meta.env`.

anagstef: Also, this would require users to use the `VITE_` prefixed variables. 

Do you think it would be confusing?

octoper (approved):
💯

docs/references/remix/spa-mode.mdx
anagstef: @royanger Unfortunately, after investigating a bit, the `import.meta` is not available in CommonJS modules, which we still support for Remix, so it's not an easy win to support this. We will revisit this in the future, maybe when we drop commonjs support in a future major version.

anagstef: I decided to change it and provide a similar code snippet as we do in our React quickstart, where we set the `.env` file and then we suggest the users uses the `import.meta` themselves in their `app/root.tsx` file. I think this is a good middleground for now, and we will revisit in the future.

PR #1267, opened 2024-07-15
Update sidenav for better docs organization

kylemac:
- removes icons for a bit more clarity
- adds section headers
- moves a few items into the "top-level" nav
- moves community SDKs into a dropdown
- adds warning icon to _custom flows_
- shifted 1 or 2 items around (ie. _Testing_ into "Development")

CleanShot 2024-07-15 at 12 50 16@2x


docs/manifest.json
```diff
@@ -246,575 +565,791 @@
...
+                              }
+                            ]
+                          ]
                         }
                       ]
                     ]
                   },
                   {
-                    "title": "Versioning",
+                    "title": "⚠ Custom Flows",
```
LekoArts, line 891: The ⚠ emoji/unicode will render differently (or not at all) and won't fit in with the rest of the icons. So I think you need to use the `icon` property and get that icon created (if not existent already).

Why do you think we need to add this icon there?

panteliselef:
I kinda liked the icons, what's driving this ?

kylemac:
> I kinda liked the icons, what's driving this ?

A desire to create a better visual hierarchy. Without the icons the top-level items like guides and ui components stick out a bit more. 

and personally i feel like the icons created too much visual noise. at best they were ignored visually to focus on the text and at worst they distracted. 

that said they are beautiful freaking icons so I'm also sad to see them go, in that way 😭

docs/manifest.json
royanger: I don't think there is a ton of value with the icon -- I'm not sure what its trying to say to the reader.

royanger:
> > I kinda liked the icons, what's driving this ?
> 
> A desire to create a better visual hierarchy. Without the icons the top-level items like guides and ui components stick out a bit more.
> 
> and personally i feel like the icons created too much visual noise. at best they were ignored visually to focus on the text and at worst they distracted.
> 
> that said they are beautiful freaking icons so I'm also sad to see them go, in that way 😭

Agree that the icons looked great, but I also much prefer the cleanliness of this PR. I think it much clearer and easier to read.

alexisintech (approved):

jescalan (approved):

docs/manifest.json
kylemac: 👍  makes sense. in some of our future designs its a bit more relevant, because we show prominent warnings for all custom flows - telling the user that this is the advanced way to do things and not the best/easiset... but that work isn't part of this speedrun

jescalan: We could also move it under "advanced"...

kylemac: yeah, i like that idea

jescalan: Ok, will shift that and get this merged! 🎉

jescalan:
Requires the redirect PR to merge in tandem: https://github.com/clerk/clerk/pull/573

docs/manifest.json
```diff
@@ -463,169 +729,206 @@
...
-                    "title": "User Components",
+                    "title": "Custom Pages",
+                    "collapse": true,
                     "items": [
                       [
                         {
-                          "title": "`<UserButton />`",
-                          "wrap": false,
-                          "href": "/docs/components/user/user-button"
+                          "title": "User Profile",
```
alexisintech, line 801: is there a reason we removed the component syntax?

```suggestion
                          "title": "`<UserProfile />`",
```

docs/manifest.json
```diff
@@ -463,169 +729,206 @@
...
+                          "href": "/docs/customization/user-profile"
                         },
                         {
-                          "title": "`<UserProfile />`",
-                          "wrap": false,
-                          "href": "/docs/components/user/user-profile"
+                          "title": "Organization Profile",
```
alexisintech, line 805:
```suggestion
                          "title": "`<OrganizationProfile />`",
```

PR #1274, opened 2024-07-17
Update SMS template docs about verification code OTP templates

mzhong9723:
We are switching over to a new SMS provider (Prelude) very soon, and this makes changes to the SMS message body for verification OTP templates a manual process. In addition, the changes possible are very limited.

As a result, we want to update the docs to call this out and tell customers to contact support if they must change it.

### DIFF
--- a/docs/authentication/configuration/email-sms-templates.mdx
+++ b/docs/authentication/configuration/email-sms-templates.mdx
@@ -133,7 +133,3 @@ The following settings can be changed for an SMS template:
 
 - **Name**: a name for the template on the template listing page. It does not affect the outgoing email in any way.
 - **Message**: a text area where you can author the content of the SMS message. The SMS content should fit within a limit of 160 simple GSM-7 or 140 unicode characters. To insert a [variable](#terminology) at the current cursor position, select the corresponding variable badge. You can interpolate a particular metadata key by using the following syntax: `{{user.metadata.foo.bar}}`.
-
-<Callout type="info">
-  Modifications to Verification OTP SMS templates are very limited due to strict regulations. If changing the Reset password or Verification code SMS template is required, please contact Clerk support.
-</Callout>
--- a/docs/authentication/configuration/email-sms-templates.mdx
+++ b/docs/authentication/configuration/email-sms-templates.mdx
@@ -114,7 +114,7 @@ To access the SMS templates:
 
 ### Delivered by Clerk
 
-Out of the box, Clerk will deliver your SMS messagess using its own SMS Gateway. However, if you wish to handle SMS delivery on your own, then you can toggle **Delivered by Clerk** off.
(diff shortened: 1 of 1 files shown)


docs/authentication/configuration/email-sms-templates.mdx
```diff
@@ -133,3 +133,7 @@ The following settings can be changed for an SMS template:
 
 - **Name**: a name for the template on the template listing page. It does not affect the outgoing email in any way.
 - **Message**: a text area where you can author the content of the SMS message. The SMS content should fit within a limit of 160 simple GSM-7 or 140 unicode characters. To insert a [variable](#terminology) at the current cursor position, select the corresponding variable badge. You can interpolate a particular metadata key by using the following syntax: `{{user.metadata.foo.bar}}`.
+
+<Callout type="info">
+  Modifications to Verification OTP SMS templates are very limited due to strict regulations. If changing the Reset password or Verification code SMS template is required, please contact Clerk support.
```
alexisintech, line 138:
```suggestion
  Modifications to **Verification OTP** templates are very limited due to strict regulations. If changing the **Reset password** or **Verification code** template is required, please contact [Clerk support](https://clerk.com/contact/support).
```

alexisintech (approved):

docs/authentication/configuration/email-sms-templates.mdx
jescalan: At the moment, users can request changes from the dashboard without contacting support. If this being removed?

mzhong9723: @jescalan Only for the verification OTP templates for instances using the new SMS provider (we want to switch over almost all of them to). This is because Prelude really only allows a custom suffix of an app name, and even them, it's a manual process.

jescalan: So would we still be sending reset password and verification code through twilio? If that is the case and we're keeping the dashboard "request changes" flow for this, we may be able to just remove the part about reaching out to support right?

mzhong9723: Reset password code and verification code will all be through Prelude, which we can't support "request changes" for in the dashboard. The other templates grouped under "account notifications" will still be through Twilio, and we can support request changes for.

mzhong9723: I guess my original goal with adding this section to the docs was to explain why the two OTP templates didn't have a request change button in the dashboard, but maybe it's causing confusion. What do you think would be the best?

mzhong9723: synced with the team in slack: https://clerkinc.slack.com/archives/C06TXR2HT6U/p1721317557564899

We will be removing any mention of potentially changing verification OTP templates. I will merge this PR only with the typo fix on line 117.

PR #1288, opened 2024-07-18
Add "Team Roles & Permissions" page 

LauraBeatris:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1288/dashboard/team-roles-and-permissions

### Explanation:

Introduces documentation for Dashboard roles, detailing each persona and permissions as a team guide. 

CleanShot 2024-08-06 at 15 39 28

### DIFF
--- a/docs/dashboard/team-roles-and-permissions.mdx
+++ b/docs/dashboard/team-roles-and-permissions.mdx
@@ -5,19 +5,12 @@ description: Learn about the different roles available for team members on a Cle
 
 Clerk Dashboard offers various roles to manage team members access. Different roles ensure that team members have the appropriate level of access according to their responsibilities.
 
-### Admin Role
+[TODO - Add screenshot containing dropdown with roles / OR containing the Dashboard UI protected]
 
-Admins have full access and control over all team and applications settings.
+- **Admin**: Full access and control over all team and project settings.
+- **Member**: Can view content but cannot make changes.
+- **Developer**: Integrates Clerk's authentication, manages user data, customizes the user experience, maintains security, and collaborates with other team members.
+- **Support**: Integrates Clerk's authentication, manages user and organization data, impersonates users, and bans users.
+- **Viewer**: Can view content but cannot make changes.
 
-### Member Role
-
-Viewers can view content but cannot make changes.
-
-### Developer Role
-
-Developers integrate Clerk's authentication, manage user and organizations data.
-
-### Support Role
-
-Support has access to impersonate users, and ban users, with less autonomy to
-integrate Clerk's authentication.
+Refer to the permissions that each role provides mapped to Dashboard resources:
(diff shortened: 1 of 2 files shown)


docs/manifest.json
```diff
@@ -1212,7 +1212,20 @@
                 ]
               ]
             },
-            { "title": "Core 1 Documentation", "href": "/docs/core-1", "icon": "link" }
+            { "title": "Core 1 Documentation", "href": "/docs/core-1", "icon": "link" },
+            {
+              "title": "Dashboard",
```
LauraBeatris, line 1217: 💬 Should we create another section specific to Dashboard features? I haven't found one with similar purposes.

docs/dashboard/team-roles-and-permissions.mdx
```diff
@@ -0,0 +1,36 @@
...
+| **Organization settings** | Read | ✓ | ✓ | ✓ | ✓ |
+| **Organization settings** | Manage | ✓ | | ✓ | |
+| **Applications** | Manage | ✓ | | | |
+| **Developer Resources** | Manage | ✓ | | ✓ | |
+| **API Keys** | Manage | ✓ | | | |
+| **Billing** | Read | ✓ | ✓ | ✓ | ✓ |
+
+## How to Change a Team Member's Role
+
+To change a team member's role, you must be an **Admin**. Follow these steps in the `<OrganizationProfile />` component:
```
LauraBeatris, line 29: I think we shouldn't mention the SDK component here on a Dashboard walkthrough, it's confusing to guide the users to look for it.

NicolasLopes7 (approved):

alexisintech:
So we call this "Team Roles and Permissions" but the feature in the Dashboard is not called Teams - its called Organizations. That's something to think about.

In the introductory paragraph, I'd like to see an explanation of this feature - something like "In the **Clerk Dashboard**, you can have applications for your personal account or for an organization. Organizations allow you to manage multiple applications and team members. Each team member can have a different role with different permissions. Different roles and permissions ensure that team members have the appropriate level of access according to their responsibilities.

For example, you may have an organization where the **Admin** role has full access to all resources, while the **Member** role has read-only access to some resources. This way, you can control who can manage your users and specific settings for your applications."

Also, for the table that maps the Dashboard resources, "Developer resources" is pretty vague. I know it's tough to get this table right - we don't want it too big and cumbersome to keep up with, but we want to explain the different roles/permissions well.

LauraBeatris:
> So we call this "Team Roles and Permissions" but the feature in the Dashboard is not called Teams - its called Organizations. That's something to think about.
> 
> In the introductory paragraph, I'd like to see an explanation of this feature - something like "In the **Clerk Dashboard**, you can have applications for your personal account or for an organization. Organizations allow you to manage multiple applications and team members. Each team member can have a different role with different permissions. Different roles and permissions ensure that team members have the appropriate level of access according to their responsibilities.
> 
> For example, you may have an organization where the **Admin** role has full access to all resources, while the **Member** role has read-only access to some resources. This way, you can control who can manage your users, specific settings for your applications, and more."
> 
> Also, for the table that maps the Dashboard resources, "Developer resources" is pretty vague. I know it's tough to get this table right - we don't want it too big and cumbersome to keep up with, but we want to explain the different roles/permissions well.

@alexisintech Thanks Alexis! That all makes sense to me. We'll make changes next week once we have more feedback on the responsibilities of each user role, cause this will determine how the table should be structured here.

kylemac:
@alexisintech it's not approved yet. i would expect it in the next 2 weeks

LauraBeatris:
This project got paused for a while, as we're still defining the priorities on when it's going to be back. I'll close this PR for now as a lot of the decisions here might change later.

PR #1298, opened 2024-07-19
Refactor: Update order of lib managers in the CodeBlockTabs on the add-react-router reference page

kduprey:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1298/references/react/add-react-router

### Explanation:
In order to improve DX within our docs, the lib managers order should be consistent from page to page, to reduce confusion and maintain patterns across the docs.

### This PR:

- Update the add-react-router page codeblock options prop to have an array of `['npm', 'yarn', 'pnpm']` to match the same order as the previous React quick start guide's codeblock order

### DIFF
--- a/docs/references/react/add-react-router.mdx
+++ b/docs/references/react/add-react-router.mdx
@@ -41,7 +41,7 @@ Learn how to add React Router to your application using React Router's new Data 
 
 React Router's `react-router-dom` is a mature, battle tested routing package for React that gives you many options. As it is the most popular routing option, it will be used for this guide.
 
-<CodeBlockTabs type="installer" options={["npm", "pnpm", "yarn" ]}>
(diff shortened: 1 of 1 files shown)


royanger (approved):
🚀

LekoArts (changes requested):
When changing the `options` prop you'll also need to change the order of the code blocks inside the `CodeBlockTabs` component. When visiting the preview at https://clerk.com/docs/pr/1298/references/react/add-react-router you'll see that after selecting `yarn` you'll see the pnpm instructions.

Swap the two code blocks and it'll be fine 👍

kduprey (commented):
Addressed in last commit

alexisintech (approved):

PR #1343, opened 2024-07-30
update force-organizations.mdx

kylemac:
when using the server `auth()` it comes back as `undefined` and not `null` (when using the client-side `useAuth()` it comes back as `null`)

check me on this please :)


royanger (approved):
🚀

docs/guides/force-organizations.mdx
```diff
@@ -86,7 +86,7 @@ This guide will be written for Next.js applications using App Router, but the sa
       export default function Layout() {
         const { orgId } = auth();
 
-        const hasActiveOrg = orgId !== null;
+        const hasActiveOrg = orgId !== undefined;
```
panteliselef, line 89: we need to check against both, null means that the user is signedOut, undefined means that the user is signed in but the don't have an active org selected.

https://github.com/clerk/javascript/blob/80e64773135865434cf0e6c220e287397aa07937/packages/backend/src/tokens/authObjects.ts#L50
```suggestion
        const hasActiveOrg = !!orgId
```

alexisintech (approved):
format the file and you should be good to go😸💖

kylemac:
@alexisintech the lints seems to indicate the issue is on `docs/integrations/databases/fauna.mdx` which is unrelated to this PR so it must have snuck in elsewhere. that said, i fixed it in a subsequent commit. i'm gonna pass the baton to you. you can feel free to merge this but i just want you to peek at the prettier changes on the unrelated file before i YOLO it into prod

kylemac:
okay, force push incoming

PR #1345, opened 2024-07-31
feat(deployments): Migrate from Cognito

thiskevinwang:
# Description

This adds a new page for our Cognito password migrator

Preview page @ https://clerk.com/docs/pr/1345/deployments/migrate-from-cognito

This unblocks the changelog entry in https://github.com/clerk/clerk/pull/532


colinclerk:
* Use `@clerk/backend` instead of `@clerk/clerk-sdk-node`.  sdk-node is just a re-export of backend, but we've kept it around for backward compat reasons.  The code is all isomorphic for cross-runtime compat, though, so we prefer backend for guides going forward.
* I don't think I understand the "Caution" callout. I would expect guidance on setting this for my existing user pool instead of a link to create a new user pool.

kylemac:
i think the right idea is to move this from "guides" and put it the "deployments and migrations ->  migrations" section

thiskevinwang:
> * Use `@clerk/backend` instead of `@clerk/clerk-sdk-node`.  sdk-node is just a re-export of backend, but we've kept it around for backward compat reasons.  The code is all isomorphic for cross-runtime compat, though, so we prefer backend for guides going forward.
> * I don't think I understand the "Caution" callout. I would expect guidance on setting this for my existing user pool instead of a link to create a new user pool.

@colinclerk 
1. Good suggestion, I'll update to use `@clerk/backend`
2. The caution callout mentions that you can create a new **client** for your user pool. This is not to be confused with the user pool itself. Do you think this is misleading and should be removed? Or should we keep it and reword to be less confusing?

thiskevinwang:
> i think the right idea is to move this from "guides" and put it the "deployments and migrations -> migrations" section

@kylemac done. The page is now at `/docs/deployments/cognito-password-migrator`

CleanShot 2024-07-31 at 16 24 26@2x

kylemac:
How about just generically "migrate from Cognito" and then inside there we share about the migrator feature

thiskevinwang:
> How about just generically "migrate from Cognito" and then inside there we share about the migrator feature

@kylemac Sounds good to me. That should be an easy title + slug change. The document is pretty un-opinionated w.r.t it being about Migrating from Cognito vs. the password migrator feature. 👍

colinclerk (approved):

PR #1359, opened 2024-08-01
Rename Expo read-session-user-data doc; remove unused import

alexisintech:
> [!WARNING]
> Need [clerk/clerk#549](https://github.com/clerk/clerk/pull/549) merged for redirects

### Explanation:

Needed to remove an unused import from an expo doc. Also found that the /expo/expo-read-session-user-data doc doesn't /expo twice.

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -1728,7 +1728,7 @@
                   },
                   {
                     "title": "Read session and user data",
-                    "href": "/docs/references/expo/expo-read-session-user-data"
+                    "href": "/docs/references/expo/read-session-user-data"
                   },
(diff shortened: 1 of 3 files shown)


jescalan (approved):

jescalan (changes requested):
Actually hold up don't we now need to add a redirect for the old path

alexisintech:
> Actually hold up don't we now need to add a redirect for the old path

https://github.com/clerk/clerk/pull/549

jescalan (approved):
Approved contingent on redirect PR merging 😁

PR #1409, opened 2024-08-07
Fixes typo in the get-token page

davidgomes:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### This PR:

-


alexisintech:
`clerkClient` is called as a function if you are using Next.js - calling it as `clerkClient` is being deprecated. check your console for the deprecation warning when using it

PR #1410, opened 2024-08-07
(/clerk-provider) Update info about dynamic rendering

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1410/components/clerk-provider

### Explanation:

[User feedback ticket](https://linear.app/clerk/issue/DOCS-8960/componentsclerk-provider-add-info-about-dynamic-rendering) reads:
> The docs for this page are extremely lacking and don’t cover how it’s possible to still use ClerkProvider as a client component which doesn’t force dynamic rendering. Needs to be added and explained deeper.

### This PR:

- Updates the information to make usage of the provider clear; adds info on using it in client components

### DIFF
--- a/docs/components/clerk-provider.mdx
+++ b/docs/components/clerk-provider.mdx
@@ -13,9 +13,11 @@ The `<ClerkProvider>` component must be added to your React entrypoint.
 
 <Tabs type="framework" items={["Next.js", "React"]}>
   <Tab>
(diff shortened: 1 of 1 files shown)


jescalan (approved):

BRKalow (approved):
🙏

PR #1412, opened 2024-08-08
docs: add step to custom sign up pages and routes about adding routes to be ignored by middleware

kduprey:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1412

### Explanation:

`clerkMiddleware()` does not make routes public, so if a customer is making most/all routes private they can mistakenly make the sign in and up routes private. Customers following this guide and then making all/most routes public run into private auth routes.

### This PR:

- Adds step to custom sign up pages and routes about adding routes to be ignored by middleware

### DIFF
--- a/docs/references/nextjs/custom-signup-signin-pages.mdx
+++ b/docs/references/nextjs/custom-signup-signin-pages.mdx
@@ -62,9 +62,15 @@ If Clerk's prebuilt components don't meet your specific needs or if you require 
     ```
   </CodeBlockTabs>
 
-  ### Add custom /sign-in and /sign-up routes to middleware
+  ### Make the sign-up and sign-in routes public
 
-  `clerkMiddleware()` does not explicitly make `/sign-in` and `/sign-up` routes public, so we need to be sure they will be public for people to access the custom routes to sign in and sign up. In your `middleware.ts` file, update the following code to explicitly allow the `sign-in` and `sign-up` pages to be public. This will allow users to access the custom sign-in and sign-up pages, while still protecting the rest of your application's routes.
+  By default, `clerkMiddleware()` makes all routes public. This step is specifically for Next.js applications that have configured `clerkMiddleware()` to make [all routes protected](/docs/references/nextjs/clerk-middleware#protect-all-routes). If you have not configured `clerkMiddleware()` to protect all routes, you can skip this step.
+
+  To make the sign-up and sign-in routes public:
+
+  - Navigate to your `middleware.ts` file.
+  - Create a new [route matcher](/docs/references/nextjs/clerk-middleware#create-route-matcher) that matches the sign-up and sign-in routes.
+  - Create a check to see if the current route is a public route. If it is not a public route, use [`auth().protect()`](/docs/references/nextjs/auth#protect) to protect the route.
 
   ```tsx {{ prettier: false, filename: 'middleware.ts' }}
   import { clerkMiddleware } from "@clerk/nextjs/server";
--- a/docs/references/nextjs/custom-signup-signin-pages.mdx
+++ b/docs/references/nextjs/custom-signup-signin-pages.mdx
@@ -62,6 +62,22 @@ If Clerk's prebuilt components don't meet your specific needs or if you require 
     ```
   </CodeBlockTabs>
 
+  ### Add custom routes to middleware
+
+  `clerkMiddleware()` does not make routes public by default, so we need to make them public for people to access the custom routes to sign in and sign up. In your `middleware.ts` file, update the following code to allow the `sign-in` and `sign-up` pages to be public.
+
+  ```tsx {{ prettier: false, filename: 'middleware.ts' }}
+  import { clerkMiddleware } from "@clerk/nextjs/server";
+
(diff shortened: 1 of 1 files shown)


docs/references/nextjs/custom-signup-signin-pages.mdx
```diff
@@ -62,6 +62,22 @@ If Clerk's prebuilt components don't meet your specific needs or if you require
     ```
   </CodeBlockTabs>
 
+  ### Add custom routes to middleware
+
+  `clerkMiddleware()` does not make routes public by default, so we need to make them public for people to access the custom routes to sign in and sign up. In your `middleware.ts` file, update the following code to allow the `sign-in` and `sign-up` pages to be public.
```
alexisintech, line 67: clerkMiddleware() does make routes public by default. 
let me know if you want to call on this or pair with roy on it - the concept for this section isn't totally correct

royanger: The key concept, as I mentioned to @kduprey just now in DMs, is that `authMiddleware()` makes the sign-in and a sign-up routes public. It had to do this as it would protect everything by default. My linear comment was a little unclear.

`clerkMiddleware()` makes everything public and does not explicitly make sign-in and sign-up routes public. This can lead to customers who will make all or most routes private and in doing so make their sign-in and sign-up routes private. The linear is just about calling this potential foot gun out and showing how to avoid it.

We can do a call, but I think after DMs with Kenton we're aligned.

docs/references/nextjs/custom-signup-signin-pages.mdx
```diff
@@ -62,6 +62,22 @@ If Clerk's prebuilt components don't meet your specific needs or if you require
     ```
   </CodeBlockTabs>
 
+  ### Add custom /sign-in and /sign-up routes to middleware
+
+  `clerkMiddleware()` does not explicitly make `/sign-in` and `/sign-up` routes public, so we need to be sure they will be public for people to access the custom routes to sign in and sign up. In your `middleware.ts` file, update the following code to explicitly allow the `sign-in` and `sign-up` pages to be public. This will allow users to access the custom sign-in and sign-up pages, while still protecting the rest of your application's routes.
```
royanger, line 67: nit: we don't use 'we' in the docs as per the Style guide. The reader is 'you' and Clerk is Clerk. 

This is getting better, but I think its still missing the when/why. Many customers will not need this, but this reads like all/most people will need it.

This needs something covering that this is for when all (or maybe most) routes are protected, so that users can still sign-up and sign-in.

royanger (changes requested):

alexisintech (approved):

alexisintech:
thank you for being on top of this <3

kduprey:
> thank you for being on top of this <3

Such a small thing, figured I'd get it done :)

docs/references/nextjs/custom-signup-signin-pages.mdx
```diff
@@ -62,6 +62,28 @@ If Clerk's prebuilt components don't meet your specific needs or if you require
     ```
   </CodeBlockTabs>
 
+  ### Make the sign-up and sign-in routes public
+
+  By default, `clerkMiddleware()` makes all routes public. This step is specifically for Next.js applications that have configured `clerkMiddleware()` to make [all routes protected](/docs/references/nextjs/clerk-middleware#protect-all-routes). If you have not configured `clerkMiddleware()` to protect all routes, you can skip this step.
```
royanger, line 67: minor nit: Any reason we're saying Next.js here? That should be implied and feels slightly redundant. This isn't enough to be a blocker though.

docs/references/nextjs/custom-signup-signin-pages.mdx
```diff
@@ -62,6 +62,28 @@ If Clerk's prebuilt components don't meet your specific needs or if you require
...
+
+  To make the sign-up and sign-in routes public:
+
+  - Navigate to your `middleware.ts` file.
+  - Create a new [route matcher](/docs/references/nextjs/clerk-middleware#create-route-matcher) that matches the sign-up and sign-in routes.
```
royanger, line 72:
```suggestion
  - Create a new [route matcher](/docs/references/nextjs/clerk-middleware#create-route-matcher) that matches the sign-up and sign-in routes, or you can add them to an existing route matcher that is making routes public.
```

docs/references/nextjs/custom-signup-signin-pages.mdx
```diff
@@ -62,6 +62,28 @@ If Clerk's prebuilt components don't meet your specific needs or if you require
...
+  - Create a check to see if the current route is a public route. If it is not a public route, use [`auth().protect()`](/docs/references/nextjs/auth#protect) to protect the route.
```
royanger, line 73:
```suggestion
  - Create a check to see if the user's current route is a public route. If it is not a public route, use [`auth().protect()`](/docs/references/nextjs/auth#protect) to protect the route.
```

royanger (approved):
Couple of minor suggestions, but otherwise the this looks good.

PR #1424, opened 2024-08-09
Add strict-dynamic CSP docs

jescalan:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1424/security/clerk-csp#implementing-a-strict-dynamic-csp

### Explanation:

Clerk now support strict-dynamic CSPs thanks to https://github.com/clerk/javascript/pull/3858 - this PR updates the docs to explain how to configure your app to take advantage of this.


docs/security/clerk-csp.mdx
```diff
@@ -54,5 +54,107 @@ module.exports = nextConfig;
...
+    <ClerkProvider nonce={headers().get("x-nonce")}>
+      <html lang="en">
+        <body>{children}</body>
+      </html>
+    </ClerkProvider>
+  );
+}
+```
+
+If you are using Next.js and your layout file is rendered on the server (as is the case with the example above), our SDK will automatically load the nonce into ClerkProvider, and the explicit passing of the nonce would be unnecessary. However, if you are rendering the provider on the client, you will need to explicitly pass the nonce value to ClerkProvider.
```
BRKalow, line 160:
```suggestion
If you are using Next.js and your layout file is rendered on the server (as is the case with the example above), Clerk's SDK will automatically read the `nonce` from the request and pass it into `<ClerkProvider>`. If you are rendering the provider on the client, you will need to explicitly pass the `nonce` value to the client and into `<ClerkProvider>`.
```

docs/security/clerk-csp.mdx
```diff
@@ -54,5 +54,107 @@ module.exports = nextConfig;
...
+  import { headers } from "next/headers";
+
+  export default function Page() {
+    const nonce = headers().get("x-nonce");
+
+    return <p>{nonce}</p>;
+  }
+```
+
+If you're using one of Clerk's React-based SDKs, in order for Clerk to load correctly, the nonce value must also be passed to Clerk's `ClerkProvider` component. This can be done by passing the nonce value as a `nonce` prop to the `ClerkProvider` component. For example:
```
BRKalow, line 142:
```suggestion
If you're using one of Clerk's React-based SDKs, in order for Clerk to load correctly, the nonce value must also be passed to Clerk's `<ClerkProvider>` component. This can be done by passing the nonce value as a `nonce` prop to the `<ClerkProvider>` component. For example:
```

docs/security/clerk-csp.mdx
```diff
@@ -54,5 +54,107 @@ module.exports = nextConfig;
...
+  export default clerkMiddleware((auth, request) => {
+    return applyCsp(request);
+  });
+
+  function applyCsp(request) {
+    // create a randomly generated nonce value
+    const nonce = Buffer.from(crypto.randomUUID()).toString("base64");
+
+    // format the CSP header
+    const cspHeader = `
```
BRKalow, line 77: The CSP directives here differ from the one earlier in the page, let's make sure they're consistent (considering the usage of `strict-dynamic`)

BRKalow (changes requested):

docs/security/clerk-csp.mdx
BRKalow: @jescalan this is missing `frame-src`

docs/security/clerk-csp.mdx
```diff
@@ -54,5 +55,103 @@ module.exports = nextConfig;
...
+      default-src 'self';
+      script-src 'self' 'strict-dynamic' 'nonce-${nonce}' https: http: ${
+      process.env.NODE_ENV === "production" ? "" : `'unsafe-eval'`
+    };
+      img-src 'self' https://img.clerk.com;
+      worker-src 'self' blob:;
+      style-src 'self' 'unsafe-inline';
+      form-action 'self';
```
BRKalow, line 86:
```suggestion
      form-action 'self';
      frame-src 'self' https://challenges.cloudflare.com;
```

jescalan: With strict-dynamic, whitelisting domains doesn't work at all. I could just add `frame-src 'self'` here i guess but it didn't seem necessary

BRKalow: Is this the case for other directives that aren't script-src too? 🤔

jescalan: Did some research - you're right, it only applies to script-src. Modified this to mirror in frame-src. Thanks for catching this!

BRKalow (approved):
👏

PR #1471, opened 2024-08-20
fix: Update handshake in SDK Development spec

LekoArts:
Updates the SDK types table to correctly note handshake support only for fullstack SDKs. Previously it was marked for backend-only and fullstack.


anagstef (approved):
LGTM

PR #1478, opened 2024-08-21
Fix `biometricType` type typo

bradlc:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1478/references/expo/use-local-credentials

I believe that `face-recognition` and `fingerprint` are meant to be string values.


panteliselef (approved):
Correct

PR #1514, opened 2024-09-05
Remove OAuth references from custom SAML connection docs

LauraBeatris:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1514

### Explanation:

We are mentioning "OAuth" in a couple of places and code snippets from the custom SAML connections docs. I believe it was previously copied from the custom OAuth connections docs.

I've removed those references to use SAML nomenclature.

### DIFF
--- a/docs/custom-flows/saml-connections.mdx
+++ b/docs/custom-flows/saml-connections.mdx
@@ -71,7 +71,6 @@ The following example shows two files:
       import { AuthenticateWithRedirectCallback } from '@clerk/nextjs'
 
       export default function SSOCallback() {
-        console.log('testing')
         // Handle the redirect flow by rendering the
         // prebuilt AuthenticateWithRedirectCallback component.
         // This is the final step in the custom SAML flow.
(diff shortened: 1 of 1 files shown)


docs/custom-flows/saml-connections.mdx
```diff
@@ -165,12 +163,10 @@ The following example shows how to handle these cases in your sign-in flow. The
         } else {
           // If the user has an account in your application
           // and has an SAML account connected to it, you can sign them in.
-          signInWith(e)
```
LauraBeatris, line 168: `signInWith` was a bit vague here, this also seems to be copied from the OAuth connections code, where you would call like `signInWith('oauth_google')`

docs/custom-flows/saml-connections.mdx
```diff
@@ -73,7 +71,6 @@ The following example shows two files:
       import { AuthenticateWithRedirectCallback } from '@clerk/nextjs'
 
       export default function SSOCallback() {
-        console.log('testing')
```
LauraBeatris, line 76: I don't think we should have a `console.log` here, even if it's to debug at first, this can be added by the dev.

alexisintech (approved):

PR #1515, opened 2024-09-05
add instructions to sign up on localhost

mwickett:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

We don't currently explicitly tell users to sign up on their localhost, which creates a gap in metrics between users who run locally, and create their first user (which is what we consider to be an `install` event)

### This PR:

- Adds explicit instructions as the last step of our quickstarts - the goal here is to make it clear that they should sign up to create their first user.

[Companion PR](https://github.com/clerk/dashboard/pull/3571) making the same updates for Dashboard

### DIFF
--- a/docs/quickstarts/astro.mdx
+++ b/docs/quickstarts/astro.mdx
@@ -161,7 +161,7 @@ Clerk's [Astro SDK](/docs/references/astro/overview) provides a set of component
   </SiteLayout>
   ```
 
-  ### Sign up to create your first user
+  ### Create your first user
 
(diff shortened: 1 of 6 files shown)


mwickett:
[Some discussion](https://github.com/clerk/dashboard/pull/3571#pullrequestreview-2284331770) in the Dashboard PR on simplifying these instructions further.

mwickett:
Ok, @alexisintech I think this is ready for a review. Thanks!

alexisintech (approved):
interested to see the metrics on this !

PR #1535, opened 2024-09-13
Update Staging guides: support subdomains over buying separate domains

alexisintech:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

Clerk now supports multiple applications being hosted on the same domain. Users no longer need to purchase a separate domain to set up a "staging" application - they can use subdomains.
Relevant ticket: https://linear.app/clerk/issue/DOCS-9206/update-staging-environment-docs

### This PR:

-

### DIFF
--- a/docs/deployments/set-up-staging.mdx
+++ b/docs/deployments/set-up-staging.mdx
@@ -3,9 +3,9 @@ title: Set up a staging environment with Clerk
 description: Learn how to set up a staging environment with Clerk authentication.
 ---
 
-Staging environments enable you to internally test and demo changes to your application or website before deploying them to production. Currently, Clerk only offers **Development** and **Production** instances. Official support for **Staging** instances is still on [Clerk's roadmap](https://feedback.clerk.com/roadmap/de417dd1-fa2e-4997-868f-4c9248027e7d). However, you can set up a "staging environment" by creating a separate Clerk application with a separate domain.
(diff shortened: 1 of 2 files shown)


docs/deployments/set-up-staging.mdx
```diff
@@ -3,27 +3,27 @@ title: Set up a staging environment with Clerk
 description: Learn how to set up a staging environment with Clerk authentication.
 ---
 
-Staging environments enable you to internally test and demo changes to your application or website before deploying them to production. Currently, Clerk only offers **Development** and **Production** instances. Official support for **Staging** instances is still on [Clerk's roadmap](https://feedback.clerk.com/roadmap/de417dd1-fa2e-4997-868f-4c9248027e7d). However, you can set up a "staging environment" by creating a separate Clerk application with a separate domain.
+Staging environments enable you to internally test and demo changes to your application or website before deploying them to production. Currently, Clerk only offers **Development** and **Production** instances. Official support for **Staging** instances is still on [Clerk's roadmap](https://feedback.clerk.com/roadmap/de417dd1-fa2e-4997-868f-4c9248027e7d). However, you can set up a "staging environment" by creating a subdomain for a separate Clerk application.
```
nikosdouvlis, line 6: Hey!
I noticed that the feedback.clerk.com link simply open our productlane page but does not show a specific entry - I guess that the entry with id `de417dd1-fa2e-4997-868f-4c9248027e7d` got deleted at some point. 

Does anyone remember what that product lane was about? We still don't have support for staging instances so there *should* be an entry for it but I can't find it.

royanger: I'm pretty sure this was for multiple instances, but I don't have anything regarding it in my browser history. @kylemac or @jescalan might know more here. I don't see anything in Product Lane's public facing side for multiple instances/staging instances.

kylemac: ah this is a bit of a mystery we should figure out. we linked to the same "staging instances" item in the changelog too - but its been removed from productlane somehow

jescalan: Got this re-added!

PR #1536, opened 2024-09-13
docs-9205:  change appearance prop color from red to purple

victoriaxyz:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1536

### This PR:

- changes the appearance prop color from red to purple so the code example matches Clerk's colors

PR #1537, opened 2024-09-13
Change appearance prop examples to not use red as primary color & use black as text color

victoriaxyz:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1537

### This PR:

- changes the `ClerkProvider` text color from `white` to `black`
- previous PR changed primary color from `red` to `purple`: https://github.com/clerk/clerk-docs/pull/1536#discussion_r1759323247 ; this PR changes `purple` to `blue` as purple is not an accepted value

Screenshot 2024-09-16 at 16 24 08

Screenshot 2024-09-16 at 16 24 18

### DIFF
--- a/docs/customization/themes.mdx
+++ b/docs/customization/themes.mdx
@@ -607,7 +607,7 @@ In the following example, the primary color of the themes are customized.
               variables: { colorPrimary: 'blue' },
               signIn: {
                 baseTheme: [shadesOfPurple],
-                variables: { colorPrimary: 'blue' },
+                variables: { colorPrimary: 'green' },
               },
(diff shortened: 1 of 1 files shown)


alexisintech:
purple cannot be used - let's use blue instead.
also, there are some places where there is still `colorPrimary: 'red'` so make sure to update all of those instances

and make sure to test these different examples please :) I should have done it on the last PR - that's my bad too 🙈

Screenshot 2024-09-16 at 10 07 00

alexisintech (approved):

PR #1540, opened 2024-09-14
Added note about logout

ntotten:
Logout is not supported when using clerk OAuth. I added a note indicating this.

https://discord.com/channels/856971667393609759/1283094343779090443/1283094343779090443

> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

It notes that OAuth logout is not supported to save other people time.

### This PR:

Adds a FAQ question and answer that states logout is not support with Clerk's OAuth endpoint.

### DIFF
--- a/docs/advanced-usage/clerk-idp.mdx
+++ b/docs/advanced-usage/clerk-idp.mdx
@@ -105,3 +105,7 @@ Access tokens expire after 2 hours, and refresh tokens expire after 3 days.
 ### Is OpenID Connect supported?
 
 At this time, OIDC is not supported.
+
+### How can I log a user out of the OAuth session?
+
+Clerk does not support logging a user out when using this OAuth endpoint. The only option is to wait until the tokens expire.


royanger:
@alexisintech -- Clerk as an OAuth provider does not issue an auth token, so there is no need for a logout endpoint. I've communicated that in the Discord thread. IMO this should be closed and not merged. The FAQ and the warning at the top of the page make it clear this is not an authentication token.

> This feature will not return a Clerk JWT and does not provide authentication. Please see the [FAQ](https://clerk.com/docs/advanced-usage/clerk-idp#how-can-the-access-tokens-obtained-from-the-token-url-be-used) below for more information.

ntotten:
Here are more details on repro steps (from discord thread). I recommend this be noted explicitly as this is a complex issue that many users will not understand without investing significant time. The not about the Clerk JWT IMO is not sufficient to explain the problem.


Setup:
- Clerk configured with Google, etc. as identity providers (essentially the default Clerk configuration)
- A web app (SPA) configured to use Clerk OAuth authorization. This app stores its own session after completing the auth from clerk.

Repro Steps

1. I open the app, lets call it `portal.example.com`, I click Login
2. `portal.example.com` redirects me to to Clerk's OAuth endpoint, say `my-domain.clerk.accounts.dev/oauth/authorize`
3. The user selects the downstream IdP to use, lets pick Google
4. The user authorizes and consents with Google, they are redirected back to clerk, then redirected back to `portal.example.com`'s callback endpoint
5. `portal.example.com` completes the OAuth exchange, gets a JWT token, sets its own session (memory, local storage, cookie, whatever)

The state after these steps is:
1. The user is authenticated to `portal.example.com` with its "local" (i.e. self managed) session
2. The user is authenticated to Clerk's OAuth endpoint (i.e. `domain.clerk.accounts.dev` and has session cookies stored.
3. The user is authenticated to Google, again with session cookies stored.

Now here is the issue:
1. The user wants to log out of everything. 
2. They first go to `gmail.com` and log out of Google - this clears the Google Auth session (i.e. deletes the session cookoies among other things). **They are logged out of google**
3. Now they go to `portal.example.com` and logout - this clears the apps local session. **They are logged out of the app**
4. Now the user clicks Login in `portal.example.com`. **This is where the problem starts**. They are redirected by the app to Clerk's OAuth endpoint `my-domain.clerk.accounts.dev/oauth/authorize`. 
4. Clerk notices that it still has a session cookie stored for the user on its domain so it **skips any login prompt and immediately authenticates the user**. Clerk then redirects the user back to `portal.example.com` with a new OAuth authorization resposne, the app gets a new JWT token and reauthenticates the user.

**This means the user can re-authenticate to a session even after both the downstream IDP and the app itself were signed out**. 

This is because Clerk stores the session and automatically bypasses the auth prompt AND because there is no way for me as the owner of `portal.example.com` to tell Clerk to clear the session.

royanger:
> Now the user clicks Login in portal.example.com. This is where the problem starts. They are redirected by the app to Clerk's OAuth endpoint my-domain.clerk.accounts.dev/oauth/authorize.

As noted in Discord, this occurs because in the given example the user signed out of the Google and of the application using Clerk as an IdP. They did not sign out of the application that sits in the middle -- the one where the user used Google to sign up/in and is acting as the Clerk IdP for the `portal.example.com`. 

> because there is no way for me as the owner of portal.example.com to tell Clerk to clear the session.
* allow the user to sign out of the Clerk powered application that is providing the Clerk as IdP service. This would be a sign out route on that application or the account portal (https://accounts.<yourdomain>/user`
* the Backend API can be used to revoke the user's sessions
* create a route that blindly signs out anyone who visits it on the middle/main webapp

PR #1552, opened 2024-09-18
fix: Step 7 & 8 of Neon integration guide

zeguillen:
> [!IMPORTANT]
> 🔎 Previews:
>
> - 

### Explanation:
Fixes misspelled user ID from `user.id` to `userId` on step 7, and adds `await` to the messages fetch on step 8

### This PR:

- Fixes the misspelled and missing code to make the guide work out of the box


LekoArts (approved):
Thanks 👍

PR #1594, opened 2024-10-03
Clarify usage of system permissions on the server-side

LauraBeatris:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1594

### Explanation:

Resolves ORGS-229

System permissions aren't included in the user session claims, therefore cannot be used on the server-side with helpers such as `auth.has`

We should educate others to rely on their custom permissions instead of checking for system ones.

### DIFF
--- a/docs/organizations/roles-permissions.mdx
+++ b/docs/organizations/roles-permissions.mdx
@@ -60,6 +60,9 @@ Clerk’s System Permissions consist of the following:
 
 You can assign these System Permissions to any role.
 
+> [!WARNING]
+> System permissions are not included in user session claims and should not be used to protect server-side business logic. For server-side protection, create custom permissions instead.
+
 ### Custom permissions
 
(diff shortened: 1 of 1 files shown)


docs/organizations/roles-permissions.mdx
```diff
@@ -60,6 +60,9 @@ Clerk’s System Permissions consist of the following:
 
 You can assign these System Permissions to any role.
 
+> [!WARNING]
+> System permissions are not included in session claims. On the server side, [create custom permissions](/docs/organizations/create-roles-permissions) (e.g., `org:invoices:read`) to use with [`auth()`](/docs/references/nextjs/auth).
```
alexisintech, line 64:
```suggestion
> System permissions are not included in session claims. To do permission-checks on the server-side, you must [create custom permissions](/docs/organizations/create-roles-permissions).
```

PR #1595, opened 2024-10-03
docs(express): Start de-emphasizing usage of `getAuth`

wobsoriano:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1595/quickstarts/express#protect-your-routes-using-require-auth
> - https://clerk.com/docs/pr/1595/upgrade-guides/node-to-express

### Explanation:

The `getAuth()` helper is basically an unnecessary abstraction, which just calls `req.auth` under the hood and throws an error if the middleware is not installed.

### This PR:

This PR de-emphasizes the use of `getAuth()` and promote direct access to `req.auth` to access authenticate state. We're still keeping it to support the [Backend SDK](https://clerk.com/docs/references/sdk/backend-only#create-a-require-auth-helper) spec.


alexisintech (approved):
YASS!!! love it !!

PR #1609, opened 2024-10-08
chore(docs): Add information about customization limitations in Account Portal and Localization guides

kduprey:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1609

### Explanation:

There's no mention on the localisations guide page that the usage guide will only affect components that are used in the application, versus when developers are still using the hosted sign in and sign up pages, etc.

### This PR:

- Adds information about customisation limitations in the Account Portal and Localisation guides
- Fixes [DOCS-9359](https://linear.app/clerk/issue/DOCS-9359/clarify-localization-guide-on-usage-affecting-only-ui-components)


alexisintech (approved):
Perfection!! Love the copy used here. 😸💖

PR #1629, opened 2024-10-16
Add SignUp button for user registration

AnasSiddiqui18:
While following the Clerk documentation, I noticed that the SignUp button was missing from the SignedOut component. To enhance user experience and provide clearer navigation for new users, I have added the SignUp button alongside the existing SignIn button component.

This PR solves the problem by adding a SignUp button to the SignedOut component, allowing users to easily register for a new account. This enhances user experience by providing a clear option for new users to sign up, streamlining the registration process directly from the SignedOut state.

### DIFF
--- a/docs/quickstarts/nextjs.mdx
+++ b/docs/quickstarts/nextjs.mdx
@@ -125,6 +125,7 @@ description: Add authentication and user management to your Next.js app with Cle
               <header>
                 <SignedOut>
                   <SignInButton />
+                  <SignUpButton />
                 </SignedOut>
                 <SignedIn>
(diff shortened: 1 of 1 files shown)


alexisintech:
The sign in button takes the user to the sign in page, where a user can sign in, but they can also sign up from that same page. For brevity, we are only going to include a sign in button.

AnasSiddiqui18:
> The sign in button takes the user to the sign in page, where a user can sign in, but they can also sign up from that same page. For brevity, we are only going to include a sign in button.

Aah, understand. Thanks

PR #1659, opened 2024-10-26
Update local-credentials.mdx

george-raphael:
In the example, biometricType is imported from the useLocalCredentials() hook, but biometryType is used on the sign-in button. I think this was a typo.


alexisintech (approved):

alexisintech:
Good find! Thank you 😸💖

PR #1661, opened 2024-10-26
Change redirect variables description in clerk-environment-variables.mdx

derweili:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:
I think the description is wrong. The Signin page should render the `<SignIn /> component and the Signup page should render the `<SignUp />` component, not reverse (as the doc say currently)

### This PR:

- Corrects the components to be used with the signing and signup pages

### DIFF
--- a/docs/deployments/clerk-environment-variables.mdx
+++ b/docs/deployments/clerk-environment-variables.mdx
@@ -23,8 +23,8 @@ However, it's recommended to use environment variables instead of these props wh
   <Tab>
     | Variable | Description |
     | - | - |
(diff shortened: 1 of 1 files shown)


alexisintech:
actually, it's written correctly. 
there is a "sign up" link that's is rendered in the `<SignIn />` component. and vice versa.
see the `<SignIn />` component as an example:

Screenshot 2024-10-28 at 14 35 07

It says "No account? Sign up" and the "Sign up" is a link.

PR #1694, opened 2024-11-10
Replace 'response' with 'res'

Ipankaj07:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/references/express/overview

### Explanation:
The word "response" is replaced with "res" in the middleware to fix a response not found error.

### This PR:

- Replaces "response" with "res" in hasPermission.


alexisintech (approved):
good catch! thank youuu 😸💖

PR #1724, opened 2024-11-20
(/maintenance-mode) add details to errors

braden-clerk:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1724/maintenance-mode

### Explanation:

- 

### This PR:

- 

### Checklist

- [X] I have clicked on "Files changed" and performed a thorough self-review
- [X] I have added the "deploy-preview" label and added the preview link(s) to this PR description
- [X] All existing checks pass


docs/maintenance-mode.mdx
```diff
@@ -3,14 +3,35 @@ title: Maintenance Mode
...
+> [!IMPORTANT]
+> Maintenance mode will be deprecated in 2025 as we upgrade our infrastructure.
+
+Once or twice per year, Clerk undergoes maintenance on its infrastructure and enters **Maintenance Mode**. During this time, users who are already signed in will not be signed out, and will continue to have access to your app. However, new sign-ups, sign-ins and user mutations will return an error. **Maintenance Mode** is a special operational state designed to minimize disruption for signed-in users during critical database upgrades or outages.
 
 ## Production instances
 
-Mutation methods (`POST`, `PATCH`, `PUT`, `DELETE`) will be rejected with a `503 Service Unavailable` status, along with an error message informing users that the system is temporarily unavailable. This includes all sign-up and sign-in attempts. Clerk's prebuilt components and the hosted Account Portal will display errors for these requests.
+Mutation methods (`POST`, `PATCH`, `PUT`, `DELETE`) will be rejected with a `SystemUnderMaintenance` error. This includes all new sign-ups and sign-ins.
```
alexisintech, line 13: It's not a `SystemUnderMaintenance` error though, its a `503 StatusServiceUnavailable` error with the short message of "System under maintenance".

So I'm not sure if we want to reference it as `SystemUnderMaintenance` ? @jescalan what do you think

braden-clerk: I was copying the pattern in the "Errors" pages, which seems to "name" the errors -- I'm not sure where those come from.  I added the 503 in the definition, as I felt that was necessary

jescalan: I think we keep the 503 service unavailable here if that helps at all

alexisintech: ahh yes Bray thats bc we pull that from the function name
```
func AllowlistIdentifierNotFound(identifierID string) Error {
	return New(http.StatusNotFound, &mainError{
		shortMessage: "Identifier not found",
		longMessage:  "No identifier was found with id " + identifierID,
		code:         ResourceNotFoundCode,
	})
}
```

I see what you're saying though.
I could go for either way here. If I had a preference, I'd say we tell the users specifically they will receive a `503 Service Unavailable` error

alexisintech (approved):

PR #1725, opened 2024-11-20
feat: Machine Token Docs

Ben2W:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

Currently all customer-facing tokens are session tokens. We are adding `machine tokens` which is another customer-facing token. This PR adds documentation on:
- Understanding machine tokens
- Creating machine tokens
- Using machine tokens to authenticate requests
- authenticating requests with machine tokens.

Slack thread:
https://clerkinc.slack.com/archives/C079HR9DKPG/p1731406277241269

Linear Project:

### Checklist


docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+         ```
+    </Tab>
+
+    <Tab>
+    ```tsx import { auth } from '@clerk/nextjs/server';
+    import { NextResponse } from 'next/server';
+    import { runCronJob } from './cronJob';
+
+    export const POST = async () => {
+      const { isMachineAuthenticated } = await auth({ entity: 'machine' });
```
jakobevangelista, line 95:
```suggestion
      const { machineId } = await auth({ entity: 'machine' });
```
this works the same way as user `auth()` where you should check for `machineId`

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+      if (!isMachineAuthenticated) {
```
jakobevangelista, line 97:
```suggestion
      if (!machineId) {
```

to satisfy the change to machineId

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+By default, authenticating requests, will default to authenticating user [session requests](/docs/session-requests/overview).
+
+For machine requests, you can use the `entity: 'machine'` option to authenticate requests.
+
+> [!WARNING]
+> The `entity: 'machine'` option is only available in the Clerk Next.js SDK and Clerk Backend Javascript SDK during the beta.
+
+#### Examples
+
+  <Tabs items={['Next.js Middleware', 'Next.js API Route']}>
```
jakobevangelista, line 69: feels weird to me that the `authenticateRequest()` example is in a whole different tab, but it makes sense because `authenticateRequest()` is it's own thing, i might be too in the weeds / looking too deep into it

docs/machine-requests/manual-jwt.mdx
```diff
@@ -0,0 +1,134 @@
...
+</Steps>
+
+### Example
+
+The following example manually verifies a machine token.
+
+```tsx
+import jwt from 'jsonwebtoken'
+
+export default async function (req: Request, res: Response) {
```
jakobevangelista, line 92: should probably have a name on this function

docs/machine-requests/overview.mdx
Ben2W: Yea for me I think these examples are how to use machine to machine _in backend frameworks_.

However, since we only have Next.js, it feels weird that these are our only examples.

docs/machine-requests/manual-jwt.mdx
Ben2W: Done, thanks!

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## Verifying Machine Requests
+
+For a machine request to be valid, it must include a valid [machine token](/docs/machine-requests/machine-tokens) in the Bearer `Authorization` header.
+
+You can verify machine tokens in two ways:
+
+1. Using Clerk's Backend SDKs (recommended)
+2. Manually verifying the machine token JWT using your instance's public key.
```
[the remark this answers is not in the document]
Ben2W: Done

docs/machine-requests/overview.mdx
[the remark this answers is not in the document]
Ben2W: I agree, I switched the order around, Next.js API Route is first now

Ben2W:
@brkalow And @jakobevangelista 

I addressed all of the comments (thanks for reviewing it).

Few things:
- After rebasing the `auth()` reference doc was changed by someone else, I also found a small issue with `authenticateRequest()` so I pushed this commit: https://github.com/clerk/clerk-docs/pull/1725/commits/d7da65d9a602ed93466460a586174273fc5f0e57
- I changed the [manifest.json](https://github.com/clerk/clerk-docs/blob/d7da65d9a602ed93466460a586174273fc5f0e57/docs/manifest.json) to put "Backend Requests" at the same level as "Configuration. 
Image 2-6-25 at 11 42 PM
Image 2-6-25 at 11 43 PM

jakobevangelista:
lgtm, prolly should ask alexis and/or nick for an extra opinions about the backend section change

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+By default, authenticating requests will default to authenticating them as user [session requests](/docs/session-requests/overview).
+
+For machine requests, you can use the `entity: 'machine'` option to authenticate requests.
+
+> [!WARNING]
+> The `entity: 'machine'` option is only available in the Clerk Next.js SDK and Clerk Backend Javascript SDK during the beta.
+
+#### Examples
+
+<Tabs items={['Next.js API Route', 'Next.js Middleware']}>
```
NWylynko, line 69: Are you suggesting these two examples get used in unison? To me I feel that the best option is to have an explicit ignore in the middleware, and then do the check in the route. But this looks to suggest doing one or the other?

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+```tsx
+import { createClerkClient } from '@clerk/backend'
+
+export default async function cronJob() {
+  const clerkClient = createClerkClient({ secretKey: process.env.CLERK_SECRET_KEY })
+
+  const { token } = await clerkClient.machineTokens.create({
+    machineId: 'mch_cron',
+    claims: {
+      permissions: ['read', 'write'],
```
NWylynko, line 32: Some explanation around what the purpose of this is would be good, my guess is for the route to check that not only is the POST request authenticated but that its also authorized? I feel it would be beneficial to have down in the example below pulling these claims out and checking against them.

docs/machine-requests/machine-tokens.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+They are similiar to [session tokens](/docs/session-requests/session-tokens) however, unlike session tokens, machine tokens are not associated with a user. Instead **you** are responsible for determining the identity of the machine making the request.
+
+## Creating machine tokens
+
+### Machine ID
+
+Every machine token you create needs to be associated with a `machine_id`. You can pick any value for the `machine_id` as long as it meets the following requirements:
+
+- It must be prefixed with `mch_`
```
NWylynko, line 16: Since I am already explicitly calling `clerkClient.machineTokens.create` couldn't the sdk just prefix the id with `mch_` for me? If it has to be this way then I think this is pretty clearly outlined. The types for this field should definitely be `` `mch_${string}` `` to enforce this.

docs/machine-requests/machine-tokens.mdx
```diff
@@ -0,0 +1,65 @@
...
+- Examples:
+  - ✅ `mch_cron_service`
+  - ✅ `mch_background_worker`
+  - ✅ `mch_device_6580fc77_afca_47ac_8973_b7261d14e4c7`
+  - ❌ `user_2p94zsO6sBvVZR5Ca0KfBNLM36Z`
+  - ❌ `mch-invalid`
+  - ❌ `MCH_UPPERCASE`
+
+### Claims
+
```
NWylynko, line 32: An example here of how this works would be helpful, even better if the example showed how to type the claims, or validate them with a schema.

docs/machine-requests/machine-tokens.mdx
```diff
@@ -0,0 +1,65 @@
...
+**Machine tokens** are JWTs used to authenticate machine requests your application makes.
+
+They are similiar to [session tokens](/docs/session-requests/session-tokens) however, unlike session tokens, machine tokens are not associated with a user. Instead **you** are responsible for determining the identity of the machine making the request.
+
+## Creating machine tokens
+
+### Machine ID
+
+Every machine token you create needs to be associated with a `machine_id`. You can pick any value for the `machine_id` as long as it meets the following requirements:
```
NWylynko, line 14: It's not necessarily clear why a machine_id is required. Is it purely an internal thing for clerk? Does this give us controls in clerk dashboard to "ban" a machine? Are we supposed to validate the machine_id where we are receiving the request?

docs/machine-requests/machine-tokens.mdx
Ben2W: Yea this a debated point in the [#proj-m2m](https://clerkinc.slack.com/archives/C083Q5HRJM9) slack channel

My take is we shouldn't _automatically_ prefix the id, since customers will surely do checks in their api based on the machine id. For example:

```ts
const { machineId } = await auth();

if (machineId === "mch_cron" ) {
   await performCron();
}

if (machineId === "mch_microservice" ) { 
   await performSomeMicroserviceThingy();
}
```

But this will be _**very**_ confusing if the `mch_` is automatically added in the sdk.

I will change the types for the machineId to be `mch_${string}` though.

docs/machine-requests/machine-tokens.mdx
Ben2W: > Is it purely an internal thing for clerk? Does this give us controls in clerk dashboard to "ban" a machine? Are we supposed to validate the machine_id where we are receiving the request?

No. The `machine_id` is what m2m uses for the `sub` (subject) claim in the JWT. With session tokens _clerk_ defines the subject of the JWT with the `user_id`. With `m2m` however, the _customer_ defines the subject of the token. So simply, the `machine_id` is whatever the customer wants it to be.

What do you think about changing the wording from:

> Every machine token you create needs to be associated with a `machine_id`. You can pick any value for the `machine_id` as long as it meets the following requirements:

To:

> When creating a machine token, you must specify a machine_id. You can choose any value as long as it meets the following requirements:

docs/machine-requests/overview.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+### Example using the Clerk JavaScript Backend SDK
+
+```tsx
+import { createClerkClient } from '@clerk/backend'
+
+export default async function cronJob() {
+  const clerkClient = createClerkClient({ secretKey: process.env.CLERK_SECRET_KEY })
+
+  const { token } = await clerkClient.machineTokens.create({
```
jakobevangelista, line 29:
```suggestion
  const { jwt } = await clerkClient.machineTokens.create({
```

follows the jwt template api

PR #1735, opened 2024-11-21
How Clerk Works

jescalan:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1735/how-clerk-works/overview

It's been a long time coming, but finally it's here - how Clerk actually works, in the docs 🙌

### Checklist


docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+- **Cookie name:** `__session`
+- **Contents:** A Clerk-signed JWT containing [a set of default claims](/docs/backend-requests/resources/session-tokens#default-session-claims), which can also be customized to include additional claims via the Clerk dashboard.
+- **Domain:** Set on your application's domain directly, scoped strictly so it cannot be shared across subdomains. This is done for security, to prevent a security issue on a different app on a different subdomain from being able to take over your users' accounts. If you need to send the session token value across subdomain boundaries (like from `example.com` to `api.example.com`), you can [put the token in a request header instead](/docs/backend-requests/making/cross-origin).
+- **Expiration:** 60 seconds
+- **HttpOnly:** No - must be able to be accessed by client-side SDKs
+- **SameSite:** Lax
+
+The logistics of how these cookies are set and managed are handled by Clerk's frontend SDKs, as many of the details can't be handled purely by an API endpoint response. For this reason, we strongly recommend using one of Clerk'said frontend SDKs to implement authentication flows in general.
```
mwickett, line 150:
```suggestion
The logistics of how these cookies are set and managed are handled by Clerk's frontend SDKs, as many of the details can't be handled purely by an API endpoint response. For this reason, we strongly recommend using one of Clerk's frontend SDKs to implement authentication flows in general.
```

docs/how-clerk-works/tokens-signatures.mdx
```diff
@@ -0,0 +1,65 @@
...
+Digital signatures are a mathematical process that can be used to guarantee that a message sent from one person to another in public is genuine. In other words, it can guarantee for the recipient of a message that:
+
+1. The message definitely came from a specific sender (authenticity)
+1. The message’s contents were not modified from how they were written by the sender (integrity)
+
+It’s important to note that signatures do not actually encrypt a message - anyone can read the message’s contents.
+
+The way this works is as such:
+
+- Some fancy math is used to generate a pair of keys: a public key and a private key. This process is referred to as "public key cryptography".
```
mwickett, line 17: Nit: I get the point of "some fancy math", but maybe we use a slightly less throw away approach? I feel like there are _some_ who would read this as us being unserious, even though that's certainly not the case.

mwickett: or maybe we link "fancy math" to a legitimate deep dive? That would mitigate the concern IMO.

mwickett (commented):
This is legit awesome. Left a couple of random comments that popped up as I read it.

docs/how-clerk-works/tokens-signatures.mdx
jescalan:
```suggestion
- Some [fancy math](https://en.wikipedia.org/wiki/RSA_(cryptosystem)#Operation) is used to generate a pair of keys: a public key and a private key. This process is referred to as "public key cryptography".
```

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,256 @@
+---
+title: How Clerk works
+description: Learn how Clerk is architected and how it works under the hood.
+---
+
+This guide provides a deep dive into Clerk's architecture and internal workings. For developers who are simply looking to add authentication to their app, see the [quickstart guides](/docs/quickstarts/overview).
+
+## The frontend API
+
+When you create a new application through the [Clerk Dashboard](https://dashboard.clerk.com/), Clerk provisions a dedicated frontend API (FAPI) instance for your application. It is hosted at `https://<slug>.accounts.dev` in development environments, where `<slug>` is a unique identifier generated for your application. You can find your application's FAPI URL in the [**Domains**](https://dashboard.clerk.com/last-active?path=domains) page of the Clerk Dashboard.
```
alexisintech, line 10:
```suggestion
When you create a new application through the [Clerk Dashboard](https://dashboard.clerk.com/), Clerk provisions a dedicated frontend API (FAPI) instance for your application. It is hosted at `https://<slug>.clerk.accounts.dev` in development environments, where `<slug>` is a unique identifier generated for your application. You can find your application's FAPI URL in the [**Domains**](https://dashboard.clerk.com/last-active?path=domains) page of the Clerk Dashboard.
```

docs/how-clerk-works/tokens-signatures.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+1. A private key is used to create a **signature**:
+   1. The message is [hashed](https://www.techtarget.com/searchdatamanagement/definition/hashing) (a unique, fixed-length representation of the message).
+   1. The hash is encrypted using the sender's private key, creating the signature.
+1. The message and signature are sent to the recipient.
+1. The recipient verifies the signature using the sender's public key by:
+   1. Decrypting the signature to retrieve the hash.
+   1. Independently hashing the message and comparing the two hashes. If they match, the message is authentic and untampered.
+
+So for example, you get a publicly readable message, like `hello world`, and the message is “signed” with the signature `j2e80w8dj9f8`. If you'd like to be sure that the message is genuine, and you have a copy of the sender's public key, you can use this to decrypt the signature and make sure that it works. If it does, you know for sure who sent the message, and that nobody intercepted the message and messed with it in the middle, since part of the verification process involves hashing the message and comparing it to the decrypted signature.
```
jescalan, line 27:
```suggestion
So for example, imagine you get a publicly readable message, like `hello world`, and the message is “signed” with the signature `j2e80w8dj9f8`. If you'd like to be sure that the message is genuine, and you have a copy of the sender's public key, you can use this key to decrypt the signature and make sure that it is valid. If it is, you know two things for sure: who sent the message, and that nobody intercepted the message and messed with it in the middle, since part of the verification process involves hashing the message and comparing it to the decrypted signature.
```

docs/how-clerk-works/cookies.mdx
```diff
@@ -0,0 +1,103 @@
...
+Set-Cookie: fb_tracker=user123; SameSite=none
+
+...the content of the image
+```
+
+Now let's say you visit another website, `foobar.com`, and that website is using a script from Facebook for tracking the effectiveness of Facebook ads. So now `foobar.com` makes a request to `facebook.com`, and Facebook gets back the cookie that it set from the image on `example.com`. But how could this happen? Let's revisit this statement one more time:
+
+> Cookies are typically created by the server and communicated to the browser through the [`Set-Cookie` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie). When the browser receives this header, it stores the cookie and includes it in future requests to the **same domain**.
+
+Despite being set on `example.com`, the cookie's domain value is `facebook.com`, since it was set by Facebook. And remember that the cookie was set with the `SameSite` value set to “none”, which, according to [the docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value), means that the browser sends the cookie **with both cross-site and same-site requests.**
```
jescalan, line 80:
```suggestion
Despite being set on `example.com`, the cookie's domain value is `facebook.com`, since it was set _by_ Facebook. And remember that the cookie was set with the `SameSite` value set to “none”, which, according to [the docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value), means that the browser sends the cookie **with both cross-site and same-site requests.**
```

jescalan (commented):
Ok sorry this took so long to take a pass on, here's a first round. I have a couple more minor edits I'm pushing in a commit direct. I think there's a few more adjustments needed here, but it's close. Let's ship this week!

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,256 @@
...
+     width="1600"
+     height="900"
+     autoPlay
+     muted
+     loop
+     playsInline
+   />
+
+> [!NOTE]
+> If an attacker obtains your session ID, they can fully you as a user and take over your account. To mitigate this risk, Clerk implements the following security measures: 1. [HTTPS](https://developer.mozilla.org/en-US/docs/Glossary/HTTPS) is enforced to prevent [man-in-the-middle attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) that could intercept session IDs in transit, and 2. the [`HttpOnly` flag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#httponly) is set on session cookies to prevent malicious JavaScript from accessing them through cross-site scripting (XSS) attacks. [Read more about Clerk's security measures](/docs/security/overview).
```
jescalan, line 94: This section needs to be heavily modified - we're not talking about Clerk's auth scheme here yet so an injection about how Clerk does auth feels out of place. Also, we don't actually use httponly for session cookies, so this isn't accurate. Will take another pass on this later but wanted to drop a note so i dont forget at least

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,256 @@
...
+
+FAPI _does not_ handle administrative actions that impact multiple users, such as listing all users, banning users, or impersonating a user. These types of tasks are handled by [the backend API](#backend-api).
+
+Some tasks, such as [signing up a user](https://clerk.com/docs/reference/frontend-api/tag/Sign-Ups#operation/createSignUps), don't require authentication, as that would defeat the purpose of the endpoint. However, endpoints designed for authenticated users, like [updating a user's details](https://clerk.com/docs/reference/frontend-api/tag/User#operation/patchUser), require FAPI to 1. identify the user making the request and 2. verify their authorization. This ensures that users cannot modify another user's details. Typically, this is achieved by sending a signed token with the request, either as a cookie or a header. You can [learn more about Clerk's authentication tokens later in this guide](#stateful-authentication).
```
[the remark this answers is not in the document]
alexisintech: oops, I can fix that

docs/how-clerk-works/cookies.mdx
```diff
@@ -0,0 +1,103 @@
...
+</html>
+```
+
+This response sets a `session_id` cookie with the value `sess123`. To view cookies in your browser's developer tools, navigate to the **Application** tab. Then in the **Storage** section, select **Cookies**. Here's an example from Clerk's website:
+
+![cookies in browser console](/docs/images/how-clerk-works/devtools-cookies.png)
+
+## Cookie domains and scope
+
+Each cookie has a **Domain** that indicates **the domain from which the cookie was set**. This determines when the cookie will be included in requests.
```
alexisintech, line 36:
```suggestion
Each cookie has a [`Domain`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent) that indicates **the domain from which the cookie was set**. This determines when the cookie will be included in requests.
```

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,255 @@
...
```
alexisintech, line 30:
```suggestion
Some tasks, such as [signing up a user](https://clerk.com/docs/reference/frontend-api/tag/Sign-Ups#operation/createSignUps), don't require authentication, as that would defeat the purpose of the endpoint. However, endpoints designed for authenticated users, like [updating a user's details](https://clerk.com/docs/reference/frontend-api/tag/User#operation/patchUser), require FAPI to first identify the user making the request and then verify their authorization. This ensures that users cannot modify another user's details. Typically, this is achieved by sending a signed token with the request, either as a cookie or a header. You can [learn more about Clerk's authentication tokens later in this guide](#stateful-authentication).
```

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+## The backend API
+
+The frontend API (FAPI) is designed for use primarily from the frontend of your application. Its methods focus on signing in users and managing user-related resources and data once they are authenticated. However, as an application developer, you might also need to perform administrative tasks, such as modifying multiple user or organization details, retrieving a list of all users, banning or impersonating a user, and more.
+
+To maintain security, these administrative tasks should only be executed on the server side using a secret key inaccessible to your users or the browser. These operations are handled by a separate API known as the backend API (BAPI). You can find the BAPI documentation [here](https://clerk.com/docs/reference/backend-api).
+
+Like FAPI, while you can interact directly with BAPI, most developers opt to use Clerk's SDKs for smoother integration with their preferred language or framework. Documentation for Clerk's SDKs is available in [the left sidebar of the docs](https://clerk.com/docs).
```
alexisintech, line 60: ahh good point - what do you think about this update:

```suggestion
Like FAPI, while you can interact directly with BAPI, most developers opt to use Clerk's SDKs for smoother integration with their preferred language or framework. Documentation for Clerk's SDKs is available in [the left sidebar of the docs](https://clerk.com/docs). That being said, FAPI is an extremely more complex and nuanced API that relies on more custom logic outside of its endpoints to create a functional SDK on top of it. As such, **interacting directly with FAPI is not recommended**, whereas interacting directly with BAPI is generally reasonable.
```

docs/how-clerk-works/overview.mdx
alexisintech: oh, i wrote that incorrectly about the `httponly` flag :(
and i see, good point about injecting how clerk handles it! i thought we were mentioning things about other flows (in this case, stateful) in order to compare to clerk's flows. 
i would ask then, what is the purpose of providing this information? why provide this callout for users?

docs/how-clerk-works/overview.mdx
```diff
@@ -0,0 +1,256 @@
...
+Furthermore, even after rotating keys, the revocation may be delayed if your application caches the public key used for verification - a common practice for performance optimization. The cached key would continue to validate the old tokens until the cache expires.
+
+This limitation poses significant security risks, as it hampers your ability to quickly respond to security incidents that require immediate session termination for specific users.
+
+## Clerk's authentication model
+
+Clerk implements a hybrid authentication model that combines the benefits of both stateful and stateless approaches while mitigating their respective drawbacks, at the cost of adding a substantial amount of complexity to the implementation on Clerk's side. For a developer implementing Clerk, like you, this is all upside since the complexity is handled internally by Clerk. The hybrid model incorporates the same flow when signing in as the stateless flow, but with a few key innovations:
+
+- A short-lived **session token** (60 seconds) combined with a background token refresh mechanism.
+- A **client token** to verify the user's authentication state, in conjunction with the session token.
```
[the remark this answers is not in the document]
alexisintech: let's pair and drive it home more!

alexisintech (approved):

PR #1813, opened 2024-12-16
feat: Added doc for for web server deployment

royanger:
> [!IMPORTANT]
> 🔎 Previews:
>
> - https://clerk.com/docs/pr/1813/deployments/deploying-behind-a-proxy

### Explanation:

This creates a new Deployment doc covering the two headers that must be forwarded from the server/container to the application for Clerk to work correctly.

### This PR:

Lists the headers, has a basic nginx example and covers how to test that the config is correct.

### Checklist


braden-clerk:
Would "Deploying behind a proxy" be a better way to frame this?

JacobMGEvans:
I think the more explicit/specific the title might help people coming to us with these issues. I tend to agree with Roy in this case, the type of people that I tend to help that need a doc like this are not super familiar with the infrastructure they are using or understand what it is doing fully.

jescalan:
agreed with braden here

royanger:
I will get a new version up tomorrow. I'll make sure the content should be good for searches.

docs/deployments/deploying-behind-a-proxy.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: Deploying a Clerk app behind a proxy
+description: Learn how to deploy a Clerk app to a web server using proxies and reverse proxies like nginx or Caddy, or using Docker
+---
+
+When deploying a Clerk app beind a proxy, you may need to proxy two headers for Clerk to work correctly. The two headers that you must forward are:
```
alexisintech, line 6: the first sentence says they "may need to" proxy the two headers. the second sentence says they "must" forward the headers.

is it required, or is it not required?

docs/deployments/deploying-behind-a-proxy.mdx
```diff
@@ -0,0 +1,44 @@
...
+The following can be added to your `nginx` configuration to forward these two headers to your application.
+
+```nginx
+proxy_set_header X-Forwarded-Host $host;
+proxy_set_header X-Forwarded-Proto https;
+```
+
+## How to test if the headers are forwarded correctly:
+
+You can console log the `X-forwarded-host` and `X-forwarded-proto` to check what the value is. The following example is for Next.js.
```
alexisintech, line 22: What should the values be, so that the user knows that the forwarding worked?

docs/deployments/deploying-behind-a-proxy.mdx
JacobMGEvans: Lets use more assured its required tense, I think its safe to assume it needed in nearly if not all network configs.

docs/deployments/deploying-behind-a-proxy.mdx
JacobMGEvans: - `X-Forwarded-Host`: The original host that the client requested, like if your public domain is `example.com`, then `X-Forwarded-Host` should appear as `example.com`

- `X-Forwarded-Proto`: The original protocol used by the client. If your site is served over HTTPS, then `X-Forwarded-Proto` should appear as https... Clerk requires HTTPS so it needs to be HTTPS

royanger: Example expected output from the console log, assuming that the application's production domain is `clerk.com` below. I'll add this to the PR now.

```
=========================
Request Headers
x-forwarded-host clerk.com
x-forwarded-proto https
=========================
```

royanger: Added in https://github.com/clerk/clerk-docs/pull/1813/commits/ed19a2dea296c2ce6f42d98684070813d3289bc0

alexisintech (approved):
I've posted my docs review, lmk what you think! if you want to change it, feel free, or if you like it, feel free to merge! 😸💖

PR #1816, opened 2024-12-16
Jeremy/add clerk jwt debugger

jeremy-clerk:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

Add Clerk JWT debugging tool. 

### This PR:

Adds component and page in troubleshooting section. 

### Checklist

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -1253,6 +1253,10 @@
                         }
                       ]
                     ]
+                  },
+                  {
+                    "title": "Clerk JWT Debugger",
+                    "href": "/docs/troubleshooting/jwt-debugger"
                   }
                 ]
               ]
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -1234,6 +1234,10 @@
                     "href": "/docs/troubleshooting/script-loading"
                   },
                   {
+                    "title": "Clerk JWT Debugger",
+                    "href": "/docs/troubleshooting/jwt-debugger"
+                  },
+                  {
                     "title": "Help & Support",
                     "items": [
                       [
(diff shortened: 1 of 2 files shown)


alexisintech:
hmm the page is giving a 400 https://clerk.com/docs/pr/jeremy-add-clerk-jwt-debugger/guides/development/troubleshooting/jwt-decoder

jeremy-clerk:
> How would I be able to test this? I'm getting this locally (even when on your clerk PR branch), and wondering if there is a specific setup I need to follow here:

@SarahSoutoul - I have no idea, I would guess something isn't installed or linked, happy to pair with you when you have a moment to sort it out though!

jeremy-clerk:
@alexisintech I also wanted to put something somewhere about "How this works" to clarify for users but I didn't have any good ideas of where/what it should look like. Curious if you had any good ideas or insight into that.

jeremy-clerk:
Added some more info on the bottom 
image_11_21_4FBwgPVb@2x

jeremy-clerk:
> > Added popover for 'How do I get a Clerk session token?'
> 
> > 
> 
> > image_11_21_bOLdj8vt@2x
> 
> 
> 
> it feels like it should be hoverable.
> 
> if you want it to be clickable instead of hoverable, i'd update the design so it looks more like a button (indicating clickability).

I fought with trying to get the underline for a while and gave up 😂

The only reason I opted for clickable was so you don't have that annoying thing where you're trying to interact with it and it goes away

alexisintech:
> I agree with moving the text to the top.
> 
> Once that's complete, is this ready to go?

no there's still the design aspect i left a suggestion for - making the "How do I get a Clerk session token?" design fit better to its action. so if its clickable, it should look like a button. if its hoverable, it should indicate that.

jeremy-clerk:
image_12_18_GZrf6SMO
image_12_18_40KEe2fu

alexisintech (approved):
🚢 🚢 🚢

PR #1842, opened 2025-01-06
Use updateUserMetadata instead of updateUser

hktang:
According to the JS SDK doc, `updateUser()` overrides existing public/private metadata. If so, wouldn't it be better to use `updateUserMetadata()` instead?

Or, perhaps an info/warning message about the behavior of `updateUser()` in terms of how it handles public/private metadata would be nice.


alexisintech:
@royanger I know you had a big part in this guide; what do you think of this suggestion!

royanger (approved):
This is a reasonable change. Thanks for the PR!

victoriaxyz (approved):

PR #1891, opened 2025-01-14
Fix page title

bradlc:
Page titles are parsed as MDX so `<RedirectToCreateOrganization />` needs to be escaped or wrapped in `` ` ``

https://clerk.com/docs/pr/1891/components/control/redirect-to-createorganization


alexisintech (approved):
oh, woops! now we know 😓

PR #1929, opened 2025-01-17
(sso microsoft): add warning about client credentials expiring; align oauth and easie guides

victoriaxyz:
- **preview**: https://clerk.com/docs/pr/1929/authentication/social-connections/microsoft
- **ticket**: https://linear.app/clerk/issue/DOCS-9776/add-a-note-about-client-credentials-expiring-to-msft-sso-docs
- **note**: added warnings to two different places, one under production instances and one right next to the step where the client sets up the secret


alexisintech:
i've updated to use partials and align both the oauth and easie docs - and included the callout in the "get client secret" step

alexisintech (approved):

PR #1940, opened 2025-01-22
sidenav suggestion

alexisintech:
I've done a few things here:
1. I've removed collapsed sections (dropdowns) so that when you choose an SDK, its contents are fully revealed. They are organized as "Quickstart", "References", "Guides".
    - For example, choosing Next.js reveals "Quickstart", "General References", "App Router References", "Pages Router References", "Guides".
    - Choosing React reveals "Quickstart", "References" and inside "References" is "Client-side helpers". If there were guides, they'd then be under "Guides", but they are not.
2. I've reverted "Available Methods" back to "Overview". Methods doesn't feel like the right word here, and the page _practically is_ more of an overview of the reference section. So I've updated it to be "Overview" under "References", as its the Reference Overview.

https://github.com/user-attachments/assets/e3aead7c-b132-49dd-a8e5-958873c4eb98

3. I've renamed "UI Components" --> "Clerk Components", as not all of them are UI and we usually refer to them as "Clerk Components" anyways. This of course wasn't your doing, but I'm putting this here for visibility.
4. At first, I lumped the sections "UI Components", "Configuration", "Customization", "Development", and "Advanced" into one lump, and they were all drop downs ([1a38a0b](https://github.com/clerk/clerk-docs/pull/1940/commits/1a38a0bcf4fff1a2b0db33aff66c1343f830e9dc)). My qualms with this is that I don't like them being dropdowns, I don't like content being hidden because users don't know where to look for the guides they're looking for. So I have another option, which is what's shown in the second screenshot ([59d9e84](https://github.com/clerk/clerk-docs/pull/1940/commits/59d9e84fd7bcef72a2a3cb4d93cb1350e8a7227f)). Now, "Clerk Components" is **not** a dropdown, but "UI Components" is. I do like this better then all of these sections being hidden into dropdowns, but we should ask some others:

Screenshot 2025-01-22 at 15 43 42

Screenshot 2025-01-22 at 15 38 25


NWylynko:
I'm happy with all these changes, just feel that guides should be above references as a new dev will start by reading guides on getting clerk setup for there sdk then go in to the references to learn about specific functions available.

jescalan:
Only issue I see here is that only next.js has the "setting up your clerk account" in it's nav, all the others start with quickstart. Feels like we should be consistent there?

jescalan (approved):

alexisintech:
> Only issue I see here is that only next.js has the "setting up your clerk account" in it's nav, all the others start with quickstart. Feels like we should be consistent there?

yes, i meant to do it for all sdks ! forgot to mention that

PR #1942, opened 2025-01-23
update `style guide` reference

programming-with-ia:
update `style guide` reference to renamed STYLEGUIDE file


alexisintech (approved):
nice catch!

PR #1945, opened 2025-01-24
Updated NextJS proxy example to use nextUrl.pathname

jeremy-clerk:
### 🔎 Previews:

- https://clerk.com/docs/pr/1945/advanced-usage/using-proxies
- (nextjs tab)

### What does this solve?

The current proxy example for NextJS matches `__clerk` anywhere in the URL which will match search params like `__clerk_ticket` or `__clerk_status` and proxy those requests. 

Developers who use the example scheme we have of `/__clerk` may run into issues. 

### What changed?

Updated the example to use `nextUrl.pathname` as this will exclude search params. 

### Checklist


jescalan (approved):
great catch

PR #1699, opened 2024-11-12
[Draft] Accountless docs

mwickett:
> [!IMPORTANT]
> 🔎 Previews:
>
> -

### Explanation:

Stub in docs for Accountless mode

PR #1978, opened 2025-02-04
[fix] Make nextjs quickstart search visible

mwickett:
In #1699 we had the quickstart hidden from search and crawlers.


bradlc (approved):

alexisintech (approved):
oops!

PR #1994, opened 2025-02-06
(/customization/user-profile) fix broken javascript example

nme22:
### 🔎 Previews:

https://clerk.com/docs/pr/1994

### What does this solve?
We've had a handful of folks attempting to utilize the JavaScript example on this page and reporting that the feature is broken or bugged. This will solve the broken behavior experienced from our users.

 

After testing and reproducing the behavior we found the current example in the docs doesn't work. The former example reflects using the openUserProfile() method which opens this component as a modal. Custom pages cannot be mounted this way. Adjusted example to utilize the mountUserProfile() instead which works as intended.

### What changed?
- Change `clerk.openUserProfile()` -> `clerk.mountUserProfile()`

 
This change reflects the correct method to use in the example code we provide.

### Checklist

### DIFF
--- a/docs/customization/user-profile.mdx
+++ b/docs/customization/user-profile.mdx
@@ -208,7 +208,7 @@ The following example demonstrates two ways that you can render content in the `
 
 ### JavaScript example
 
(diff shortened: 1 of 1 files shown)


alexisintech (approved):
nice catch!! it must have been updated then because I definitely remember testing this, and it worked :'(

nme22 (commented):
LGTM 🚀

PR #2032, opened 2025-02-17
remove multisessionappsupport component

alexisintech:
⚠️  Redirects necessary: https://github.com/clerk/clerk/pull/1030

### 🔎 Previews:

- https://clerk.com/docs/pr/2032/authentication/configuration/session-options

### What does this solve?

From https://linear.app/clerk/issue/SDK-2060/export-multisessionappsupport-for-other-sdks#comment-f76a9530
> In Core 2 we intentionally moved MultisessionAppSupport from @clerk/nextjs to @clerk/nextjs/internal 
>
> Our own dashboard implements it on it's own. So i'm trusting our past selves and suggestions we should just have a snippet of how to DIY it.
```tsx
function MultisessionAppSupport({
  children,
}: React.PropsWithChildren<unknown>) {
  const { session } = useSession();
  return (
    <React.Fragment key={session ? session.id : 'no-users'}>
      {children}
    </React.Fragment>
  );
}
```

### What changed?

- Removes `<MultisessionAppSupport />` component reference
- Add this code snippet to the docs

### Checklist


docs/authentication/configuration/session-options.mdx
```diff
@@ -70,14 +70,43 @@ To enable multi-session in your application, you need to configure it in the Cle
 
 ### Add multi-session support to your app
 
-<If sdk={['nextjs', 'react']}>
-  It's highly recommended to wrap your application in the [`<MultisessionAppSupport />`](/docs/components/control/multi-session) component. This guarantees a full rerendering cycle every time the current session and user changes.
-</If>
+It's highly recommended to wrap your application with the following `<MultisessionAppSupport />` component. This guarantees a full rerendering cycle every time the current session and user changes.
+
+```tsx
+function MultisessionAppSupport({ children }: React.PropsWithChildren<unknown>) {
```
jescalan, line 76: What's going on with this typing? Can't you use `React.ReactNode` here instead?

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,14 +70,43 @@ To enable multi-session in your application, you need to configure it in the Cle
...
 </If>
 
+The following example demonstrates adding multi-session support to a Next.js App Router application.
+
+```tsx
+import React from 'react'
+import { ClerkProvider } from '@clerk/nextjs'
+
+function MultisessionAppSupport({ children }: React.PropsWithChildren<unknown>) {
+  const { session } = useSession()
```
jescalan, line 93: Missing the import for `useSession` here

jescalan: Also wouldn't this need to be a client component in order to use this hook? I think it needs to live in a separate file with `use client;` in it.

jescalan (changes requested):
There are a couple of errors in here still, I think, but please call me out if I'm wrong. Also just want to drop a general note around how this component actually works for consideration.

If you use this component and wrap your whole app with it, it does mean that every component in your app ends up being rendered on the client - there's no more "pure" server rendered components.

A server component inside a client component still does render its contents on the server, so if you're doing something like fetching from a database, it's not going to error out as this logic will run on the server still. However, the resulting output is hydrated on the client, which does make your app a little less efficient than a "pure" server component which doesn't hydrate at all.

Generally, this way of writing your app breaks the pattern of "use server components by default, add client components for islands of interactivity", as it wraps the entire app with a client component (worth noting, `ClerkProvider` is not a client component).

Also, adding 'import "server-only"' to any component in your app will produce an error with this wrapper over it.

I'm not exactly suggesting this is the wrong pattern - this is, as far as I know, the only actual way to support multi-session apps, aside from writing logic that pushes a hard refresh to the page when the session changes. But I do think it's worth clarifying this in the docs, since it would be easy to read this without a thorough understanding of how server and client components work and get into a confusing or frustrating situation.

docs/authentication/configuration/session-options.mdx
alexisintech: yeah that's what I usually use - @panteliselef was there a specific reason to using the types you supplied?

docs/authentication/configuration/session-options.mdx
alexisintech: oh yeah good catch - I didn't test this code just yet, I wanted to get feedback first incase it ended up changing. my bad

alexisintech:
> There are a couple of errors in here still, I think, but please call me out if I'm wrong. Also just want to drop a general note around how this component actually works for consideration.
> 
> If you use this component and wrap your whole app with it, it does mean that every component in your app ends up being rendered on the client - there's no more "pure" server rendered components.
> 
> A server component inside a client component still does render its contents on the server, so if you're doing something like fetching from a database, it's not going to error out as this logic will run on the server still. However, the resulting output is hydrated on the client, which does make your app a little less efficient than a "pure" server component which doesn't hydrate at all.
> 
> Generally, this way of writing your app breaks the pattern of "use server components by default, add client components for islands of interactivity", as it wraps the entire app with a client component (worth noting, `ClerkProvider` is not a client component).
> 
> Also, adding 'import "server-only"' to any component in your app will produce an error with this wrapper over it.
> 
> I'm not exactly suggesting this is the wrong pattern - this is, as far as I know, the only actual way to support multi-session apps, aside from writing logic that pushes a hard refresh to the page when the session changes. But I do think it's worth clarifying this in the docs, since it would be easy to read this without a thorough understanding of how server and client components work and get into a confusing or frustrating situation.

This is really good feedback. So would something like this work instead?

```tsx
function MultisessionAppSupport({ children }: React.ReactNode) {
  const { sessionId } = auth()
  return <React.Fragment key={sessionId ? sessionId : 'no-users'}>{children}</React.Fragment>
}
```

jescalan:
I think the issue is that `auth()` is not a hook - so when its value changes, it won't trigger a re-render 😞

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,14 +70,49 @@ To enable multi-session in your application, you need to configure it in the Cle
 
 ### Add multi-session support to your app
 
-<If sdk={['nextjs', 'react']}>
-  It's highly recommended to wrap your application in the [`<MultisessionAppSupport />`](/docs/components/control/multi-session) component. This guarantees a full rerendering cycle every time the current session and user changes.
+It's highly recommended to wrap your application with the following `<MultisessionAppSupport />` component. This guarantees a full rerendering cycle every time the current session and user changes.
+
+<If sdk={["nextjs"]}>
+  For Next.js App Router applications, using the following component requires converting your `app/layout.tsx` file to a client component. Server components will still render its contents on the server, so if you're doing something like fetching from a database, it will not error. However, the server components' rendered output will be hydrated on the client, resulting in slightly reduced performance compared to "pure" server components.
```
panteliselef, line 76: Not necessarily, if they define the `MultisessionAppSupport` in another file, they can move the 'use client' barrier to that file. This will allow the root layout to still be a RSC.

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,14 +70,49 @@ To enable multi-session in your application, you need to configure it in the Cle
...
+
+function MultisessionAppSupport({ children }: { children: React.ReactNode }) {
+  const { session } = useSession()
+  return <React.Fragment key={session ? session.id : 'no-users'}>{children}</React.Fragment>
+}
+
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <html lang="en">
+      <ClerkProvider afterMultiSessionSingleSignOutUrl="/">
```
panteliselef, line 106: Interesting case:

If you leave it undefined, that what we do is when signInUrl is available we are redirecting to `${signInUrl}/choose`. This results to `/sign-in/choose` which actually render a nice UI for you.

Screenshot 2025-02-19 at 1 30 25 PM

panteliselef (approved):

docs/authentication/configuration/session-options.mdx
alexisintech: ooo nice info!! i've added it [b950970](https://github.com/clerk/clerk-docs/pull/2032/commits/b950970da65307d8fa72e31c3c5e4361f44dff64)

docs/authentication/configuration/session-options.mdx
alexisintech: ohhh, you're so right!! we can just do a client component imported into the rsc
i've updated: [b950970](https://github.com/clerk/clerk-docs/pull/2032/commits/b950970da65307d8fa72e31c3c5e4361f44dff64)

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,19 +70,68 @@ To enable multi-session in your application, you need to configure it in the Cle
...
+function MultisessionAppSupport({ children }: { children: React.ReactNode }) {
+  const { session } = useSession()
+
+  return <React.Fragment key={session ? session.id : 'no-users'}>{children}</React.Fragment>
+}
+```
+
+The following example demonstrates adding multi-session support to a Next.js App Router application. It requires creating the `<MultisessionAppSupport />` as a client component, and then using it in the `app/layout.tsx` file to wrap the entire application. This ensures the `layout.tsx` remains a server component.
+
+<CodeBlockTabs items={["app/layout.tsx", "MultisessionAppSupport.tsx"]}>
```
bradlc, line 90:
```suggestion
<CodeBlockTabs options={["app/layout.tsx", "MultisessionAppSupport.tsx"]}>
```

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,19 +70,68 @@ To enable multi-session in your application, you need to configure it in the Cle
...
+}
+```
+
+The following example demonstrates adding multi-session support to a Next.js App Router application. It requires creating the `<MultisessionAppSupport />` as a client component, and then using it in the `app/layout.tsx` file to wrap the entire application. This ensures the `layout.tsx` remains a server component.
+
+<CodeBlockTabs options={["app/layout.tsx", "MultisessionAppSupport.tsx"]}>
+  ```tsx {{ filename: 'app/layout.tsx' }}
+  import React from 'react'
+  import { ClerkProvider } from '@clerk/nextjs'
+  import { MultisessionAppSupport } from './MultisessionAppSupport'
```
alexisintech, line 94:
```suggestion
  import { MultisessionAppSupport } from './components/MultisessionAppSupport'
```

docs/authentication/configuration/session-options.mdx
```diff
@@ -70,19 +70,68 @@ To enable multi-session in your application, you need to configure it in the Cle
...
+          <MultisessionAppSupport>
+            <body>{children}</body>
+          </MultisessionAppSupport>
+        </ClerkProvider>
+      </html>
+    )
+  }
+  ```
+
+  ```tsx {{ filename: 'components/MultisessionAppSupport.tsx' }}
```
alexisintech, line 109:
```suggestion
  ```tsx {{ filename: 'app/components/MultisessionAppSupport.tsx' }}
```

PR #2034, opened 2025-02-18
Include port on Next.js proxy example

dandanthedev:
### 🔎 Previews:

-

### What does this solve?

The current Clerk docs do not specify a port for the Next.js proxy. As the url is taken from the request url, any port that the user is accessing the server from (or more likely, the port that a reverse proxy listens to) gets used on the clerk request. So you end up with frontend-api.clerk.dev:8080 being fetched.

### What changed?

I have added a specification of the correct SSL port.

### Checklist


alexisintech (approved):
great catch!! thank you so much 😸💖

PR #2041, opened 2025-02-20
Docs Build Script

NWylynko:
Things this script does

Validates
- The manifest structure and its contents
- Markdown files and their required frontmatter fields:
  - Ensures title is present (required)
  - Warns if description is missing (optional)
  - Validates SDK declarations in frontmatter
- Validates internal doc links exist
- Validates hash links point to headings
- SDK filtering in three contexts:
  1. Manifest: Ensures SDK scoping is properly defined and inherited
  2. Frontmatter: Validates SDK declarations in document metadata
  3. `<If />` components: Ensures:
     - Referenced SDKs exist in the manifest
     - SDKs are available in the frontmatter
     - SDK values match the list of valid SDKs
     - Parent group SDK compatibility
- Unique headings within documents
- Typedoc content structure and references
- Validates all embedded content (partials, typedocs) exists and is properly formatted

Transforms
- Content Integration:
  - Embeds partial content into markdown files
  - Embeds typedoc content where referenced
  - Handles special character encoding in typedoc tables
- Link Processing:
  - Updates links to SDK-specific docs to use `<SDKLink />` components
  - Removes .mdx extensions from doc links
- SDK-Specific Processing:
  - Generates SDK-specific versions of docs in their respective folders
  - Creates "landing" pages for SDK-specific docs at original URLs
  - Strips out content filtered by SDKs
- Manifest Processing:
  - Generates processed manifest.json with SDK scoping
  - Applies inheritance rules for SDK scoping in the navigation tree

This is run on every push in a github action to help lint for authors and to build the dist package for clerk.com to use.
Authors can run `npm run build` locally to use it

This is all to enable sdk based filtering at a fine-grained control. The goal is:

Starting from the top:
1. the "sdk" key value pair in the manifest.json file
   - By setting the sdk key on a group of guides, it hides that group when a user is using a different sdk
   
2. the "sdk" key value pair in the frontmatter of a guide
   - This controls which sdks a specific guide is available to
   - If this is left unset, it will show up for all sdks
   - If the guide sits within a filtered group of the manifest but attempts to use a different sdk, it will throw an error
   
3. The `<If />` component
   - This allows showing parts of the guide to a subset of sdks
   - If this guide is being filtered by 1 or 2, this will error if you attempt to use a sdk outside of the filter
   
There is no requirement to use filter 3 to use 2 or using 1 to use 3, all three should be optional tools to achieve the docs we are aiming for. But the filtering down should follow a logical order.


CONTRIBUTING.md
```diff
@@ -379,6 +379,28 @@ You may also set `search` to a boolean value, which acts as an `exclude` value.
...
+
+The `sdk` frontmatter field defines what SDKs a page supports and makes the page visible in the sidenav only when one of those SDKs is selected by the [SDK selector](#sidenav).
+
+For example, if `nextjs` and `react` were passed to the `sdk` frontmatter field, like so:
+
+```diff
+  ---
+  title: `'<ClerkProvider>'`
+  description: Lorem ipsum...
++ sdk: nextjs, react
```
alexisintech, line 392: we should list here the list of supported keys

NWylynko: We already have the list under the `<If />` component section, I don't think it makes sense to duplicate it, but maybe move it somewhere relevant to both?

jescalan (approved):

PR #2042, opened 2025-02-20
Update "API Key Changes" in manifest to redirected href

NWylynko:
### 🔎 Previews:

- https://clerk.com/docs/pr/2042/deployments/clerk-environment-variables

### What does this solve?

The link `/docs/upgrade-guides/api-keys` gets redirected to `/docs/deployments/clerk-environment-variables` so faster to just have it link directly. It's possible instead we just want to remove this item? instead of having it duplicated as "Clerk environment variables" also links to `/docs/deployments/clerk-environment-variables`.

### What changed?

- Updated link in manifest

### Checklist

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -3239,10 +3239,6 @@
                     "items": [
                       [
                         {
-                          "title": "API Key Changes",
-                          "href": "/docs/deployments/clerk-environment-variables"
-                        },
-                        {
                           "title": "URL-based session syncing",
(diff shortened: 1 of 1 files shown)


alexisintech:
yeah I think it's from a reaally old version of clerk, and we removed that guide because we checked and no users are using the old API keys, so we can just remove it from the manifest completely

alexisintech (approved):

PR #2046, opened 2025-02-21
feat(captcha): Add customization options for the CAPTCHA widget

anagstef:
### 🔎 Previews:

- https://clerk.com/docs/pr/2046/custom-flows/bot-sign-up-protection
- https://clerk.com/docs/pr/2046/customization/captcha
- https://clerk.com/docs/pr/2046/customization/overview

### What does this solve?

This gives the ability to developers to customize the appearance of the CAPTCHA widget.

### What changed?

Adds the new `appearance.captcha` property for customizing the CAPTCHA widget. It has 3 properties:

- `theme`
- `size`
- `language`

Also, for custom flows we have added 3 data attributes:

- `data-cl-theme`
- `data-cl-size`
- `data-cl-language`

More info on this PR: https://github.com/clerk/javascript/pull/5184

### Checklist


docs/customization/captcha.mdx
```diff
@@ -0,0 +1,191 @@
...
+
+  The CAPTCHA widget size. Defaults to `normal`.
+
+  ---
+
+  - `language`
+  - `string`
+
+  The CAPTCHA widget language. Must be either: `auto` (default) to use the language that the visitor has chosen, or language and country code (e.g. `en-US`).
+  If Clerk UI Components are used, if this property is omitted, the language will be automatically set based on the [localization](/docs/customization/localization) used.
```
alexisintech, line 31: I'm not too familiar with how localization works, but if the `language` prop is omitted, would it not default to `'auto'`? And `'auto'` uses the language the visitor has chosen, not the language based on the localization used? 

Or is that the same thing - the language the visitor chooses is the same thing as the localization used?

anagstef: Hmm this is a bit confusing indeed. I may need to explain it better on the docs here. So, what happens currently is the following:

If the developer is using our UI components:
- The priority is:
1. `appearance.captcha.language`
2. `localization.locale`
3. `en-US`
- We default to `en-US` here because this is the default localization we use for our components.

If the developer is building custom flows and is using the `clerk-captcha` element:
- The priority is 
1. `data-cl-language` attribute 
2. `auto`
- We default to `auto` because we don't know the localization used for the rest of the components here. The `auto` uses the localization that the visitor's browser is using, if it's supported.

anagstef: Thanks for pointing this out, I will try to make it more clear here. If you have any suggestions, feel free to apply any changes! 🙏

alexisintech: I can update the copy and let you review! @anagstef

alexisintech (approved):

PR #2068, opened 2025-02-27
Update client token docs for clarity in "how clerk works" guide

jescalan:
### 🔎 Previews:

- https://clerk.com/docs/pr/2068/how-clerk-works/overview#client-token

### What does this solve?

We had a customer write in mentioning that they were confused by the part of "how clerk works" that described the client token, as it was not clear how this cookie would be allowed to be sent cross-domain from your app to FAPI. When they realized that we require that you alias FAPI on to your domain using a CNAME, it was the piece they needed to understand it, but it took them a while to figure out.

### What changed?

This is reasonable feedback, and to hopefully help others to avoid running into the same confusion in the future, I added a bit more clarity to the section describing client tokens. While there, I added some extra detail around prod vs dev environments. It was not clear previously that the client token is only used in production environments, or why, and I think that should be more clear now.

### Checklist


docs/how-clerk-works/overview.mdx
```diff
@@ -245,13 +245,17 @@ Clerk's authentication model relies on two distinct tokens - a "client token" an
 - **Contents:** A Clerk-signed JWT containing:
   - `session_id`: Unique session identifier
   - `rotating_token`: [Anti-session-fixation](/docs/security/fixation-protection) token that rotates on each sign-in across any device
-- **Domain:** Set on your FAPI domain (`clerk.example.com` or `<slug>.accounts.dev`), rather than on your app domain. This is a security measure - for example if your app logs are leaked, they wouldn't contain client token values, since it's scoped to a different domain.
+- **Domain:** Set on your FAPI domain (`clerk.example.com`), rather than on your app domain (`example.com`). It's set on the FAPI domain as a security measure - for example, if your app logs are leaked, they wouldn't contain client token values, since it's scoped to a different domain.
```
myandrienko, line 248: To leave even less room for confusion, perhaps you could use `yourdomain.com` instead of `example.com` here, like you did in the paragraph below?

myandrienko (commented):
Hi @jescalan, I'm the guy who wrote in about being confused by this part of the docs. This update looks great, it clears up the confusing part while going even more in-depth on how this all works in development. So it's a 👍 from me, plus a little suggestion.

docs/how-clerk-works/overview.mdx
jescalan: since the rest of the doc uses `example.com`, i switched it in the opposite direction haha. thanks for catching!

alexisintech (approved):

PR #2103, opened 2025-03-14
docs: Simplify Expo quickstart

wobsoriano:
### 🔎 Previews:

- https://clerk.com/docs/pr/2103/quickstarts/expo

### What does this solve?

- Simplifies the Expo quickstart a bit

### What changed?

- Removed installation of `@clerk/types` step. Installing `@clerk/clerk-expo` includes `@clerk/types` by default, so they can just import if needed. Also, it raises the question on why do we have the step here but not in other SDKs.
- `<ClerkLoaded>` wrapping root layout means the entire app won't render until Clerk is loaded, which sounds like an anti-pattern. I tried our quickstart and removed this, everything is still expected to work 👍🏼 🫡 

### Checklist

### DIFF
--- a/docs/quickstarts/expo.mdx
+++ b/docs/quickstarts/expo.mdx
@@ -64,30 +64,6 @@ description: Add authentication and user management to your Expo app with Clerk.
     ```
   </CodeBlockTabs>
 
-  ## Install `@clerk/types` (optional)
-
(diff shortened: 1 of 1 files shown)


alexisintech (approved):
love it ! yeah not sure why the `clerkloaded` component was added in the first place, but i never questioned it when i should have!

PR #2118, opened 2025-03-19
fix: use `role` in public metadata example (to be consistent)

jfoshee:
I believe the intention here was to use `role` as an example of public metadata, and that using the `stripeId` was a copy/paste mistake from the private metadata example.


alexisintech (approved):
nice catch! 😸💖

PR #2125, opened 2025-03-21
Add Next.js route protection example with `auth().orgId`

LauraBeatris:
### 🔎 Previews:

-

### What does this solve?

- 

### What changed?

- 

### Checklist


alexisintech:
I changed the explanation because it wasn't accurate. It wasn't showing how to "prevent users from creating resources for other organizations they don't belong to." An example for that would have involved multiple organizations, and then using the user's org ID to only perform actions for the organization that matched that ID. But the example that was provided doesn't do that - so I updated the explanation to better fit the example provided. Let me know if you have any questions!

alexisintech (approved):
😸💖

LauraBeatris:
> I changed the explanation because it wasn't accurate. It wasn't showing how to "prevent users from creating resources for other organizations they don't belong to." An example for that would have involved multiple organizations, and then using the user's org ID to only perform actions for the organization that matched that ID. But the example that was provided doesn't do that - so I updated the explanation to better fit the example provided. Let me know if you have any questions!

@alexisintech that's correct, thanks for adjusting it!

PR #2137, opened 2025-03-28
Added clarity on JWT console command for postman-or-insomnia.mdx

vKenjo:
### 🔎 Previews:

-

### What does this solve?

Added more information on how to run the command properly (prerequisite and clarity for running the code). 

### What changed?

In the js command, I added some text to clarify if a user would not be getting the JWT because they are using the wrong credentials and not in their respective account to pull the token to be used in Postman or Insomnia.

### Checklist


alexisintech:
I think this is a given, and don't agree this is necessary context. But thank you for your time

PR #2139, opened 2025-03-29
Update add-onboarding-flow.mdx

ioilmio:
a very small but significant change, to be clear


alexisintech:
This is not correct. As the explanation says, if a user has completed their onboarding, they need to get redirected to the homepage. If they have not completed their onboarding, the guide has you set up your app to use middleware that will redirect the user to the onboarding page.
So,
are you onboarded? You don't need to go to onboarding page, so you just get redirected to homepage.
are you not onboarded? You get redirected to the onboarding page, so that you can finish onboarding.

ioilmio:
Thank you for clarifying

PR #2159, opened 2025-04-03
Add docs using `verifyWebhook` per-SDK

wobsoriano:
### 🔎 Previews:

- https://clerk.com/docs/pr/2159/webhooks/sync-data

### What does this solve?

- We released a new `verifyWebhook` that abstracts a way the logic required to verify a signature of Clerk webhook.  See https://github.com/clerk/javascript/pull/5468

### What changed?

- Updated the Clerk sync data page to use the new function, and added examples per-SDK

### Checklist


docs/webhooks/sync-data.mdx
```diff
@@ -60,12 +58,12 @@ These steps apply to any Clerk event. To make the setup process easier, it's rec
   To verify the webhook payload, you'll need your endpoint's **Signing Secret**. Since you don't want this secret exposed in your codebase, store it as an environment variable in your `.env` file during local development.
 
   1. On the endpoint's settings page in the Clerk Dashboard, copy the **Signing Secret**. You may need to select the eye icon to reveal the secret.
-  1. In your project's root directory, open or create an `.env` file, which should already include your Clerk API keys. Assign your **Signing Secret** to `SIGNING_SECRET`. The file should resemble:
+  1. In your project's root directory, open or create an `.env` file, which should already include your Clerk API keys. Assign your **Signing Secret** to `CLERK_WEBHOOK_SIGNING_SECRET`. The file should resemble:
```
royanger, line 61: Is this env var read by the SDKs like `CLERK_SECRET_KEY`? If so it should be added to https://clerk.com/docs/deployments/clerk-environment-variables

docs/webhooks/sync-data.mdx
```diff
@@ -254,7 +300,7 @@ These steps apply to any Clerk event. To make the setup process easier, it's rec
   }
   ```
 
-  To handle types manually, import the following types from your backend SDK (e.g., `@clerk/nextjs/server`):
+  To handle types manually, import the following types from your backend SDK (e.g., `@clerk/nextjs/webhooks`):
```
royanger, line 303: Is this available in other SDKs? eg, `import { EmailJSON} from '@clerk/tanstack-react-start/webhooks'`?

docs/webhooks/sync-data.mdx
```diff
@@ -105,122 +103,170 @@ These steps apply to any Clerk event. To make the setup process easier, it's rec
...
       ```ts {{ filename: 'app/api/webhooks/route.ts' }}
-      import { Webhook } from 'svix'
-      import { headers } from 'next/headers'
-      import { WebhookEvent } from '@clerk/nextjs/server'
+      import { verifyWebhook } from '@clerk/nextjs/webhooks'
 
       export async function POST(req: Request) {
-        const SIGNING_SECRET = process.env.SIGNING_SECRET
+        try {
+          const evt = await verifyWebhook(req)
```
royanger, line 113: A few questions about this function:
* if a customer can't set the env var, can it be passed as a parameter?
* if webhook verification fails, what response code is sent? Or more specifically will a failed verification prompt retries?
* how clear are the errors for a failed verification? eg, if there is a missing key is that clear?

royanger (commented):
Looks good! I left a couple of questions but this is great to see.

docs/webhooks/sync-data.mdx
wobsoriano: Yes! All types re-exported per-SDK in the `/webhooks` path

docs/webhooks/sync-data.mdx
wobsoriano: Yes it's read per SDK as well. Which section on that page do you think is best to put it? Or should we add a new one?

docs/webhooks/sync-data.mdx
wobsoriano: > if a customer can't set the env var, can it be passed as a parameter

Yes, the `verifyWebhook` accepts a 2nd argument where you can pass a `{ signingSecret: '' }` property. I'll add it!

> if webhook verification fails, what response code is sent? Or more specifically will a failed verification prompt retries?

Right now there is no response code 👀  Just whatever the `wh.verify` function throws hmm

> how clear are the errors for a failed verification? eg, if there is a missing key is that clear?

We have 3 errors:

1. [Missing webhook signing secret](https://github.com/clerk/javascript/blob/250b3117b409626b476c6d3ffbda13c06fcfcb54/packages/backend/src/webhooks.ts#L58)
2. [Missing svix header](https://github.com/clerk/javascript/blob/250b3117b409626b476c6d3ffbda13c06fcfcb54/packages/backend/src/webhooks.ts#L64)
3. Whatever the `wh.verify` throws 👀 

Let me know if we can improve this!

docs/webhooks/sync-data.mdx
royanger: I don't think it fits anywhere current, but I think this is @alexisintech's call.

docs/webhooks/sync-data.mdx
```diff
@@ -105,122 +103,117 @@ These steps apply to any Clerk event. To make the setup process easier, it's rec
...
+          console.error('Error verifying webhook:', err)
+          return reply.code(400).send('Error verifying webhook')
+        }
+      })
+      ```
+    </Tab>
 
-          let evt
+    <Tab>
+      ```ts {{ filename: 'server/api/webhooks.post.ts' }}
```
alexisintech, line 211: did you mean to have a slash here?
```suggestion
      ```ts {{ filename: 'server/api/webhooks/post.ts' }}
```

wobsoriano: Hi! It's a built-in Nuxt feature here - https://nuxt.com/docs/guide/directory-structure/server#matching-http-method

It means `POST /api/webhooks`

alexisintech (approved):
teaaamwoorrkk 👏

PR #2175, opened 2025-04-09
Update Validation script to check typedoc files

NWylynko:
### What does this solve?

- With the addition of typedoc generated markdown in to the docs, the existing validation script fails to check them

### What changed?

- Now if any docs files are using the new <Typedoc /> component the script will check for the typedoc folder and read in its contents, validating the links between docs.

### Checklist


package.json
```diff
@@ -11,7 +11,9 @@
     "lint:check-frontmatter": "node ./scripts/check-frontmatter.mjs",
     "lint:validation": "npm run build",
     "build": "tsx ./scripts/build-docs.ts",
-    "test": "vitest --silent"
+    "test": "vitest --silent",
+    "typedoc:link": "node scripts/typedoc-link.mjs",
```
LekoArts, line 15: We'll need to document this

scripts/build-docs.test.ts
```diff
@@ -95,6 +95,7 @@ const baseConfig = {
   docsPath: '../docs',
   manifestPath: '../docs/manifest.json',
   partialsPath: '../docs/_partials',
+  typedocPath: '../typedoc',
```
LekoArts, line 98: Is this the right path? I thought we're downloading it to `clerk-typedoc` folder

scripts/build-docs.ts
```diff
@@ -15,6 +15,7 @@ import path from 'node:path'
 import remarkMdx from 'remark-mdx'
 import { remark } from 'remark'
 import { visit as mdastVisit } from 'unist-util-visit'
+import { map, map as mdastMap } from 'unist-util-map'
```
LekoArts, line 18: Typo?

```suggestion
import { map as mdastMap } from 'unist-util-map'
```

scripts/build-docs.test.ts
NWylynko: It can be anything, as long as the tests use the same path.

package.json
LekoArts: Yes, I explained in this step here: https://github.com/clerk/clerk#5-set-up-local-docs-support
So I think for actually working on the docs the `typedoc:link` script isn't even necessary

LekoArts (approved):
Thanks! 🙇

PR #2177, opened 2025-04-11
Add validation for links inside typedoc files

NWylynko:
### What does this solve?

- When updating the validation script to validate typedoc files, I added to the script to pull out the headings / hashes, but failed to add validation to check the links inside the typedoc files.

### What changed?

- Added validation to ensure the links and hashes inside the typedoc markdown point to valid urls and headings

### Checklist

### DIFF
--- a/scripts/build-docs.test.ts
+++ b/scripts/build-docs.test.ts
@@ -1647,4 +1647,94 @@ description: Generated API docs
     // Should warn about missing src attribute
     expect(output).toContain('warning <Typedoc /> component has no "src" attribute')
   })
+
+  test('Should fail it typedoc file links to a non-existent file', async () => {
+    const { tempDir } = await createTempFiles([
+      {
+        path: './docs/manifest.json',
+        content: JSON.stringify({
+          navigation: [[{ title: 'API Doc', href: '/docs/api-doc' }]],
+        }),
+      },
+      {
+        path: './typedoc/api/client.mdx',
+        content: `[Non-existent File](/docs/non-existent-file)`,
+      },
+      {
+        path: './docs/api-doc.mdx',
+        content: `---
+title: API Documentation
+description: Generated API docs
+---
+
+# API Documentation
+
+<Typedoc src="api/client" />
+`,
+      },
+    ])
+
+    const output = await build(
+      createConfig({
+        ...baseConfig,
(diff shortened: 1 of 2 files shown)


alexisintech (approved):
LGTM 🤷‍♀️  but it looks like the build action is failing?

NWylynko:
> LGTM 🤷‍♀️ but it looks like the build action is failing?

yes because the typedoc generated markdown has links and hashes that are invalid

LekoArts:
Two things:

- I fixed all warnings in the "Build docs" step that I could. The ones remaining (https://github.com/clerk/clerk-docs/actions/runs/14447028814/job/40510014085?pr=2177#step:5:19) can't be fixed in the content. They are correct. I think the proper solution is that the build docs script is only validating/looking at files that the `<Typedoc />` component actually uses.
- In the main PR I had to resolve merge conflicts with the build script. Maybe I did it incorrectly and that's why the test build script is failing now, so you need to check if it's correct

NWylynko:
@LekoArts I fixed up the merge, but I'm a little confused with the remaining warnings, your saying the warnings are false? But when I try to use the `#factorverificationage` hash on the page its warning about it doesn't scroll to that section of the page, so seems like a valid warning?

https://github.com/user-attachments/assets/6ad7438d-6cdf-449b-8b2e-d830efa53948

LekoArts:
Thanks for the merge fix!

No, the warnings are correct in the sense that it's working as intended (from the script's perspective). But we're giving it the wrong input.

So, it complains about the missing hashes for links inside https://github.com/clerk/generated-typedoc/blob/main/types/active-session-resource.mdx?plain=1.

Since not every page of the docs is generated by Typedoc yet I'm replacing relative links / links to other Typedoc files with links to pages that are live in the docs. So `/docs/references/javascript/session#status` was originally `session-resource#status`. Which refers to this file: https://github.com/clerk/generated-typedoc/blob/main/types/session-resource.mdx?plain=1

That file has the anchor links. But both files are not used in production yet (so no `<Typedoc />` usage) so I'm only spending time fixing links etc. on files that are live.

So dunno what the best solution to this might be since it'll take a while until most pages are powered by Typedoc. That's why one of my suggestions was to skip the warnings for files that are not actually used right now.

PR #2183, opened 2025-04-14
Validate a hash is legit when it links to its own page

NWylynko:
### What does this solve?

The validation script expects all the links to be both a path, then possibly a hash (eg `/docs/quickstart#login`) but sometimes a link only has a hash (eg `#login`) pointing to another point on the page, so we need to validate those links. Currently the script just skips over those links

### What changed?

Added tests with links that only use a hash, then fixed the script, to pull in its own doc to validate against.

### Checklist


alexisintech (approved):
awesome ! thanks so much 🫶

PR #2187, opened 2025-04-15
(wip) (supabase) add guide on supporting orgs with supabase integration

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/2187/integrations/databases/supabase-and-orgs

### What does this solve?

- We have a blog on supporting Clerk orgs in a Clerk + Supabase app: https://clerk.com/blog/multitenancy-clerk-supabase-b2b and we wanted to bring this to the docs.

### What changed?

- 

### Checklist

### DIFF
--- a/docs/integrations/databases/supabase-and-orgs.mdx
+++ b/docs/integrations/databases/supabase-and-orgs.mdx
@@ -14,7 +14,7 @@ description: Learn how to integrate Clerk and support Clerk organizations in you
   exampleRepo={[
     {
       title: "Supabase, Next.js, and Clerk Demo",
-      link: "https://github.com/clerk/clerk-supabase-nextjs",
+      link: "(TODO)",
       icon: "github",
     },
   ]}
--- a/docs/integrations/databases/supabase-and-orgs.mdx
+++ b/docs/integrations/databases/supabase-and-orgs.mdx
@@ -113,20 +113,6 @@ ALTER TABLE tasks
 ADD COLUMN org_id text DEFAULT auth.jwt()->>'o.id';
 ```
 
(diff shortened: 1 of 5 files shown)


manovotny:
Closing this. The gap it targets is still real — the Supabase integration guide doesn't mention organizations anywhere across its 356 lines — but this branch isn't the fastest way there.

Two things. `supabase/config.toml` and `supabase/.gitignore` came along by accident, so 316 of the 495 added lines are a local Supabase scaffold rather than docs.

The bigger one is that the RLS policies don't work as written. `auth.jwt()->>'o.id'` looks for a top-level claim literally named `o.id`, but the org ID is nested — the JWT example on this page shows `o: { id: 'org_123' }`. So that expression is always NULL: `auth.jwt()->>'o.id' is null` is always true, `= org_id` never matches, and the new column's default is always NULL. Org support quietly does nothing. The TODO in the test section, about non-org tasks possibly still being visible, is that bug showing up.

The [blog post](https://clerk.com/blog/multitenancy-clerk-supabase-b2b) this was meant to be based on already has the correct form, and a cleaner shape — one `owner_id` column and a helper function instead of two columns and a four-clause OR:

```sql
create or replace function requesting_owner_id()
returns text language sql stable as $$
  select coalesce(
    (auth.jwt() -> 'o'::text) ->> 'id'::text,
    (auth.jwt() ->> 'sub'::text)
  );
$$;
```

DOCS-10212 stays open and now carries a note recording the `->>'o.id'` trap so it doesn't get rediscovered. Starting fresh from the blog post should be quicker than untangling this branch, which would also need every internal link repointed after the docs restructure.

PR #2190, opened 2025-04-16
Mark Core 1 as unsupported, clarify browser support

dstaley:
### 🔎 Previews:

- https://clerk.com/docs/pr/2190/upgrade-guides/long-term-support

### What does this solve?

Core 1 is now unsupported, so we want to call that out in our docs. Also, we want to clarify the actual supported browsers for Core 2, which targets "last 2 years". Finally, we want to specifically note that Core 1's browser support is frozen at the versions released for the final Core 1 release.

### What changed?

- This PR marks `Core 1` as unsupported on our Long Term Support page.
- Additionally, it clarifies the browser support for Core 1 and Core 2

### Checklist


alexisintech (approved):
LGTM! 😸💖

PR #2191, opened 2025-04-19
Add handshake animation to HCW

jescalan:
### 🔎 Previews:

- https://clerk.com/docs/pr/2191/how-clerk-works/overview#the-handshake

### What does this solve?

Handshake is a challenging architectural concept and I felt like adding an animation to clarify the flow might help readers to better contextualize how it works. Big props to the design team for nailing the graphics here on the first shot!

### Checklist


alexisintech:
Screenshot 2025-04-21 at 12 43 20 

"User requests app from server" I'm not really sure what this is trying to say - Maybe "User makes a request to the server with an expired token"? Also "w/expired token" should be "with expired token" as not everyone, especially non-english native speakers, will know what "w/" stands for.

<br />
<br />
Screenshot 2025-04-21 at 12 48 22

Should be "Server returns a redirect response to the browser. This includes a callback URL to the server."

Screenshot 2025-04-21 at 12 49 50 

Should be "FAPI checks the authentication state and if valid, returns a new session token when it returns the request to the server callback URL." we can say returns a new session token or mints a new session token or whatever jargon sounds better here, but we should make it clear that FAPI checks the client cookie to check the auth state. if it's valid, it will return a new session token, but if it's not valid, it won't.

docs/how-clerk-works/overview.mdx
```diff
@@ -304,6 +304,8 @@ However, server-rendered applications present a unique challenge. Server-to-serv
 1. FAPI receives the request with the `__client` cookie
 1. FAPI validates the authentication state and issues a new session token
 
+<Video src="/docs/images/how-clerk-works/handshake.mp4" width="1600" height="900" autoPlay muted loop playsInline />
```
bradlc, line 307: It won't actually make a difference because the _ratio_ is correct, but `width` and `height` should be set to the video's natural size:

```suggestion
<Video src="/docs/images/how-clerk-works/handshake.mp4" width="1920" height="1080" autoPlay muted loop playsInline />
```

jescalan:
Ok @alexisintech updated video with text edits

alexisintech (approved):
yup!! love it 💘

PR #2192, opened 2025-04-21
move redirects and move-doc script from clerk/clerk to clerk-docs

alexisintech:
### 🔎 Previews:

-

Related: https://github.com/clerk/clerk/pull/1137

### What changed?

Moves the redirects to `clerk-docs` so that one less PR needs to be made when creating redirects.
Moves the `move-doc` script to `clerk-docs` as well, as it makes more sense living here now that all the redirect logic is here, and the MDX files are here.

### Checklist


NWylynko (approved):
Great work!

PR #2196, opened 2025-04-23
Add security warning for Protect and SignedIn components

jescalan:
### 🔎 Previews:

- https://clerk.com/docs/pr/2196/components/protect#usage
- https://clerk.com/docs/pr/2196/components/control/signed-in

### What does this solve?

These two components return their children via source code due to the way that React hydration works. This change adds a note to the docs to clarify that if there is sensitive information that should be entirely inaccessible to the current user, these components are not enough to secure it.

### Checklist

### DIFF
--- a/docs/components/control/signed-in.mdx
+++ b/docs/components/control/signed-in.mdx
@@ -8,7 +8,7 @@ description: Conditionally render content only when a user is signed in.
 The `<SignedIn>` component offers authentication checks as a cross-cutting concern. Any children components wrapped by a `<SignedIn>` component will be rendered only if there's a User with an active Session signed in your application.
 
 > [!CAUTION]
-> This component only **visually hides** its children when the current user is not authenticated. The contents of its children remain accessible via the browser's source code even if the user fails the authentication check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform authentication checks on the server before sending the data to the client.
+> This component only **visually hides** its children when the current user is not authenticated. The contents of its children remain accessible via the browser's source code even if the user fails the authentication check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform [authentication checks](/docs/organizations/verify-user-permissions) on the server before sending the data to the client.
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


docs/components/control/signed-in.mdx
```diff
@@ -7,6 +7,9 @@ description: Conditionally render content only when a user is signed in.
 
 The `<SignedIn>` component offers authentication checks as a cross-cutting concern. Any children components wrapped by a `<SignedIn>` component will be rendered only if there's a User with an active Session signed in your application.
 
+> [!CAUTION]
+> This component only **visually hides** its children when the current user is not authenticated. The contents of its children remain accessible via the browser's source code even if the user fails the authentication check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform authentication checks on the server before sending the data to the client.
```
alexisintech, line 11:
```suggestion
> This component only **visually hides** its children when the current user is not authenticated. The contents of its children remain accessible via the browser's source code even if the user fails the authentication check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform [authentication checks](/docs/organizations/verify-user-permissions) on the server before sending the data to the client.
```

docs/components/protect.mdx
```diff
@@ -5,6 +5,9 @@ description: The Protect component is used for authorization. It only renders it
 
 The `<Protect>` component is used for authorization. It only renders its children when the current user has the specified [permission or role](/docs/organizations/roles-permissions) in the organization.
 
+> [!CAUTION]
+> This component only **visually hides** its children when the current user is not authorized. The contents of its children remain accessible via the browser's source code even if the user fails the authorization check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform authorization checks on the server before sending the data to the client.
```
alexisintech, line 9:
```suggestion
> This component only **visually hides** its children when the current user is not authorized. The contents of its children remain accessible via the browser's source code even if the user fails the authorization check. Do not use this component to hide sensitive information that should be completely inaccessible to unauthorized users. For truly sensitive data, perform [authentication checks](/docs/organizations/verify-user-permissions) on the server before sending the data to the client.
```

alexisintech (approved):
lgtm !

PR #2231, opened 2025-05-13
add nextjs LLM prompt to quickstart

mwickett:
### 🔎 Previews:

-

### What does this solve?

Adds our Next.js LLM prompt as a copyable markdown block (code block) to the nextjs quickstart guide.

It's at the bottom, but I'm not convinced that's the right place - but since we don't have any kind of straightforward to way to include it as a banner higher up, I thought we could start with this. 

### Checklist


mwickett:
Already handled elsewhere, so closing this out.

PR #2248, opened 2025-05-22
fix: Frontend API proxy example for Next.js

brkalow:
### 🔎 Previews:

- https://clerk.com/docs/pr/2248/advanced-usage/using-proxies

### What does this solve?

- 
- The proxy middleware example does not work properly as the proxy logic is not executed until `clerkMiddleware()` authenticates the request. This is unnecessary and causes issues in the handshake flow.

### What changed?

- 
The implementation has been improved to guarantee that proxy routes are always checked and handled before the Clerk middleware executes. This execution order is critical because it ensures that requests to the proxy path (__clerk) are correctly intercepted and routed to Clerk's Frontend API before any authentication middleware runs, preventing potential conflicts in request handling.

### Checklist


alexisintech (approved):
😸💖

PR #2257, opened 2025-05-27
Links from public to protected pages should use prefetch=false

alexisintech:
https://linear.app/clerk/issue/DOCS-9506/document-that-links-from-public-pages-to-protected-pages-should-be


SarahSoutoul (approved):
Content looks great! Just curious, is there a specific reason it appears three times on the page?

PR #2263, opened 2025-05-27
fix: Update b2b billing docs with more clarity on custom permissions

mzhong9723:
### 🔎 Previews:

-

### What does this solve?

The B2B billing docs are lacking details on how custom permission checks work in tandem with billing plans.

### What changed?

This PR adds a callout that explains custom permissions won't appear in the session jwt, API response, or pass `has()` checks unless the `feature` part of the permission key is part of the org's active plan, if billing is enabled.

### Checklist

### DIFF
--- a/docs/billing/b2b-saas.mdx
+++ b/docs/billing/b2b-saas.mdx
@@ -59,6 +59,9 @@ You can use Clerk's features, plans, and permissions to gate access to the conte
 
 The `has()` method is available for any JavaScript framework, while `<Protect>` is only available for React-based frameworks.
 
+> [!NOTE]
+> When billing is enabled, custom permissions will only appear in the session token and in API responses (including the result of the `has()` check) if the feature part of the permission key (`org:<feature>:<action>`) is a feature included in the organization's active plan. If the feature is not part of the plan, the `has()` check for permissions using that feature will return `false`.
+
 ### Example: Using `has()`
 
 Use the `has()` method to test if the organization has access to a **plan**:


docs/billing/b2b-saas.mdx
```diff
@@ -59,6 +59,9 @@ You can use Clerk's features, plans, and permissions to gate access to the conte
 
 The `has()` method is available for any JavaScript framework, while `<Protect>` is only available for React-based frameworks.
 
+> [!NOTE]
+> When billing is enabled, custom permissions will only appear in the session token and in API responses (including the result of the `has()` check) if the feature part of the permission key (`org:<feature>:<action>`) is a feature included in the organization's active plan. If the feature is not part of the plan, the `has()` check for permissions using that feature will return `false`.
```
panteliselef, line 63: should we mention the JWT itself along side `has` ? This is useful info for folks that sdks where `has` does not exist, but otherwise this looks good.

panteliselef (approved):
left a comment for a minor improvement

docs/billing/b2b-saas.mdx
mzhong9723: I mention "session token" but I can definitely include the term `JWT` to make it more obvious

NWylynko:
Just to make sure my understanding is current. If the billing feature is enabled then If I want to create a permission `org:support:edit-ticket` for the role `support-staff` in my SaaS, and I want to bill for a `pro` plan for orgs, it needs to have the `support` feature on that plan, even if its just hidden, for a `has({ permission: "support:edit-ticket" })` to return `true`?

mzhong9723:
> Just to make sure my understanding is current. If the billing feature is enabled then If I want to create a permission `org:support:edit-ticket` for the role `support-staff` in my SaaS, and I want to bill for a `pro` plan for orgs, it needs to have the `support` feature on that plan, even if its just hidden, for a `has({ permission: "support:edit-ticket" })` to return `true`?

@NWylynko yes, if you want org members with role `support-staff` to have that permission, then `support` has to be a feature on the `pro` plan (assuming that the org is on the `pro` plan). By default we add all custom permission features to the free orgs plan as well.

NWylynko:
Sorry can you expand on what you mean by that all the custom permission features get added to the free orgs plan? that sounds like the free plan just gets all the features? but that sounds wrong

NWylynko:
I've wrote this:


When using the Clerk Organizations and Billing features together, if you want to use the permissions authorization checks, it will now link with the billing feature check. For example for a user to have the custom permission `org:finance:approve-purchase` then the plan that organization is subscribed to would need to have the feature `finance`. This gives you the control to sell features in your application based on the plan, but tied to only certain roles as controlled by your users.

These custom permissions will show up in the API responses and session tokens (JWT) if both the user has the permission based on their role, and their active organization has subscribed to that plan. You can also check the access using the `has()` or `<Protect />` functions.


It's def not perfect but I recon giving an example and explaining why this happens would be helpful.

mzhong9723:
@NWylynko Updated the docs with more detail and an example!

PR #2273, opened 2025-05-29
Add new sdk-messages page for auth() called without clerk middleware

guilherme6191:
### 🔎 Previews:

-

### What does this solve?

- 

This new page is meant to guide users that often fall into the same error but can't initially figure it out. Given how Next.js 404 mechanism works -- it's common to see the `auth() was called but Clerk can't detect usage of clerkMiddleware()` error message when invalid requests are being made to asset paths.
This error message is not wrong, but if users lack the specific Next.js context here, they might not understand 100% what's the source of the issue.

### What changed?

- 
- Added new page to guide users through a specific error message scenario.

### Checklist

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -3651,6 +3651,10 @@
                   }
                 ]
               ]
+            },
+            {
+              "title": "SDK Messages",
+              "href": "/docs/sdk-messages/auth-was-called-but-clerk-cant-detect-usage-of-clerkmiddleware"
             }
           ]
         ]
--- a/docs/sdk-messages/auth-was-called-but-clerk-cant-detect-usage-of-clerkmiddleware.mdx
+++ b/docs/sdk-messages/auth-was-called-but-clerk-cant-detect-usage-of-clerkmiddleware.mdx
@@ -1,0 +1,168 @@
+---
+title: "Clerk: auth() was called but Clerk can't detect usage of clerkMiddleware()"
+description: Learn how to handle auth() was called but Clerk can't detect usage of clerkMiddleware() when getting 404 errors in Next.js.
+---
+
+**Error when requests that are not covered by clerkMiddleware get 404 return in a Clerk + Next.js App Router setup**
+
+## Why this error occurs
+
+This error commonly appears when using Clerk with Next.js and the `clerkMiddleware()` helper is configured *not* to run on static asset routes. Which is part of the default setup.
+
+### First scenario
+
+The most common cause of this error is when `auth()` is called **without** a `ClerkProvider` wrapping the component tree where Clerk components are used. This is a common pattern with React context and also used by many libraries.
+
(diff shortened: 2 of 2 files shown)


docs/manifest.json
```diff
@@ -3651,6 +3651,10 @@
                   }
                 ]
               ]
+            },
+            {
+              "title": "SDK Messages",
```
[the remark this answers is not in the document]
royanger: I think the pattern we will go with is the same as the Next docs -- this isn't in the table of contents and can only be found by the link in the error message. Once we have this up for review Alexis can weigh in though. For now let's not add to the manifest.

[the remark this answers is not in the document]
guilherme6191: No problem. Heads up that this warning manifest seems to break the checks/build though.

royanger:
Thanks for the commit @NWylynko!

Looking at that, I wonder if we use something more like the following for the directory Alexis?

```
'references/nextjs/errors/auth-was-called-but-clerk-cant-detect-usage-of-clerkmiddleware.mdx': ['doc-not-in-manifest'],
```

docs/references/nextjs/errors/auth-was-called.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Select the appropriate solution based on your use case:
+
+<Tabs items={["Correct the static asset paths (recommended)", "Scope the <ClerkProvider> to a sub-layout"]}>
+  <Tab>
+    The ideal solution is to fix any incorrect asset paths, as this is the **root cause** of the error and the cleanest fix. Ensure that all images, icons, or files are requested using valid URLs. When these requests succeed, no 404 is triggered, and the Clerk provider context works as expected.
+  </Tab>
+
+  <Tab>
+    This fix will prevent the `auth()` error as it moves `<ClerkProvider>` from the root layout. However, it does not resolve the underlying issue, which is a reference to a missing static asset.
```
alexisintech, line 84: so will the error still occur? i'm confused why it's being offered as a "fix" but then says it won't actually fix the issue

scripts/build-docs.ts
```diff
@@ -81,6 +81,7 @@ async function main() {
         'maintenance-mode.mdx': ['doc-not-in-manifest'],
         'deployments/staging-alternatives.mdx': ['doc-not-in-manifest'],
         'references/nextjs/usage-with-older-versions.mdx': ['doc-not-in-manifest'],
+        'sdk-messages/auth-was-called-but-clerk-cant-detect-usage-of-clerkmiddleware.mdx': ['doc-not-in-manifest'],
```
alexisintech, line 84:
```suggestion
        'references/nextjs/errors/auth-was-called.mdx': ['doc-not-in-manifest'],
```

docs/references/nextjs/errors/auth-was-called.mdx
royanger: The error from Clerk will be resolved. The app will still have a failed 404 network request for the missing asset, but it will resolve all Clerk errors.

PR #2291, opened 2025-06-04
Updated docs with passkey limitation information (see context section in description)

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/2291/authentication/configuration/sign-up-sign-in-options.mdx 
- https://clerk.com/docs/pr/2291/custom-flows/passkeys
- https://clerk.com/docs/pr/2291/advanced-usage/satellite-domains

### What does this solve?

This PR documents a known limitation regarding passkeys not being supported on satellite domains, clarifying that passkeys are tied to the domain they are created on. This helps users understand why passkeys won’t work across different domains (even if they are considered "satellite" domains by Clerk) and prevents confusion during implementation.

**More context @jescalan:** While investigating a bit more around this, I thought the wording could be misleading beyond satellite domains. How do passkeys behave in the case of subdomains for e.g? From my understanding, passkeys set on a lambda domain would work on a subdomain (`your-domain.com` and `accounts.your-domain.com`). I wanted to convey this nuance in the docs to better understand the need for a workaround in development AND the absence of this limitation in production due to the nature of Clerk's setup (if you're not using satellite domains of course). 

Hopefully, this is correct. Otherwise, it would make this part in the docs incorrect:

Screenshot 2025-06-04 at 1 42 47 pm

### What changed?

- Updated the documentation to include an `Important` callout note for users planning to use satellite domains, encouraging them to avoid using passkeys in such scenarios.
- Clarified passkeys limitations in general across diff domains vs subdomains, and how that plays out in dev vs prod environments. 
- In the `Warning` callout within the [Satellite Domains docs](https://clerk.com/docs/advanced-usage/satellite-domains), added a note about passkeys.
- Removed `in development` in headings mentioning passkeys to make this more general. 

### Checklist


docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
-Passkeys are tied to the domain they are created on and **cannot be used across different domains**. For example:
+Passkeys are tied to the domain they are created on and **cannot be used across different domains**. However, passkeys **do work on subdomains if they are registered on the root domain**, so all passkeys registered on `domain.com` would work for any subdomains. For example:
 
-- Passkeys created on `localhost` will only work on `localhost`.
-- Passkeys created on your Account Portal (e.g., `your-app.accounts.dev`) will only work on that domain.
+- Passkeys created on `your-domain.com` **cannot be used** on `your-domain-admin.com` (different domains).
+- Passkeys created on `your-domain.com` **can be used** on `accounts.your-domain.com` (subdomain of the same root domain).
+- Passkeys created on `staging1.your-domain.com` **cannot be used** on `staging2.your-domain.com` (sibling subdomains).
```
jescalan, line 5: This is a much more clear explanation - I just want to make sure you're very confident this is true or have tested it?

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
...
 
-- Use Clerk's [Components](/docs/components/overview), [Elements](/docs/customization/elements/overview), or [custom flows](/docs/custom-flows/overview), instead of using Account Portal.
-- Create the passkey directly through Account Portal instead of your local application.
+If you're **not** using satellite domains:
 
-**This issue does not affect production environments**, as your Account Portal will be hosted on a subdomain of your main domain (e.g., `accounts.your-domain.com`), enabling passkeys to work seamlessly across your application.
+- **In development**, you can either:
+
+  - Use Clerk's [Components](/docs/components/overview), [Elements](/docs/customization/elements/overview), or [custom flows](/docs/custom-flows/overview), instead of using Account Portal. This means the passkey is created and used entirely on the same domain (e.g. `localhost`), avoiding the need to transfer passkeys across domains.
+  - Create the passkey directly through Account Portal instead of your local application to keep passkeys tied to the Account Portal's domain.
```
jescalan, line 15: The way I'm reading this is "only use AP or never use AP if you want to use passkeys", is that right? If so, I wonder if we'd want to take a stronger stance here tbh. Exactly replicating AP with a custom page is like a 5 minute task, and feels like a much safer approach to me than locking yourself into AP indefinitely, when realistically most businesses end up moving over to a custom page/flow as they scale for customization/branding reasons.

docs/advanced-usage/satellite-domains.mdx
```diff
@@ -6,6 +6,8 @@ description: Learn how to share sessions across different domains by adding sate
 > [!WARNING]
 > This guide addresses authentication across different domains with shared sessions. For example, `example-site.com` and `example-site-admin.com`.
 >
+> Due to the nature of satellite domains, it is not recommended to use **passkeys** as a form of authentication. [Learn more about domain restrictions for passkeys](/docs/custom-flows/passkeys#domain-restrictions-for-passkeys).
```
jescalan, line 9:
```suggestion
> It is not recommended to use **passkeys** as a form of authentication. [Learn more about domain restrictions for passkeys](/docs/custom-flows/passkeys#domain-restrictions-for-passkeys).
```

jescalan (approved):
This is really good, thank you!

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
-Passkeys are tied to the domain they are created on and **cannot be used across different domains**. For example:
+Passkeys are tied to the domain they are created on and **cannot be used across different domains**. However, passkeys **do work on subdomains if they are registered on the root domain**, so all passkeys registered on `domain.com` would work for any subdomains. For example:
 
-- Passkeys created on `localhost` will only work on `localhost`.
-- Passkeys created on your Account Portal (e.g., `your-app.accounts.dev`) will only work on that domain.
```
mzhong9723, line 4: ❓ What's the reason to remove these two examples?

docs/_partials/authentication/passkeys-restriction.mdx
mzhong9723: This case is tricky - the important part is that the Clerk app's primary domain must be `your-domain.com` in order for passkeys to work on subdomains. So it's possible that a passkey _created on_ `staging1.your-domain.com` can be used for `staging2.your-domain.com` as long as `your-domain.com` is the primary domain for the app. 

Here's the backend logic for this: https://github.com/clerk/clerk_go/blob/8bfcac7105bd40e1348a438bbdbfc3a7390b2c24/api/shared/passkeys/service.go#L101.

docs/_partials/authentication/passkeys-restriction.mdx
mzhong9723: Is the first bullet point just saying to use develop on `localhost`? I would expect this is what most developers would be doing anyways. I think the important thing to make clear is that passkeys can't be shared between localhost and development instance account portal, and there's no way around that.

docs/_partials/authentication/passkeys-restriction.mdx
SarahSoutoul: @mzhong9723 Yeah that's what I tried to convey with this sentence here: 

**However, passkeys **do work on subdomains if they are registered on the root domain**, so all passkeys registered on `domain.com` would work for any subdomains.**

To my understanding of the logic you linked above, if the app's primary domain was `staging1.your-domain.com`, then, passkeys created there would not work on `staging2.your-domain.com`, because the RpID would be too specific. But if the passkey is set on `your-domain.com`, then passkeys become usable across these subdomains.

So that last example stands but could update to: 

- Passkeys created on `staging1.your-domain.com` **cannot be used** on `staging2.your-domain.com` **as long as `your-domain.com` is the primary domain for the app**. (sibling subdomains).

docs/_partials/authentication/passkeys-restriction.mdx
SarahSoutoul: @mzhong9723 I think I'll prob re-add those under the explanations for a `development` environment linked to your comment [here](https://github.com/clerk/clerk-docs/pull/2291/files#r2140656206)

SarahSoutoul: Added this

docs/_partials/authentication/passkeys-restriction.mdx
SarahSoutoul: Added changes

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
...
+- Passkeys created on `your-domain.com` **cannot be used** on `your-domain-admin.com` (different domains).
+- Passkeys created on `your-domain.com` **can be used** on `accounts.your-domain.com` (subdomain of the same root domain).
+- Passkeys created on `staging1.your-domain.com` **cannot be used** on `staging2.your-domain.com` as long as `your-domain.com` is the primary domain for the app (sibling subdomains).
 
-To work around this in development, you can either:
+> [!IMPORTANT]
+> This means that **[satellite domains](/docs/advanced-usage/satellite-domains) do not support passkeys** because they are considered different domains. If you plan to use satellite domains - in either development or production - you should avoid using passkeys, as they won't be portable between your primary domain and your satellite domains.
```
alexisintech, line 8: since the sentence below this callout is "If you're **not** using satellite domains", then the paragraph before that _shouldn't_ be a callout, it should be in the same flow of the document. so i'm removing the callout here, and ordering the information as

**If you're using [satellite domains](/docs/advanced-usage/satellite-domains)**,

If you're **not** using satellite domains:

```suggestion
**If you're using [satellite domains](/docs/advanced-usage/satellite-domains)**, in both development and production, they do not support passkeys because they are considered different domains. With satellite domains, you should avoid using passkeys as they won't be portable between your primary domain and your satellite domains.
```

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
...
 
-- Use Clerk's [Components](/docs/components/overview), [Elements](/docs/customization/elements/overview), or [custom flows](/docs/custom-flows/overview), instead of using Account Portal.
-- Create the passkey directly through Account Portal instead of your local application.
+If you're **not** using satellite domains:
 
-**This issue does not affect production environments**, as your Account Portal will be hosted on a subdomain of your main domain (e.g., `accounts.your-domain.com`), enabling passkeys to work seamlessly across your application.
+- **In development**, you can either:
+
+  - Use Clerk's [Components](/docs/components/overview), [Elements](/docs/customization/elements/overview), or [custom flows](/docs/custom-flows/overview), instead of the Account Portal. This ensures the passkey is created and used entirely on your development domain, so passkeys created on `localhost` will only work on `localhost`. **This is the recommended approach**
```
alexisintech, line 14:
```suggestion
  - **The recommended approach**. Use Clerk's [components](/docs/components/overview), [Elements](/docs/customization/elements/overview), or [custom flows](/docs/custom-flows/overview), instead of the [Account Portal](/docs/account-portal/overview). This ensures the passkey is created and used entirely on your development domain, so passkeys created on `localhost` will only work on `localhost`.
```

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
...
+  - Create a passkey directly through the Account Portal instead of your local application to keep it tied to the portal's domain. Passkeys created on your Account Portal (e.g., `your-app.accounts.dev`) will only work on that domain, which can cause issues if you switch between `localhost` and the Account Portal during developemnt. If you choose this approach, ensure all testing happens on the same domain where the passkey was created.
```
alexisintech, line 15:
```suggestion
  - Create a passkey directly through the Account Portal instead of your local application to keep it tied to the Account Portal's domain. Passkeys created on your Account Portal (e.g., `your-app.accounts.dev`) will only work on that domain, which can cause issues if you switch between `localhost` and the Account Portal during development. If you choose this approach, ensure all testing happens on the same domain where the passkey was created.
```

docs/_partials/authentication/passkeys-restriction.mdx
```diff
@@ -1,11 +1,17 @@
...
+
+- **In production,** this limitation typically **does not apply** because your Account Portal is usually hosted on a subdomain of your main domain (e.g. `accounts.your-domain.com`), enabling passkeys to work seamlessly across your app. However, as stated above, if you use **satellite domains**, this issue **does apply** and it is not recommended to use passkeys in that scenario.
```
alexisintech, line 17:
```suggestion
- **In production,** your Account Portal is usually hosted on a subdomain of your main domain (e.g. `accounts.your-domain.com`), enabling passkeys to work seamlessly across your app. However, as stated above, if you use **satellite domains**, passkeys will not work as intended.
```

docs/_partials/authentication/passkeys-restriction.mdx
SarahSoutoul: Makes sense! Thanks!

alexisintech (approved):
awesome work sarah 😸💖

PR #2305, opened 2025-06-11
include the prompts in the dist

NWylynko:
### What does this solve?

- recently a new `prompts` folder was added to the repo, used by the `LLMPrompts` component, the folder needs to be included in the dist for clerk.com to use it in https://github.com/clerk/clerk/pull/1074

### What changed?

- Like the typedoc markdown I've added in steps to load in the prompts, write them out to dist, and update the src in the llmprompts to point to the output location

### Checklist


mwickett (approved):
Makes sense!

PR #2320, opened 2025-06-17
SDK scoping the Billing docs

NWylynko:
### 🔎 Previews:

- https://clerk.com/docs/pr/nick-billing-sdk-scoped/tanstack-react-start/billing/b2c-saas
Screenshot 2025-06-18 at 4 29 33 AM

- https://clerk.com/docs/pr/nick-billing-sdk-scoped/js-frontend/billing/b2b-saas
Screenshot 2025-06-18 at 4 30 27 AM

### What does this solve?

- This would be the first usage of the new sdk scoping tools, by setting the sdk in the frontmatter
- Adding code examples for Astro (doesn't support plans and features yet), Chrome Extension (PricingTable is broken), JS Frontend, Nuxt, Vue, React, React Router, Remix, Tanstack, Expressjs, Fastify and Expo

### What changed?

- This brings in a code example for each sdk that `<PricingTable />` is supported in
- When a user has an unsupported sdk selected these pages are hidden in the sidebar
- The users current active sdk is in the url of these pages

### Checklist


alexisintech:
This is looking fantastic - thanks so much for getting on this Nick! I think the authoring makes sense, and the partials make sense because the examples are used across more than one guide. I like the variant selector in the sidebar a lot - I'm glad it's synced with the SDK selector! Awesome work all around :)

The only thing that I noticed is that the redirect flash isn't the greatest UX:

https://github.com/user-attachments/assets/e3981d26-f68b-45de-9af7-c09aa4412499

NWylynko:
> This is looking fantastic - thanks so much for getting on this Nick! I think the authoring makes sense, and the partials make sense because the examples are used across more than one guide. I like the variant selector in the sidebar a lot - I'm glad it's synced with the SDK selector! Awesome work all around :)
> 
> The only thing that I noticed is that the redirect flash isn't the greatest UX:
> Screen.Recording.2025-06-20.at.10.56.06.mov


https://github.com/user-attachments/assets/3a8af45a-e97b-4773-a52d-904d6870a5fe

Running a production build locally, which is much closer in page load times to live production deployments with the caching, it's pretty quick, but this is something I can improve in a `clerk` repo pr.

alexisintech:
The sidenav isn't expanding for the active page

Screenshot 2025-06-20 at 12 36 18

docs/_partials/billing/checking-feature-using-protect.mdx
```diff
@@ -0,0 +1,214 @@
+<If sdk="astro">
+  > [!WARNING]
+  > Astro SDK hasn't been updated to support "feature" in the `Protect` component yet.
```
alexisintech, line 3:
```suggestion
  > Astro SDK hasn't been updated to support `feature` in the `Protect` component yet.
```

docs/_partials/billing/checking-plan-using-protect.mdx
```diff
@@ -0,0 +1,204 @@
+<If sdk="astro">
+  > [!WARNING]
+  > Astro SDK hasn't been updated to support "plan" in the `Protect` component yet.
```
alexisintech, line 3:
```suggestion
  > Astro SDK hasn't been updated to support `plan` in the `Protect` component yet.
```

docs/billing/b2b-saas.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Clerk billing for B2B SaaS
 description: Clerk billing is a feature that allows you to create and manage plans and features for your application.
+sdk: astro, chrome-extension, js-frontend, nextjs, nuxt, vue, react, react-router, remix, tanstack-react-start, expressjs, fastify, expo, js-backend
```
SarahSoutoul, line 4: Is there a specific order to follow when listing those?

NWylynko:
> The sidenav isn't expanding for the active page
> Screenshot 2025-06-20 at 12 36 18

I'm not able to reproduce this

SarahSoutoul:
> > The sidenav isn't expanding for the active page
> > Screenshot 2025-06-20 at 12 36 18
> 
> I'm not able to reproduce this

Yeah got the same issue - when reloading the page for the active SDK - the sidenav should expand with the page we're currently on. Have included a screen recording to show the behavior.

Uploading Screen Recording 2025-06-20 at 11.29.18 am.mov…

NWylynko:
> The sidenav isn't expanding for the active page
> Screenshot 2025-06-20 at 12 36 18

open a pr https://github.com/clerk/clerk/pull/1294 massive thanks for pointing this out @alexisintech and @SarahSoutoul

royanger:
Not sure this is the best place to mention this, but say this comment and wanted to mention a couple of things:

> Clerk billing is enabled for all SDKs 
Billing can't be used in Expo Mobile until we release custom flows -- the components don't work.
Billing can never work in Chrome Extensions, at least with Stripe as the only gateway. Stripe requirs remote code which is forbidden in Chrome Extensions. We strip out `@stripe/stripejs` already and I opened ticket ([COM-954](https://linear.app/clerk/issue/COM-954/billing-component-code-in-chrome-extension)) about hiding or removing the Billing pieces from the components.

royanger:
If I am on https://clerk.com/docs/pr/nick-billing-sdk-scoped/tanstack-react-start/billing/b2c-saas and, as a customer, want to select iOS its not on the right, so I try the dropdown.

I end up on https://clerk.com/docs/pr/nick-billing-sdk-scoped/quickstarts/ios and then need to select Billing and then either B2B or B2C. Then I get a screen that doesn't give me iOS as a choice or tell me why it doesn't. Its a confusing flow for a customer.

https://share.cleanshot.com/FMpCmly4

royanger:
More about the Billing overview the SDK list 
CleanShot 2025-06-20 at 16 16 46

Also, can we control the order? Having Chrome Extension at 2, JS at 3, and more doesn't feel like a good order.

Something like Next, React, Expo (with no-mobile support note on overview or actual page), React Router, Astro, Tanstack, Remix, Chrome Extension, Nuxt, Vue, JS feels more accurate in terms of popularlity.

royanger:
> Would it be better if Expo & Chrome Extension is fully removed?

Expo Web does work. Chrome Extension should probably be removed.

> Yea this is tough, what do you think should happen when selecting an SDK the guide doesn't support?
> - Could send the user back to the grid selection? (eg [clerk.com/docs/pr/nick-billing-sdk-scoped/billing/b2c-saas](https://clerk.com/docs/pr/nick-billing-sdk-scoped/billing/b2c-saas))

I would expect this, with a very clear callout telling me that SDK doesn't support the feature.

> I've updated the order to match your suggestion. But I am thinking of make the dropdown selector order control the order of the SDKs on this page.

Having them match is probably a good idea, though the main dropdown desperately needs updates.

> Largely this component's purpose is to be informative, just saying this guide is supported in x, y, z. But it wouldn't make sense to not link to those versions. I'm open to design suggestions.

This PR is moving the tabs that are currently above a code block (for Next, React, etc) to the right sidebar. That's my large concern here.

docs/_partials/billing/checking-feature-using-protect.mdx
```diff
@@ -0,0 +1,196 @@
+<If sdk="astro">
+  > [!WARNING]
+  > Astro SDK hasn't been updated to support `feature` in the `Protect` component yet.
```
SarahSoutoul, line 3: Do we want to be showing code if Astro doesn't support this yet?

SarahSoutoul: Asking cause for JS for e.g., since it doesn't support the `<Protect>` component, there is just a warning callout and no code snippet. Wouldn't we want to do sth similar with Astro or is it different?

docs/billing/b2b-saas.mdx
```diff
@@ -57,14 +57,12 @@ The `has()` method is available for any JavaScript framework, while `<Protect>`
...
 
 ```jsx
-const { has } = await auth()
 const hasPremiumAccess = has({ plan: 'gold' })
 ```
 
 Or a **feature**:
 
 ```jsx
-const { has } = await auth()
```
SarahSoutoul, line 76: Have removed those from the b2b page as you removed them in the b2c one @NWylynko to be consistent

docs/billing/b2b-saas.mdx
```diff
@@ -79,8 +77,6 @@ The [`has()`](/docs/references/backend/types/auth-object#has) method checks if t
 >
 > For example, say you want to check if an organization member has the custom permission `org:teams:manage`, where `teams` is the feature. The user's organization must be subscribed to a plan that has the `teams` feature for authorization checks to work. If the user's organization is not subscribed to a plan that has the `teams` feature, the authorization check will always return `false`, even if the user has the custom permission.
 
-The following example accesses the `auth` object and the `has()` method using the [Next.js-specific `auth()` helper](/docs/references/nextjs/auth).
```
SarahSoutoul, line 91: Same here - removed this to be consistent with b2c page

docs/_partials/billing/checking-feature-using-protect.mdx
SarahSoutoul: what about `permission`?

docs/_partials/billing/checking-plan-using-has-function.mdx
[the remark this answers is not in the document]
NWylynko: I think I was generally following the same order, except I put the non-supported call-outs at the bottom. I've got no issue with re-ordering to whatever makes sense, but I do think trying to keep everything in sync is a little bit  unnecessary.

docs/_partials/billing/checking-plan-using-protect.mdx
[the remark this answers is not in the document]
NWylynko: I'm open to ideas on how we could support this, definitely less code examples to manage would be good. I don't want to remove that warning as I think it's helpful to not get caught out that your trying to use an `<If />` for a certain sdk that will never show up as the page isn't marked to support that sdk. But one idea would be to add a `sdkFilter` prop in the `<Include />` and only pull out `<If />` from the partial that are in the `sdkFilter`.

docs/_partials/billing/checking-feature-using-protect.mdx
NWylynko: permission is good

SarahSoutoul: Yes tested it. All good!

SarahSoutoul:
@NWylynko I was investigating [this ticket](https://linear.app/clerk/issue/DOCS-10614/feedback-for-componentsprotect) as it pertains to this PR. I'm confused as Expo doesn't support the `<PricingTable>` component , so there isn't any way really to test out the `<Protect>` component with the "plan" and "feature" props. Surely, if one cannot subscribe to a plan within an Expo app, there is no reason to be using the "plan" and "feature" props at all. If that's the case, then we need to change some of the stuff in this PR, and make changes to this [page](https://clerk.com/docs/components/protect) too. 

Tagging you for visibility and thoughts too @alexisintech

royanger:
> One quick thing to point out. Is the same clerk application could be used across both a nextjs website and an expo app for example. So a user could be signed in on the website, subscribe using the `<PricingTable />` on the website. then switch back over to the mobile expo app.
> 
> I am all for adding helpful call outs, I can do my own testing of Expo support for these features and add some callouts.

royanger:
>     * Using the `<Protect>` component using the "plan" and "feature" props also worked. But, that was on Expo Web. As @royanger said above, it doesn't work on Expo Mobile tho, so we prob want to have some type of callout for this.
> 
>     * @royanger is it all of Clerk billing not working on Expo mobile, including the `<Protect>` component. If so, that would explain this ticket right [here](https://linear.app/clerk/issue/DOCS-10614/feedback-for-componentsprotect). Otherwise, I'm not sure as it definitely worked for me but that was on Expo web.

I commented in the Linear, but I would expect `<Protect />` to work in Expo Web and Mobile. This might be a bug and worth asking about in the team user help channel.

Any component and renders something in the UI (like `<SignIn />` or `<PricingTable />` does not work in Expo Mobile (the components use things like `<div>` and `<p>` which are not in Expo Mobile. That means currently a user can't subscribe, view plans, cancel plans, and so on since that's all just in components. There is work underway (maybe close to completion?) to add hooks/custom flows which might enable Billing in Expo Mobile. I say maybe as I haven't looked and don't know the state of hooks and what's being added or not.

royanger:
Sorry, I'm not sure how I ended up closing this. My mistake.

PR #2323, opened 2025-06-18
Track and invalidate linked partials and typedocs in dev cache

NWylynko:
### What does this solve?

- dev script isn't currently tracking and invalidating cache when a `<Typedoc />` or `<Include />` of a partial is used and edited

### What changed?

- Adds the typedocs and partials to be tracked
- Invalidates the pages that use them to be re-generated

### Checklist


package.json
```diff
@@ -12,7 +12,7 @@
     "lint:check-frontmatter": "node ./scripts/check-frontmatter.mjs",
     "lint:validation": "npm run build",
     "build": "tsx ./scripts/build-docs.ts",
-    "dev": "tsx ./scripts/build-docs.ts --watch",
+    "dev": "tsx watch ./scripts/build-docs.ts --watch",
```
jescalan, line 15: This seems weird that it would need both watch flags...

scripts/build-docs.ts
```diff
@@ -827,5 +881,7 @@ template: wide
     await fs.rename(config.distTempPath, config.distFinalPath)
   }
 
+  if (abortSignal?.aborted) throw new Error('Build aborted')
```
jescalan, line 884: This abortsignal thing seems crazy there are so many instances of it and it's so frequent. Is this really the best way?

jescalan (approved):

package.json
NWylynko: The first `watch` is to tell `tsx` to reload and restart the whole script if any of the `.ts` files change (eg pulling in main having an updated script). The second `--watch` is to run the script in dev mode holding the markdown in memory and all that.

scripts/build-docs.ts
NWylynko: I've cleaned it up to just `abortSignal?.throwIfAborted()` but this is largely just how abort controllers work, just check it throughout execution and exit early if need be.

PR #2337, opened 2025-06-24
How Clerk works: Add auth model summary

danielbachhuber:
Clarifies "combines the benefits of both stateful and stateless approaches while mitigating their respective drawbacks" by adding a table that summarizes the benefits and drawbacks.


alexisintech:
we don't necessarily want to say that clerk has "no" drawbacks - I appreciate the idea, but I think we're going to pass on this addition!

PR #2315, opened 2025-06-17
Updated TanStack Start quickstart and examples

wobsoriano:
### 🔎 Previews:

- https://clerk.com/docs/pr/rob-tanstack-devinxi/quickstarts/tanstack-react-start
- https://clerk.com/docs/pr/rob-tanstack-devinxi/references/tanstack-react-start/create-clerk-handler
- https://clerk.com/docs/pr/rob-tanstack-devinxi/references/tanstack-react-start/custom-sign-in-or-up-page
- https://clerk.com/docs/pr/rob-tanstack-devinxi/references/tanstack-react-start/custom-sign-up-page
- https://clerk.com/docs/pr/rob-tanstack-devinxi/references/tanstack-react-start/read-session-data
- https://clerk.com/docs/pr/rob-tanstack-devinxi/backend-requests/custom-session-token#use-the-custom-claims-in-your-application
- https://clerk.com/docs/pr/rob-tanstack-devinxi/references/backend/types/auth-object#example-use-get-token-in-the-backend

### What does this solve?

- TanStack Start 1.121.0 introduced [a breaking change](https://github.com/TanStack/router/discussions/2863#discussioncomment-13104960) and our guides and references are outdated. This PR updates all of related usage to match new API.

### What changed?

- TanStack Start quickstart, guides and references, and other Backend usage with TanStack related examples are updated to use new TanStack Start API

⚠️ Note: Please do not merge until https://github.com/clerk/javascript/pull/6115 has been released.

⚠️ Dashboard PR: https://github.com/clerk/dashboard/pull/5930

✅ Updated the demo repo: https://github.com/clerk/clerk-tanstack-react-start-quickstart

### Checklist

- [x] I have clicked on "Files changed" and performed a thorough self-review
- [x] All existing checks pass

PR #2346, opened 2025-06-26
Tanstack Start - Remove splat from sign-in page file route path

lawandothman:
### 🔎 Previews:

-

### What does this solve?
- As part of the devinxi migration I see that some docs were updated in https://github.com/clerk/clerk-docs/pull/2315 
- Those updates included moving from `/sign-in/$` to `/sign-in`
- The new behavior of TSR considers `$` a path param and defaults to undefined. So if the user is logged out and visit `/` the redirect sends them to `/sign-in/undefined` 

### What changed?

- Updated the sign in page file route path to be `sign-in.tsx` instead of `sign-in.$.tsx`

### Checklist

### DIFF
--- a/docs/references/tanstack-react-start/custom-sign-in-or-up-page.mdx
+++ b/docs/references/tanstack-react-start/custom-sign-in-or-up-page.mdx
@@ -15,11 +15,11 @@ To set up separate sign-in and sign-up pages, follow this guide, and then follow
 
   The following example demonstrates how to render the [`<SignIn />`](/docs/components/authentication/sign-in) component on a dedicated page using the [TanStack Router catch-all route](https://tanstack.com/router/latest/docs/framework/react/routing/routing-concepts#splat--catch-all-routes).
 
-  ```tsx {{ filename: 'src/routes/sign-in.$.tsx' }}
+  ```tsx {{ filename: 'src/routes/sign-in.tsx' }}
   import { SignIn } from '@clerk/tanstack-react-start'
   import { createFileRoute } from '@tanstack/react-router'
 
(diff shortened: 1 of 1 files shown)


NWylynko:
Ok cool, after switching to the server-side protection, it does redirect to `/sign-in/undefined` and switching to `sign-in.tsx` the page loads as expected.

But I am thinking this is more a bug, not just wrong docs because not having the catch-all means the auth flow can fail.

image

But this can be mitigated by using `hash` routing

```tsx
import { SignIn } from '@clerk/tanstack-react-start'
import { createFileRoute } from '@tanstack/react-router'

export const Route = createFileRoute('/sign-in')({
  component: Page,
})

function Page() {
  return <SignIn withSignUp routing="hash" />
}
```

image

wobsoriano:
Hey @lawandothman , thanks for reporting the issue and providing a PR!

I'm actually not sure if this was the behavior pre-devinxi, but I sent a message to one of the tanstaack maintainers and asked about the `undefined` value. We'll get back to you shortly.

lawandothman:
hey @wobsoriano, perfect thank you! :) I can confirm that this wasn't a thing pre-devinxi. Seems to be introduced here where the route explicitly falls back to `'undefined'` if no parameter is provided (https://github.com/TanStack/router/blob/main/packages/router-core/src/path.ts#L353)

jjrise:
@NWylynko you are the best!

wobsoriano:
https://github.com/TanStack/router/issues/4547

wobsoriano:
Fixed in https://github.com/TanStack/router/pull/4558

PR #2358, opened 2025-06-30
(WIP) - developer guide for building a custom `getUser()` function

NWylynko:
### 🔎 Previews:

- https://clerk.com/docs/pr/nick-custom-get-user/guides/create-a-custom-get-user

### What does this solve?

- We need a guide on using `auth()` successfully, staying away from `currentUser()` and webhooks. Focusing on developer experience and speed

### What changed?

- Wrote guide on creating a custom `getUser()` function that builds on top of `auth()` in a performant and successful manor.

### Checklist

### DIFF
--- a/docs/guides/create-a-custom-get-user.mdx
+++ b/docs/guides/create-a-custom-get-user.mdx
@@ -285,7 +285,7 @@ export const getUser = async () => {
   // inside the getUser function, but trust me, this is perfectly natural
 
   // One cool typescript trick, We can simply use the typescript tool `typeof`
-  // to extract out our custom type from the types.d.ts, instead of having to 
+  // to extract out our custom type from the types.d.ts, instead of having to
   // re-define the type here.
   const update = async (newDetails: Partial<typeof details>) => {
     // We are going to need to call the clerk backend api, so we will start by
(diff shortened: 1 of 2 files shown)


jescalan:
Is there any reason to run this as a guide rather than like, adding it to our SDKs? haha

docs/guides/create-a-custom-get-user.mdx
```diff
@@ -0,0 +1,757 @@
...
+```json
+{
+  "favoritePizza": "Pepperoni",
+  "favoriteSnack": "Oreos",
+  "userId": "user_2u823dCrAIzoQfgjnsimCYJvJaI",
+  "email": "clerk-test@wylynko.me"
+}
+```
+
+Our beautiful `getUser()` function gives us both our standard Clerk info (emails, phone numbers, usernames, etc) and has our custom user data.
```
royanger, line 265:
```suggestion
Our beautiful `getUser()` function gives us both our standard Clerk info (emails, usernames, etc) and has our custom user data.
```
I know what you mean, but you didn't cover/mention phone numbers so that could be confusing to a customer.

docs/guides/create-a-custom-get-user.mdx
```diff
@@ -0,0 +1,757 @@
...
+  if (user.currentOrder === null) {
+    const [newOrder] = await db
+      .insert(orders)
+      .values({
+        userId: user.userId,
+      })
+      .returning({
+        orderId: orders.orderId,
+        status: orders.status,
+      })
```
royanger, line 535: *cries in being stuck on MySQL (Planetscale) for `support-tools`*
Please Planetscale, give us access to Postgres so we can consider migrating

docs/guides/create-a-custom-get-user.mdx
```diff
@@ -0,0 +1,757 @@
...
+    quantity,
+  })
+
+  return {
+    message: 'Item added to existing order',
+  }
+}
+```
+
+### Migrating data over to metadata
```
royanger, line 571: One huge thing to callout here is that metadata can not be searched. If the customer would need to search by `orderId` or something else and that is only in metadata they can not.

royanger (commented):
Dropped a few thoughts in here. This is pretty solid as far as you've written it.

manovotny:
Closing this. Two questions raised on it never got answered, and both still stand: @jescalan asked whether this belongs in the SDKs rather than a guide, and @royanger flagged that `getUser()` collides with `clerkClient().users.getUser()` while `.update()` collides with `useUser()`. Neither has an obvious answer, and it's sat without one since July 2025.

The other thing is size. The guide builds up claims step by step — email, then metadata fields, then migrated database columns — but session claims cap out around 1.2KB after Clerk's defaults, and going over means the cookie doesn't get set at all. @royanger called that a slippery slope at the time. That caution already lives in `_partials/token-size-callout`, so the safer home for this material is a page that already includes it.

One idea in here is worth keeping though. `/docs/guides/sessions/customize-session-tokens` explains how to add custom claims but never why, and the why is the best part of this guide: developers hit rate limits calling `currentUser()` on every request, and custom claims let them read that data off the session token instead. We're adding that to the existing page, next to the size caution so the tradeoff reads as one thought.

PR #2361, opened 2025-06-30
Add example code snippets on error customization using Localization prop

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-767/customization/localization

### What does this solve?

Linear: https://linear.app/clerk/issue/DOCS-767/add-example-code-snippet-on-how-to-customize-error-messages-using

This PR adds examples of customizing error messages within Clerk components using Localization. This is in an effort to address user requests for clearer guidance around custom error messages alongside UI text.  

### What changed?

- Added new code snippets for different SDKs demonstrating how to customize error message texts via the localization object + linked list of error keys 
- Updated existing section around customizing text of Clerk components to include more examples for different SDKs.
- Added more details around each SDK's integration of the `localization` object
- Small nits: Removed duplicate `the` + added missing `defineConfig` import for astro code example

### Checklist


docs/customization/localization.mdx
```diff
@@ -294,93 +295,648 @@ Our localizations are customer-sourced and we encourage customers to add or upda
...
+
+    ```ts {{ filename: 'src/main.ts', mark: [[6, 10]] }}
+    import { createApp } from 'vue'
+    import App from './App.vue'
+    import { clerkPlugin } from '@clerk/vue'
+
+    const app = createApp(App)
+    app.use(clerkPlugin, {
+      localization: {
+        socialButtonsBlockButton: 'Sign in with {{provider|titleize}}',
```
SarahSoutoul, line 465: This needs to be changed to the new example you incorporated no? @alexisintech

alexisintech:
I've updated a few things here, with the flow of the document being the most important part. We really need to consider the order that we present information. With your proposal, you present the example first, and then present the info of "To find the key for your translation..." and then you jump back into an example "If you want to customize multiple entries"

We want to think of order of operations. Treat this almost like a tutorial. What's necessary, what's step 1? What's the next step, and the next? And can we show examples along the way of that?

Therefore, I've restructured the section to present the information in a more intuitive way.

I also don't want to emphasize the `localization` object. This is simply a best practice we're recommending, but it's not necessary. Technically, the user could be passing all of these configurations directly to the `localization` prop - we abstract it into a `localization` object to make the code more clean. Therefore, we want to emphasize the importance of the `localization` prop - not necessarily the `localization` object that we're using there. So I've removed mentions of this, and instead added code comments to provide some clarity on this pattern.

I also changed the example from using the socialButtonsBlockButton to use formButtonPrimary - because I tested those examples and didn't see any differences. also!! all of the `<SignIn />` components will _always_ have `formButtonPrimary`, so this is a better example as anyone can use it - the example before was more restrictive to users who are using social providers in their app.

Please take a look at the commit diff, and the preview, and review the differences. Let me know your thoughts before merging :)

alexisintech (approved):

SarahSoutoul:
> Awesome updates and awesome work here Sarah, thank you for this!!
> 
> I've updated a few things here, with the flow of the document being the most important part. We really need to consider the order that we present information. With your proposal, you present the example first, and then present the info of "To find the key for your translation..." and then you jump back into an example "If you want to customize multiple entries"
> 
> We want to think of order of operations. Treat this almost like a tutorial. What's necessary, what's step 1? What's the next step, and the next? And can we show examples along the way of that?
> 
> Therefore, I've restructured the section to present the information in a more intuitive way.
> 
> I also don't want to emphasize the `localization` object. This is simply a best practice we're recommending, but it's not necessary. Technically, the user could be passing all of these configurations directly to the `localization` prop - we abstract it into a `localization` object to make the code more clean. Therefore, we want to emphasize the importance of the `localization` prop - not necessarily the `localization` object that we're using there. So I've removed mentions of this, and instead added code comments to provide some clarity on this pattern.
> 
> I also changed the example from using the socialButtonsBlockButton to use formButtonPrimary - because I tested those examples and didn't see any differences. also!! all of the `<SignIn />` components will _always_ have `formButtonPrimary`, so this is a better example as anyone can use it - the example before was more restrictive to users who are using social providers in their app.
> 
> Please take a look at the commit diff, and the preview, and review the differences. Let me know your thoughts before merging :)

Nice one, thanks for the thorough review Alexis! 

On the 1st point, that was the existing order and decided to keep it as it is. But I agree with you - it's definitely better with the new flow your incorporated. 

Second, I went back and forth between the word object and prop here. Because for some SDKs, it's not really a "prop" per se. But I think it's good to keep it consistent as you did with the changes.

And finally, all good with the new example. Def better to have sth that is used more frequently.

alexisintech (approved):
feel free to merge when ready! 😸💖

PR #2363, opened 2025-06-30
sidebar reorg test

alexisintech:
https://www.notion.so/clerkdev/Workshopping-Sidebar-Alexis-PROPOSAL-2222b9ab44fe80a0a69bfdf80b1ff438?source=copy_link

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -18,27 +18,16 @@
                     "items": [
                       [
                         {
-                          "title": "Set up your Clerk account",
-                          "href": "/docs/quickstarts/setup-clerk"
-                        },
-                        {
                           "title": "Quickstart (App Router)",
                           "href": "/docs/quickstarts/nextjs"
                         },
                         {
                           "title": "Quickstart (Pages Router)",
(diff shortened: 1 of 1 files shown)


SarahSoutoul (changes requested):
Left some individual comments/ suggestions in some parts. Other than that, two things from a general standpoint:

1. Some sections don't seem to be right visually. See screenshots.

Screenshot 2025-07-01 at 11 00 10 pm
Screenshot 2025-07-01 at 10 58 32 pm

2. Made a note of the guides that aren't in the new structure atm. From my understanding, these are guides that are different depending on the SDK so it's a bit more tricky to place them. If we generalize them, then they could place under the different new headings like securing your app or development. But if they're SDK-specific, then we probably need a new heading under **Getting Started**:

- Add custom onboarding 
- Set up a waitlist
- Role Based Access Control
- Rendering modes
- Geo blocking
- Migrate from Auth.js
- tRPC
- Verifying OAuth access tokens (under Development in the Auth development section)
- Build an MCP server (under Development in the Auth development section)
- Connect an MCP client (under Development in the Auth development section)

PR #2368, opened 2025-07-02
Add context on when to sync Clerk data with webhooks

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/aa-docs-10525/webhooks/sync-data

### What does this solve?

https://linear.app/clerk/issue/DOCS-10525/add-context-on-why-we-dont-recommend-webhooksh

We don't recommend syncing Clerk's user table with an external user table using webhooks.

### What changed?

This PR adds context around why, and gives alternative approaches to using webhooks.

### Checklist


docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,27 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
 />
 
-The recommended way to sync Clerk data to your app is through webhooks.
+In some cases, you may want to sync Clerk's user table to a user table in your own database. Read the next few sections carefully to determine if this is the right approach for your app.
+
+## When to sync Clerk data
+
+You should only sync Clerk data to your database when absolutely necessary. The most notable example is if your app has social features where users can see content posted by other users.
+
+With Clerk, you can only access the currently signed in user's data from Clerk's frontend API. If you need to display a another user's name, avatar, etc., you can't access that data from Clerk's frontend API. You could use Clerk's backend API to fetch user data for each request, but it's slow and may result in rate limiting. So in this case, it's a good idea to store user data in your database, sync it to Clerk, and serve it directly.
```
jescalan, line 26: > With Clerk, you can only access the currently signed in user's data from Clerk's frontend API.

There's a logical confusion here -- I think what you intend to say is that if you're using the frontend api, you can only access the currently signed in user's data. But the way this reads is that the only way the current user's data can be accessed is with FAPI, which is not true.

jescalan: Also there is the complication here of eventual consistency as mentioned above -- if a user signs up and makes a comment on a post, then someone else goes to look at the post before the webhook sync has completed, and your code has not handled the case where the user data may not yet be in your database, but the comment would be, it could cause an error.

docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,27 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+### Alternative approaches
+
+Instead of syncing Clerk's data using webhooks, there are two other approaches you can take, depending on how much extra user data you need to store.
+
+**If it's more than \~2KB,** you could store only custom user data in your own database.
```
jescalan, line 32: I'm confused about the ~2kb constraint here...

docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,27 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+- Store the user's Clerk ID as a column in the users table in your own database, and only store extra user data. When you need to access Clerk user data, access it directly from the [Clerk session token](/docs/backend-requests/resources/session-tokens). When you need to access the extra user data, do a lookup in your database using the Clerk user ID. Consider indexing the Clerk user ID column since it will be used frequently for lookups.
```
jescalan, line 34: I think we'd want to see an example here to drive it home

jescalan (changes requested):
This is a good start but I think we'd need a bit more hand-holding here since this is a fairly complex topic, but that many Clerk users will quickly run into.

docs/webhooks/sync-data.mdx
SarahSoutoul: Yeah, I’d agree with Jeff on the first point here. I had a go at a potential rephrase to try to clarify that it’s the frontend API that’s limited to the signed-in user: 

Clerk’s frontend API only allows you to access information about the currently signed-in user. If your app needs to display information about other users. like their names or avatars, you can’t access that data from the frontend API alone. While you can fetch other users’ data using Clerk’s backend API for each request, doing so can be slow and may hit rate limits. In this case, it’s a good idea to store user data in your own database, sync it to Clerk, and serve it directly to your frontend.

docs/webhooks/sync-data.mdx
SarahSoutoul: I think we could be a bit more extensive in explaining this approach. Answering the question of **"how it is different to syncing Clerk's data with webhooks"** - from my understanding, in this scenario, you do not duplicate or sync Clerk’s standard user data (like name, email, or profile picture) into your own database, but instead, you only store the custom user data that your application needs beyond what Clerk provides. I think potentially adding a sentence reinforcing that difference could help + adding an example too.

SarahSoutoul (changes requested):
Few comments from me - looking really good overall but I think some areas need a bit more clarity and detail, especially around the alternative approaches to syncing data!

docs/_partials/metadata-callout.mdx
```diff
@@ -0,0 +1,4 @@
+> [!WARNING]
+> Metadata is limited to **8KB** maximum.
```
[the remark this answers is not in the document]
alexisintech: @jfoshee hi friend, could you give us some clarity on this so we can update the docs accordingly 🙏

docs/_partials/metadata-callout.mdx
```diff
@@ -0,0 +1,4 @@
...
+>
+> If you use Clerk metadata and modify it server-side, the changes won't appear in the session token until the next refresh. To avoid race conditions, either force a JWT refresh after metadata changes or handle the delay in your application logic.
```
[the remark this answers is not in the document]
alexisintech: yeah I agree, we can expand on it - would you want to pair with me on it, or is there someone else who could give me the context? :)

docs/webhooks/sync-data.mdx
alexisintech: > > With Clerk, you can only access the currently signed in user's data from Clerk's frontend API.
> 
> There's a logical confusion here -- I think what you intend to say is that if you're using the frontend api, you can only access the currently signed in user's data. But the way this reads is that the only way the current user's data can be accessed is with FAPI, which is not true.

yes 🤦‍♀️

docs/webhooks/sync-data.mdx
alexisintech: so everything that I wrote is taken from what you wrote: https://clerkinc.slack.com/archives/C084WHCNHCZ/p1750112692275719?thread_ts=1750109460.027299&cid=C084WHCNHCZ

docs/_partials/metadata-callout.mdx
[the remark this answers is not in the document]
jfoshee: @royanger is broadly correct. When it comes to the traditional handshake though, it's actually a little worse (explanation below)

I would recommend something like:

> User Session Metadata is practically limited to roughly 1.2KB while total user Metadata is limited to 8KB

Or however you can phrase that more succinctly/clearly... It is difficult.  We might need to link/expand to more explanation.

In a typical Handshake the contents of the session will be twice wrapped in a JWT with its base64 encoding... Along with all the other overhead of the JWT. And put in a cookie.  That cookie has a limit of 4KB. And in a quick test an "empty" handshake cookie is ~1.8KB _without_ any custom claims.

Gemini helped me come up with the 1.2KB number. [Here is a gdoc with some explanation](https://docs.google.com/document/d/1f2x8R8Zh63y2fNU7U5pwdSf4zg_dQ4WNQkTDEj4pB1s/edit?usp=sharing). Basically the encoding inflates the size by 1.8x.

All of that said, we have rolled out the new handshake flow to select customers. I've not heard of any issues thus far. The new flow should effectively eliminate cookie size problems. So a customer that encounters this limit can reach out and have the new flow enabled. 

We would still _recommend_ keeping it below 1.2KB though! Because that will perform better!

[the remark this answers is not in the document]
royanger: Thanks for the specifics around the handshake, @jfoshee!

docs/_partials/metadata-callout.mdx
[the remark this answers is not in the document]
alexisintech: what do you mean by, it would "also link to getToken({ skipCache: true }) instead."
instead of what? and where does this come into play when forcing a JWT refresh? I thought you only had to do a `user.reload`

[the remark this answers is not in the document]
royanger: You can do `getToken({ skipCache: true })` or `user.reload()`. The former skips the cookies and get and newly minted session JWT from FAPI.

docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,27 @@ description: Learn how to sync Clerk data to your app with webhooks.
   ]}
 />
 
-The recommended way to sync Clerk data to your app is through webhooks.
+In some cases, you may want to sync Clerk's user table to a user table in your own database. Read the next few sections carefully to determine if this is the right approach for your app.
+
+## When to sync Clerk data
+
+You should only sync Clerk data to your database when absolutely necessary. The most notable example is if your app has social features where users can see content posted by other users.
```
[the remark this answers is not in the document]
alexisintech: i've made some significant changes regarding this info: [85d7648](https://github.com/clerk/clerk-docs/pull/2368/commits/85d76485f57604cf251ee11b5e715e223d0e3525)

docs/webhooks/sync-data.mdx
alexisintech: updated [85d7648](https://github.com/clerk/clerk-docs/pull/2368/commits/85d76485f57604cf251ee11b5e715e223d0e3525)

alexisintech:
still waiting on responses about forcing a JWT refresh, but i've made some significant changes in the most recent commit - please let me know your thoughts, i think we're getting closer! 🤞  @jescalan @royanger

docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,27 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+With Clerk, you can only access the currently signed in user's data from Clerk's frontend API. If you need to display a another user's name, avatar, etc., you can't access that data from Clerk's frontend API. You could use Clerk's backend API to fetch user data for each request, but it's slow and may result in rate limiting. So in this case, it's a good idea to store user data in your database, sync it to Clerk, and serve it directly.
+
+### Alternative approaches
+
+Instead of syncing Clerk's data using webhooks, there are two other approaches you can take, depending on how much extra user data you need to store.
```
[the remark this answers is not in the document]
royanger: Sorry, I just didn't finish that comment. I add to the original and separated the new info with `---- edited to add ----`

docs/webhooks/sync-data.mdx
```diff
@@ -17,7 +17,36 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+The most notable use case for syncing Clerk data is if your app has social features where users can see content posted by other users. This is because Clerk's frontend API only allows you to access information about the currently signed-in user. If your app needs to display information about other users, like their names or avatars, you can't access that data from the frontend API alone. While you can fetch other users' data using Clerk's backend API for each request, the issue is that you may hit rate limits. In this case, it makes sense to store user data in your own database and sync it to Clerk.
+
+### Storing extra user data
+
+If you want to use webhooks to sync Clerk data because **you want to store extra data for the user**, consider the following approaches:
+
+1. **If it's less than 8KB,** you could use Clerk metadata.
```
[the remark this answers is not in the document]
jfoshee: If the customer is only using foundational ClerkJS and not platform SDKs, and is therefore not getting handshake behavior, then their metadata would only experience 1 base64 encoding in the JWT. (Handshake payloads get double base64 encoded).  So, the balance of space available is 0.75 ✕ (4096 - 1800 overhead) ≅ 1,722 bytes or **1.68KB**.

Not a dramatic improvement. 
(And just a reminder that this includes the property names and JSON syntax like `"onboardingComplete:" ,`)

alexisintech:
I think we are waiting on possibly adding a section or guide on how to do a JWT refresh, ideally with code examples.
@jescalan or @royanger if we could sync and get something together, I'd love to learn more about how the best practices for when a user wants to force a JWT refresh

jescalan:
This is I think the cleanest way to do it. You can also call `user.refresh()` from the `useUser` hook but that executes an extra API call. The performance difference is probably pretty close to meaningless, so if we want to recommend the `user.refresh` method instead, I don't have a problem with it. They both ultimately do the same thing.

```ts
const { session } = useSession();

useEffect(() => {
  session?.getToken({ skipCache: true });
}, []);
```

royanger:
> This is I think the cleanest way to do it. You can also call `user.refresh()` from the `useUser` hook but that executes an extra API call. 

Minor correct -- it is `user.reload()`. 

Both will do a network request, but `getToken({ skipcache: true })` will just get a token while `user.reload()` will get a new user object. The latter is useful if you want the new metadata available with `user` from `useUser()` and not just the session.

jescalan (approved):
I think this is good at this point! Nice work refining the initial version ⭐

docs/webhooks/sync-data.mdx
```diff
@@ -5,19 +5,48 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+   - For minimal custom data (under 8KB), you can store it in a user's [metadata](/docs/users/metadata) instead of dealing with a separate users table. However, if there's any chance that a user will ever have more than 8KB of extra data, you should use the other approach, as metadata is limited to 8KB.
+   - A limitation to consider is that metadata cannot be queried, so you can't use it to filter users by metadata. For example, if you stored a user's birthday in metadata, you couldn't find all users with a certain birthday. If you need to query the data that you're storing, you should use the other approach.
+
+1. **If it's more than 8KB,** you could store _only_ the extra user data in your own database.
```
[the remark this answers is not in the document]
alexisintech: [b4308c8](https://github.com/clerk/clerk-docs/pull/2368/commits/b4308c810c076a88233a8c76f42be65299d83554)

docs/webhooks/sync-data.mdx
```diff
@@ -5,19 +5,48 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+The most notable use case for syncing Clerk data is if your app has social features where users can see content posted by other users. This is because Clerk's frontend API only allows you to access information about the currently signed-in user. If your app needs to display information about other users, like their names or avatars, you can't access that data from the frontend API alone. While you can fetch other users' data using Clerk's backend API for each request, the issue is that you may hit rate limits. In this case, it makes sense to store user data in your own database and sync it to Clerk.
+
+### Storing extra user data
+
+If you want to use webhooks to sync Clerk data because **you want to store extra data for the user**, consider the following approaches:
+
+1. **If it's less than 8KB,** you could use Clerk metadata.
+   - For minimal custom data (under 8KB), you can store it in a user's [metadata](/docs/users/metadata) instead of dealing with a separate users table. However, if there's any chance that a user will ever have more than 8KB of extra data, you should use the other approach, as metadata is limited to 8KB.
```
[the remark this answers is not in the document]
alexisintech: [b4308c8](https://github.com/clerk/clerk-docs/pull/2368/commits/b4308c810c076a88233a8c76f42be65299d83554)

docs/webhooks/sync-data.mdx
```diff
@@ -5,19 +5,48 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+
+The most notable use case for syncing Clerk data is if your app has social features where users can see content posted by other users. This is because Clerk's frontend API only allows you to access information about the currently signed-in user. If your app needs to display information about other users, like their names or avatars, you can't access that data from the frontend API alone. While you can fetch other users' data using Clerk's backend API for each request, the issue is that you may hit rate limits. In this case, it makes sense to store user data in your own database and sync it to Clerk.
+
+### Storing extra user data
+
+If you want to use webhooks to sync Clerk data because **you want to store extra data for the user**, consider the following approaches:
+
+1. **If it's less than 8KB,** you could use Clerk metadata.
```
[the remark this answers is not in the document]
alexisintech: [b4308c8](https://github.com/clerk/clerk-docs/pull/2368/commits/b4308c810c076a88233a8c76f42be65299d83554)

docs/webhooks/sync-data.mdx
```diff
@@ -5,19 +5,48 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+  }
+]}
 />
 
-The recommended way to sync Clerk data to your app is through webhooks.
+In some cases, you may want to sync Clerk's user table to a user table in your own database. Read the next few sections carefully to determine if this is the right approach for your app.
+
+## When to sync Clerk data
+
+Syncing data with webhooks can be a suitable approach for some applications, but it comes with important considerations. Webhook deliveries are not guaranteed and may occasionally fail due to problems like network issues, so your implementation should be prepared to handle retries and error scenarios. Additionally, syncing data via webhooks is generally [eventually consistent](https://en.wikipedia.org/wiki/Eventual_consistency), meaning there can be a delay between when a Clerk event (such as a user being created or updated) occurs and when the corresponding data is reflected in your database. If not managed carefully, this delay can introduce bugs and race conditions.
```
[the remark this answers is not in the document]
alexisintech: [b4308c8](https://github.com/clerk/clerk-docs/pull/2368/commits/b4308c810c076a88233a8c76f42be65299d83554)

docs/backend-requests/resources/session-tokens.mdx
```diff
@@ -108,11 +110,239 @@ Claims to monitor for size limits:
...
+        const user = userId ? await clerkClient.users.getUser(userId) : null
+
+        // Return the Backend User object
+        return json({ user })
+      },
+    })
+    ```
+  </Tab>
+</Tabs>
+
```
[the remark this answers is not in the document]
alexisintech: [b4308c8](https://github.com/clerk/clerk-docs/pull/2368/commits/b4308c810c076a88233a8c76f42be65299d83554)

alexisintech:
@jescalan @royanger @SarahSoutoul I've added a guide on forcing a token refresh - would love some eyes on it ❤️

SarahSoutoul (approved):
Beautifully written guide on forcing the token to refresh! Have committed a fix to the linting issues + minor sentence phrasing :pray:

docs/backend-requests/custom-session-token.mdx
```diff
@@ -288,15 +275,14 @@ This guide will show you how to customize a session token to include additional
...
   export {}
 
   declare global {
     interface CustomJwtSessionClaims {
-      fullName?: string
-      primaryEmail?: string
+      metadata: string
     }
   }
   ```
```
jescalan, line 288: Why make this change? Feels like it's suggesting a clearly riskier option without a clear benefit

docs/backend-requests/resources/session-tokens.mdx
```diff
@@ -90,15 +90,17 @@ Read more about Clerk session tokens and how they work in [the guide on how Cler
...
+## Custom claims
+
 If you would like to add custom claims to your session token, you can [customize it](/docs/backend-requests/custom-session-token).
 
 You can also create custom tokens using a [JWT template](/docs/backend-requests/jwt-templates).
 
 ## Size limitations
 
-The Clerk session token is stored in a cookie. All modern browsers [limit the maximum size of a cookie to 4kb](https://datatracker.ietf.org/doc/html/rfc2109#section-6.3). Exceeding this limit can have adverse effects, including a possible infinite redirect loop for users who exceed this size in Next.js applications.
+The Clerk session token is stored in a cookie. All modern browsers [limit the maximum size of a cookie to **4KB**](https://datatracker.ietf.org/doc/html/rfc2109#section-6.3). Exceeding this limit can have adverse effects, such as possible infinite redirect loops in Next.js applications.
```
jescalan, line 101: Exceeding this limit will cause the cookie not to be set, which will guaranteed break your app since Clerk relies on cookies to function.

docs/backend-requests/resources/session-tokens.mdx
```diff
@@ -90,15 +90,17 @@ Read more about Clerk session tokens and how they work in [the guide on how Cler
...
 
-A session token with the [default session claims](#default-claims) won't run into this issue, as this configuration produces a cookie significantly smaller than 4kb. However, this limitation becomes relevant when implementing a [custom session token](/docs/backend-requests/custom-session-token). In this case, it's recommended to move particularly large claims out of the token and fetch these using a separate API call from your backend.
+A session token with the [default session claims](#default-claims) won't run into this issue, as this configuration produces a cookie significantly smaller than 4KB. However, this limitation becomes relevant when implementing a [custom session token](/docs/backend-requests/custom-session-token). **It's recommended to keep the total size of custom claims in the session token under 1.2KB.**
```
jescalan, line 103: Gotta explain why. I'd read this and be like, this makes no sense, you just said the limit is 4kb, so why keep it under 1.2kb? The answer generally is "clerk adds metadata in addition to the payload that reduces the available space".

Also, Foshee might have already shipped a change that makes this not the case and removes the limit, since it will just set a second cookie if we're over the limit. You should check with him and see. If this is the case, we can change the copy to be more along the lines of "you should keep your custom claims size as low as possible because the more you add, the slower your app becomes since cookies are shuttled back and forth with every request between FE and BE. We recommend 1.2kb max to keep your performance optimal"

docs/backend-requests/resources/session-tokens.mdx
```diff
@@ -108,11 +110,241 @@ Claims to monitor for size limits:
...
+
+        // Return the Backend User object
+        return json({ user })
+      },
+    })
+    ```
+  </Tab>
+</Tabs>
+
+However, if you make this call to Clerk's Backend API frequently, you risk hitting [rate limits](/docs/backend-requests/resources/rate-limits) and it's also slower than making a database query. So it's recommended to [store the extra data in your own database](/docs/webhooks/sync-data#storing-extra-user-data) instead of storing it in metadata in the session token.
```
jescalan, line 346: It seems weird to recommend making a BAPI call, put a bunch of examples of how to do it, then immediately after being like "actually, we recommend not doing this". I feel like I'd flip these around -- explicitly recommend storing larger user data in your own database, then mention that if there's some restriction that makes this impossible, it is possible to fetch via BAPI, but risky because of rate limits, so we do not recommend this.

docs/guides/force-token-refresh.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+When using metadata, it's recommended to [store it in the user's session token](/docs/backend-requests/custom-session-token) to avoid making a network request to Clerk's Backend API when retrieving it.
+
+However, when retrieving the user's metadata from the session token, because the session token refreshes every 60 seconds, the metadata may not be up to date if the token hasn't refreshed yet. In this case, you'd want to force a session token refresh. This is when you could use `getToken({ skipCache: true })`.
+
+So why does this example use `user.reload()` instead? Because on this user's profile page, the metadata is not the only information the user can update. They can also update their first and last name. Therefore, you not only want to refresh the session token, which is where the metadata is retrieved from, but also the `User` object, which is where the first and last name is retrieved from. This is why `user.reload()` is the better option. If you were only retrieving the user's metadata, and not using the `User` object, you could use `getToken({ skipCache: true })` instead.
```
jescalan, line 36: We haven't seen any example yet, so this reads very strangely

docs/guides/force-token-refresh.mdx
```diff
@@ -0,0 +1,244 @@
...
+        return Response.json({ error: 'Unauthorized' }, { status: 401 })
+      }
+
+      // Use `clerkClient()` to access the Clerk client
+      const client = await clerkClient()
+
+      // Use the Backend SDK's `updateUserMetadata()` method to update the user's metadata
+      await client.users.updateUserMetadata(userId, {
+        publicMetadata: {
+          ...metadata,
```
jescalan, line 66: I know this is just an example, but this would be really bad code to write. Any user could just dump the full works of shakespeare into this api route and probably break your app. It would probably be a better, safer example if you updated some specific piece of metadata here rather than just taking user input and piping it straight in

docs/users/metadata.mdx
```diff
@@ -289,7 +288,7 @@ There are multiple ways to retrieve public metadata.
...
+On the backend, it's available on the [Backend `User`](/docs/references/backend/types/backend-user) object which can be accessed using the JavaScript Backend SDK's [`getUser()`](/docs/references/backend/user/get-user) method. However, this makes a network request, which is subject to [rate limits](/docs/backend-requests/resources/rate-limits). To avoid a network request, you can attach the metadata as a claim in the user's session token. Then, the metadata can be retrieved from the `sessionClaims` on the [`Auth`](/docs/references/backend/types/auth-object) object. See the [guide on customizing your session token](/docs/backend-requests/custom-session-token).
```
jescalan, line 291:
```suggestion
On the backend, it's available on the [Backend `User`](/docs/references/backend/types/backend-user) object which can be accessed using the JavaScript Backend SDK's [`getUser()`](/docs/references/backend/user/get-user) method. However, this makes an API request, which is subject to [rate limits](/docs/backend-requests/resources/rate-limits). To avoid a network request, you can attach the metadata as a claim in the user's session token. Then, the metadata can be retrieved from the `sessionClaims` on the [`Auth`](/docs/references/backend/types/auth-object) object. See the [guide on customizing your session token](/docs/backend-requests/custom-session-token).
```

jescalan (commented):
This is a huge improvement for sure

docs/backend-requests/resources/session-tokens.mdx
jfoshee: I still recommend keeping the 1.2KB advice as that limit should keep users on the 1-step handshake flow.

Even with the 2-step handshake support we still have a limit of ~ 2.5KB for custom session data in session cookies. (see [this comment above](https://github.com/clerk/clerk-docs/pull/2368#discussion_r2208288450)).

This is due simply to the fact that the JWT session is confined to a single cookie. This ~2.5KB will be a "hard" limit until & if we implement chunking of session cookies.

alexisintech: it's included in the linked /custom-session-token guide at the top, but I can add context here as well!

docs/webhooks/sync-data.mdx
```diff
@@ -5,19 +5,48 @@ description: Learn how to sync Clerk data to your app with webhooks.
...
+1. **If it's more than 1.2KB,** you could store _only_ the extra user data in your own database.
+   - Store the user's Clerk ID as a column in the users table in your own database, and only store extra user data. When you need to access Clerk user data, access it directly from the [Clerk session token](/docs/backend-requests/resources/session-tokens). When you need to access the extra user data, do a lookup in your database using the Clerk user ID. Consider indexing the Clerk user ID column since it will be used frequently for lookups.
+   - For example, Clerk doesn't collect a user's birthday, country, or bio, but if you wanted to collect these fields, you could store them in your own database like this:
+     | id | clerk\_id | birthday | country | bio |
+     | - | - | - | - | - |
+     | user123abc | user\_123 | 1990-05-12 | USA | Coffee enthusiast. |
+     | user456abc | user\_456 | 1985-11-23 | Canada | Loves to read. |
+     | user789abc | user\_789 | 2001-07-04 | Germany | Student and coder. |
+
+1. A hybrid approach of the two approaches above.
```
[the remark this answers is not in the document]
alexisintech: very fair point and i've created a ticket for it https://linear.app/clerk/issue/DOCS-10682/expand-on-how-a-user-can-extend-clerk-data

but i want to go ahead and ship this ahead of the current IA project (trying to close out PR's before that gets shipped). so this will have to get revisited at a later time

PR #2394, opened 2025-07-09
Add example for using handleRedirectCallback()

SarahSoutoul:
### 🔎 Previews:

-

### What does this solve?

The `handleRedirectCallback()` method doesn't include any examples like the rest of the methods in the docs. It links to the [custom flow page](https://clerk.com/docs/custom-flows/oauth-connections), which doesn't include an example either. This PR adds an example following the structure used for other methods within this page: https://clerk.com/docs/references/javascript/clerk. 

### What changed?

- Added an example of using the `handleRedirectCallback()` method - I kept it consistent with the rest of the examples on that page and removed the custom flow mention as that page didn't seem to have any mention of that method. 

### Checklist

### DIFF
--- a/docs/references/javascript/clerk.mdx
+++ b/docs/references/javascript/clerk.mdx
@@ -739,7 +739,12 @@ function handleRedirectCallback(
 
 #### Example
 
-See the [custom flow](/docs/custom-flows/oauth-connections) guide for a comprehensive example of how to use the `handleRedirectCallback()` method.
+```js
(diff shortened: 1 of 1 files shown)


SarahSoutoul:
> while I was checking on this function, I noticed some things could be updated to get it up to date with the javascript repo! hopefully soon with typedoc, we'll get these docs automated
> 
> this looks good 😸💖

Thanks for changes - looking good!

PR #2395, opened 2025-07-09
Use useRouter() for navigation instead of expo-linking

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-10609/quickstarts/expo

### What does this solve?

Linear: https://linear.app/clerk/issue/DOCS-10609/expo-use-userouter-instead-of-expo-linking

The current code snippet for setting up an Expo sign out button uses `expo-linking`, which is not best practice for internal navigation. This PR replaces the use of `expo-linking` with `useRouter` in the code snippet to ensure the use of best practices.

### What changed?

- Replaced `expo-linking` with `useRouter` in the code and used the `replace` method from the router to achieve same outcome
- Tested the new code with an Expo Clerk app

### Checklist

### DIFF
--- a/docs/references/javascript/session.mdx
+++ b/docs/references/javascript/session.mdx
@@ -193,7 +193,7 @@ await clerk.session.attemptSecondFactorVerification({
 
 ### `checkAuthorization()`
 
-Checks if the user is authorized for the specified role or permission.
+Checks if the user is [authorized for the specified role or permission](/docs/guides/authorization-checks) or requires the user to [reverify their credentials](/docs/guides/reverification) if their last verification is older than allowed.
 
 ```ts
 function checkAuthorization(param: CheckAuthorizationParams): boolean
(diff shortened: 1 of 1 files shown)


SarahSoutoul:
> i think there's another example in the docs that uses expo-linking , if you could tackle that one as well - we want to wipe _all_ instances of expo-linking in favor of userouter()

@alexisintech - Discussed this with @panteliselef, and as I suspected, the remaining code snippet using `expo-linking` does need it, as in this scenario, it's being used for creating deeplinks between applications rather than just navigating to another internal route within the same app. So the `useOauth()` example makes a valid use of `expo-linking`, and needs to remain as is.

alexisintech:
> > i think there's another example in the docs that uses expo-linking , if you could tackle that one as well - we want to wipe _all_ instances of expo-linking in favor of userouter()
> 
> @alexisintech - Discussed this with @panteliselef, and as I suspected, the remaining code snippet using `expo-linking` does need it, as in this scenario, it's being used for creating deeplinks between applications rather than just navigating to another internal route within the same app. So the `useOauth()` example makes a valid use of `expo-linking`, and needs to remain as is.

ahhh good point!! thanks for that clarification 🙏  TIL :)

alexisintech (approved):
😸💖

SarahSoutoul:
@alexisintech PR on the expo quickstart here: https://github.com/clerk/clerk-expo-quickstart/pull/13

PR #2418, opened 2025-07-14
Add ability to define distinct docs variants per sdk

NWylynko:
### 🔎 Previews:

-

### What does this solve?

- Sometimes when creating an sdk scoped guide, it would be desirable to be able to write a completely different guide for specific sdk.

### What changed?

- This pr adds to the build script to pick up `*.react.mdx`, .`*.nextjs.mdx`, etc extensions and use that variant for the guide instead.

### Checklist

- [x] Validate the dev caching works as expected / add a test
- [x] Update CONTRIBUTING.md


alexisintech:
can you add more details to the description? what is the purpose behind this? we don't currently have `.react.mdx` or `.nextjs.mdx` files so what is your vision here, how would things get authored and then rendered, etc?

NWylynko:
> can you add more details to the description? what is the purpose behind this? we don't currently have `.react.mdx` or `.nextjs.mdx` files so what is your vision here, how would things get authored and then rendered, etc?

@alexisintech 

I've added to the contributing guide

```
If you'd like to add support for a new SDK in a guide, but don't want to use the `<If />` component to swap in sdk specific documentation. Another option is to use the `.sdk.mdx` file extension. For example say you had a `docs/setup-clerk.mdx` with `sdk: react, nextjs, expo` in the frontmatter. You could create a `docs/setup-clerk.remix.mdx` file. Then write out a remix specific version of the guide. This will act the same way as above, treating the `remix` sdk as a distinct variant of the guide. Creating a `/docs/remix/setup-clerk` page along side the `/docs/nextjs/setup-clerk` and `/docs/expo/setup-clerk` pages.
```

Essentially this is no different to putting an sdk option in the frontmatter, but it allows you to have a whole new separate variant of the guide that shares no content with the "generic" sdk scoped page.

alexisintech:
> @alexisintech
> 
> > i'm a little confused - this allows you to write `setup-clerk.react.mdx`, but we can already do that by creating `react/setup-clerk.mdx` and then it lives in the `react/` folder and doesn't bloat our main folders like `docs/` and `guides/`
> > so i still don't think i'm seeing the vision here 🤔
> 
> The plan was we wanted to do a `/mcp/build-mcp-server.mdx` (nextjs scoped) and `/mcp/build-mcp-server.expressjs.mdx` guide. Then this builds them both in to `/docs/nextjs/mcp/build-mcp-server` and `/docs/expressjs/mcp/build-mcp-server`. Creating a base `/docs/mcp/build-mcp-server`, redirecting the user to the appropriate guide based on their active sdk, or asking them to pick a version.

I think I need to be demo'd what you mean, because I'm not sure why we can't just create the file in the `express/` folder as we can with the current functionality

alexisintech (approved):
i haven't tested or done a thorough review, but approving as to not block

PR #2443, opened 2025-07-23
[IA] docs team proposal

kylemac:
Master branch for the Docs IA restructure

🔎 Preview: 
- https://clerk.com/docs/pr/ia-docs-team-proposal

⚠️  Remember to merge https://github.com/clerk/javascript/pull/6846 after this PR is merged

⚒️  Linear project: https://linear.app/clerk/project/docs-ia-1f68625ed62b/issues

PR #2453, opened 2025-07-28
IA: Configuring your app (/configure) section

alexisintech:
### 🔎 Previews:

-

Handles the **Configuring your app** (`/configure`) section of the new sidebar
Note: Deletes /authentication/overview in favor of /sign-up-sign-in-options

TODO:
- [x] webhooks/inngest.mdx - we had convos about possibly deleting this guide. moved for now to /configure/webhooks. we can decide if we want to delete or what later
- [x] webhooks/loops.mdx - we had convos about possibly deleting this guide. moved for now to /configure/webhooks. we can decide if we want to delete or what later
- [x] typedoc errors (we should probably just ignore for now, and once the main PR is ready, we can submit a corresponding typedoc PR)
- [x] can't preview if the pages are working because sidebar currently just uses a generated page for each route
- [x] Add support for headings to fix the third-level dropdown issues that break the sidebar structure
- [x] Fix issue with anything in parentheses that should appear on the sidebar 

**Review checklist:**

- [x] All the pages are rendering correctly
- [x] All the routes that needed to get handled according to the `proposal.md` are handled + replaced in links all over docs
- [x] All the paths are handled (for e.g, in the `/configure` PR, check the `proposal-mapping.json` and check that all newPaths with `/configure` in them are accounted for in the PR) - **4 sections missing under Social Connections - see comment**
- [x] For anything with a `consolidate` action, make sure that information gets consolidated and not fully deleted 
- [x] Check redirects 
- [x] Check sidebar structure isn't broke. Fixed here: https://github.com/clerk/clerk-docs/pull/2443/commits/500b336bb60a5b8cacd127ee834e8c1f0cf846fc

### DIFF
--- a/proposal.md
+++ b/proposal.md
@@ -36,34 +36,38 @@
 ### Configuring your App [configure]
 
 - Authentication Strategies [auth-strategies]
+
   - Sign-up and sign-in options [sign-up-sign-in-options]
   - Social Connections
-    - Apple
-    - Atlassian
-    - Bitbucket
-    - Box
-    - Coinbase
-    - Discord
-    - Dropbox
-    - Facebook
-    - GitHub
-    - GitLab
-    - Google
-    - HubSpot
-    - Hugging Face
-    - Line
-    - Linear
-    - LinkedIn [linkedin-oidc]
-    - LinkedIn (deprecated) [linkedin]
-    - Microsoft
-    - Notion
-    - Slack
-    - Spotify
-    - TikTok
-    - Twitch
-    - Twitter v1 (deprecated) [twitter]
-    - X/Twitter v2 [x-twitter]
-    - Xero
+    - Account linking
+    - Custom provider
+    - All providers
+      - Apple
+      - Atlassian
+      - Bitbucket
+      - Box
+      - Coinbase
+      - Discord
+      - Dropbox
+      - Facebook
+      - GitHub
+      - GitLab
+      - Google
+      - HubSpot
+      - Hugging Face
+      - Line
+      - Linear
+      - LinkedIn [linkedin-oidc]
+      - LinkedIn (deprecated) [linkedin]
+      - Microsoft
+      - Notion
+      - Slack
+      - Spotify
+      - TikTok
+      - Twitch
+      - Twitter v1 (deprecated) [twitter]
+      - X/Twitter v2 [x-twitter]
+      - Xero
   - Enterprise Connections
     - Overview
     - Authentication flows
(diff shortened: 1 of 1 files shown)


proposal.md
```diff
@@ -53,43 +53,43 @@
...
     - Notion
     - Slack
     - Spotify
     - TikTok
     - Twitch
-    - Twitter v1 (deprecated)
-    - X/Twitter v2
+    - Twitter v1 (deprecated) [twitter]
+    - X/Twitter v2 [x-twitter]
     - Xero
```
SarahSoutoul, line 66: This is missing the following pages:

- **Overview**
- **Account linking**
- **All providers**
- **Custom provider**

Screenshot 2025-08-08 at 2 07 15 pm

From the `proposal-mapping.json`, there is nothing that says these pages should be deleted. So assuming they should be under **Social Connections** as the mapping suggests. But wanted to double check this with you before making the changes @alexisintech.

alexisintech: the overview got moved, and that is in handled in `redirects/static/docs.json` as well
Screenshot 2025-08-09 at 17 07 05

added account linking

for all providers vs custom provider (good catch, forgot custom provider!!), we can either do a dropdown for all providers or a subheading, but either way, third-level dropdowns/subheadings aren't looking good:
Screenshot 2025-08-09 at 17 11 26
Screenshot 2025-08-09 at 17 12 50

i have a ticket for it here we can address at  a later time: https://linear.app/clerk/issue/DOCS-10785/support-third-level-dropdownssubheadings

SarahSoutoul: Finally, for the first point about the **Overview** page, I'm referring to this overview page under **Social Connections**: https://clerk.com/docs/authentication/social-connections/overview rather than this one: https://clerk.com/docs/authentication/overview. 

Some of the content of that overview has been moved to the `sign-up-sign-in-options` page as you mentioned, but a lot of the content isn't included. You might have decided to delete it all together, but wanna make sure that's the case, so we're not deleting important info. 

Also, the redirect you specified handles this page: https://clerk.com/docs/authentication/overview. But it doesn't handle this one then: https://clerk.com/docs/authentication/social-connections/overview, if the content is moved to `sign-up-sign-in-options`. This is the current redirect for it, which should be changed then. 

Screenshot 2025-08-11 at 2 07 36 pm

SarahSoutoul: Okay cool. I know https://clerk.com/docs/authentication/overview isn't needed and is now within `sign-up-sign-in-options` - that's all good. 

The issue is the Overview under Social Connections that need to be handled. I got confused cause I was referring to that overview page but in this comment, it's referring to a completely different page:

Screenshot 2025-08-11 at 5 38 35 pm

Anyhow, let's tackle that needed one then!

PR #2459, opened 2025-07-30
put ios components dropdown within guides

mikepitre:
### What changed?

- Moves the iOS Components within the iOS Guides in the manifest.
- Visually this shouldn't change anything, but it makes the location of this drop down consistent with others in the manifest/sidebar.

### Checklist


alexisintech (approved):
yeah this is what i was thinkin!!

PR #2460, opened 2025-07-31
Update useUser docs to indicate that reload() is only needed when updating user without using Clerk APIs

nikosdouvlis:
### What does this solve?

Currently the docs don't explicitly state when `reload()` is needed 

- 

### What changed?

Explicitly explain that this is only needed when updating `user` without using the Clerk hooks


alexisintech (approved):

alexisintech:
removed accidentally added page here: https://github.com/clerk/clerk-docs/commit/2f4ce790f4fe51af034fc9b815921568f8b82d6e

PR #2464, opened 2025-07-31
IA: Hooks and components

alexisintech:
### 🔎 Previews:

-

Moves /docs/components/** --> /docs/reference/components/**
Moves /docs/hooks/** -->  /docs/reference/hooks/**

TODO: 
- [x] support headings in sidebar: https://linear.app/clerk/issue/DOCS-10774/support-headings-in-the-sidebar 

TODO:
- [x] can't preview

**Review checklist:**

- [x] All the pages are rendering correctly
- [x] All the routes that needed to get handled according to the `proposal.md` are handled
- [ ] All the paths are handled (for e.g, in the `/configure` PR, check the `proposal-mapping.json` and check that all newPaths with `/configure` in them are accounted for in the PR) 
- [x] For anything with a `consolidate` action, make sure that information gets consolidated and not fully deleted 
- [x] Check redirects 
- [x] Check sidebar structure isn't broken


alexisintech:
we need a way to support headings, because the headings like "authentication components" are turned into dropdowns

proposal.md
```diff
@@ -247,10 +247,63 @@
...
+    - `<RedirectToOrganizationProfile />` [redirect-to-organization-profile]
+    - `<RedirectToCreateOrganization />` [redirect-to-create-organization]
+    - `<SignedIn>` [signed-in]
+    - `<SignedOut>` [signed-out]
+  - Unstyled Components [unstyled] {"collapse": false}
+    - `<SignInButton>` [sign-in-button]
+    - `<SignInWithMetamaskButton>` [sign-in-with-metamask]
+    - `<SignUpButton>` [sign-up-button]
+    - `<SignOutButton>` [sign-out-button]
 - Hooks
```
SarahSoutoul, line 290: @alexisintech Do we want to add an icon like we did for UI components? Otherwise, it looks a bit strange UI wise.

docs/reference/general/components/overview.mdx
```diff
@@ -0,0 +1,58 @@
...
+- [`<SignedIn />`](/docs/reference/general/components/control/signed-in)
+- [`<SignedOut />`](/docs/reference/general/components/control/signed-out)
+
+## Unstyled components
+
+- [`<SignInButton />`](/docs/reference/general/components/unstyled/sign-in-button)
+- [`<SignInWithMetamask />`](/docs/reference/general/components/unstyled/sign-in-with-metamask)
+- [`<SignUpButton />`](/docs/reference/general/components/unstyled/sign-up-button)
+- [`<SignOutButton />`](/docs/reference/general/components/unstyled/sign-out-button)
+
```
SarahSoutoul, line 43: @alexisintech This is missing the billing components - do we want to be adding them?

SarahSoutoul:
> @alexisintech @NWylynko Posted it as a comment on the Slack channel but noticed two behaviours on this:
> 
> 1. When clicking on each component/ hook, there is a sort of flicker with the page, which is caused by the URL switching for a split second. So the page is blank for a split second and then renders the content. This might be normal for now as we fix certain things, but just wanted to flag.
> 
>  flickering.mov 
> 3. The other thing is when switching from one SDK to another SDK via the dropdown, the URL doesn't change as it should. Again, might already be sth that we knew but flagging in case.
> 
>  url-not-changing.mov

This will be tackled later on, as explained by Nick in this [Slack thread](https://clerkinc.slack.com/archives/C084WHCNHCZ/p1755008008676099?thread_ts=1754925768.971919&cid=C084WHCNHCZ).

PR #2515, opened 2025-08-18
Add note on Apple Sign In email issues depending on user base locale

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-10642/authentication/social-connections/apple

### What does this solve?

Developers using Sign in with Apple may encounter issues for users in regions such as China and India, where Apple IDs can be tied only to a phone number with no email address. If Clerk instances enforce the requirement for an email during sign-up, these users will be unable to sign in with Apple. 

This update clarifies the limitation and provides guidance on how to configure instance settings depending on the developer’s user base.

### What changed?

Added a note to the Sign in with Apple docs warning developers that some Apple IDs may not include an email address + added note on how to disable this requirement in the Clerk Dashboard when supporting user bases from affected regions.

### Checklist

### DIFF
--- a/docs/authentication/social-connections/apple.mdx
+++ b/docs/authentication/social-connections/apple.mdx
@@ -99,7 +99,7 @@ To make the setup process easier, it's recommended to keep two browser tabs open
   ### Configure Email Source for Apple Private Relay
 
   > [!NOTE]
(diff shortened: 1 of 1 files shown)


alexisintech (approved):
thank you! 😸💖

PR #2538, opened 2025-08-25
IA: category rework proposal

alexisintech:
https://www.loom.com/share/b0bb8c34c9984914aa6c8592277079fc?sid=c2070b33-39ff-4d94-a649-ced3f3cb7d7b

🔎 Preview:
- https://clerk.com/docs/pr/aa-category-proposal-1


docs/manifest.json
```diff
@@ -450,6 +366,39 @@
               ]
             }
           ],
+          [
+            {
+              "title": "Session Management",
```
SarahSoutoul, line 371: Love this new section! I was wondering if the "Session tokens" page we have under "How Clerk works" would fit better here, this one: **docs/guides/how-clerk-works/session-tokens**. 

That way, before a user wants to customize a session token, they can learn what they are first. But again, the session token customization page does link to that general session tokens page, so no sure.

docs/manifest.json
```diff
@@ -310,112 +310,28 @@
...
-                            ]
-                          ],
-                          "collapse": true
-                        }
-                      ]
-                    ],
-                    "collapse": true
+                    "title": "Add custom onboarding",
+                    "href": "/docs/guides/development/add-onboarding-flow"
+                  },
```
SarahSoutoul, line 323: Would it make sense to nest these three pages under a dropdown named "Custom Authentication flows" or "Custom User flows"? It would make for a cleaner sidebar for Next.js imo, and logic to group them together.

SarahSoutoul (approved):

PR #2541, opened 2025-08-25
Add warning that Google Analytics integration isn't available for new applications

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-10843/integrations/analytics/google-analytics

### What does this solve?

Linear: https://linear.app/clerk/issue/DOCS-10843/google-analytics-integration-docs-should-notewarn-users-that-the

The GA integration is no longer supported. Users won’t be able to enable GA on new applications, but for apps that already have GA enabled, it can still be configured. If turned off, it cannot be re-enabled. This PR aims at adding a note to warn users about this change. 

### What changed?

- Added a warning about this change, specifying the difference btw new and existing apps 

**Important note:** The Linear ticket mentions adding a guide to show users how they can implement their own analytics tracking as a workaround. This purely involves showing an example snippet of code, showing how to use the `useUser` hook in conjunction with `useEffect` to fire the analytics events based on a user's signing state. IMO, this doesn't guarantee a guide as there wouldn't be much in there + the code would differ depending on the analytics provider. For this reason, I added this more as a general suggestion in the warning note, but open to suggestions. @alexisintech @kduprey 

### Checklist


alexisintech (approved):
yeah this is great sarah, thank you!

SarahSoutoul:
> > > Its good, maybe worth an example of the hook thing? It would be react-specific tho. It's basically running a `useEffect` with its dependency being the return value of `useUser` - when the value is null, the user has been signed out, and when its not null they are signed in
> > 
> > 
> > Yeah I don't mind adding that in. I just don't really know where it could fit in the page itself. An idea I had was have a question potentially that says: `I have a new application. Is there a workaround to track authentication events with Google Analytics?`. And then the answer shows the workaround.
> > Thoughts @jescalan @alexisintech ?
> 
> we can try to put the code example in the callout and see what that looks like

Screenshot 2025-08-26 at 5 20 08 pm

Meh? :melting_face: @alexisintech

SarahSoutoul:
@alexisintech @kduprey - Okay, have pushed the proposal if you wanted to have a look for a last check: [Proposal](https://github.com/clerk/clerk-docs/pull/2541/commits/908c70b9399df5aeca2e8eae837204b9f4ce5c62)

alexisintech:
@SarahSoutoul yep lets ship 🚢  awesome work here, and great idea sarah love it 😸💖

PR #2552, opened 2025-08-28
chore(tasks): Refer to `setActive({ navigate })` for checking pending tasks on custom flows

LauraBeatris:
### 🔎 Previews:

- https://clerk.com/docs/pr/laura-patch-custom-flow-task-instructions/custom-flows/overview

### What does this solve?

Update custom flows mentions that after sign-in or sign-up, `setActive({ navigate })` should be used to check against pending session tasks.

### Checklist

### DIFF
--- a/docs/custom-flows/overview.mdx
+++ b/docs/custom-flows/overview.mdx
@@ -28,7 +28,7 @@ The following steps outline the sign-up process:
 1. Prepare the verification.
 1. Attempt to complete the verification.
 1. If the verification is successful, set the newly created session as the active session by passing the `SignIn.createdSessionId` to the [`setActive()`](/docs/references/javascript/clerk#set-active) method on the `Clerk` object.
-1. Check for any pending tasks that need to be completed by accessing [`Session.currentTask`](/docs/references/javascript/types/session-task). Refer to [session tasks](/docs/authentication/configuration/session-tasks)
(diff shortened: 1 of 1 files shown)


docs/custom-flows/overview.mdx
```diff
@@ -28,7 +28,7 @@ The following steps outline the sign-up process:
 1. Prepare the verification.
 1. Attempt to complete the verification.
 1. If the verification is successful, set the newly created session as the active session by passing the `SignIn.createdSessionId` to the [`setActive()`](/docs/references/javascript/clerk#set-active) method on the `Clerk` object.
-1. Check for any pending tasks that need to be completed by accessing [`Session.currentTask`](/docs/references/javascript/types/session-task). Refer to [session tasks](/docs/authentication/configuration/session-tasks)
+1. Use the `navigate` parameter in [`setActive()`](/docs/references/javascript/clerk#set-active) to access `session.currentTask` and check for pending session tasks.
```
SarahSoutoul, line 31: I think it might be better to link `navigate` to here: https://clerk.com/docs/pr/laura-patch-custom-flow-task-instructions/references/javascript/types/set-active-params. Because when I clicked `setActive()`, I was trying to look for the navigate parameter on the page and couldn't find it. As a user, I think it's better to be led straight to the parameter. Thoughts @LauraBeatris ?

PR #2613, opened 2025-09-19
fix: Update `@latest` to `@5`

tmilewski:
### What does this solve?

It was reported in the JS repo that `@latest` was resolving to v4. This, however, is correct.

It appears that one of the two instances on this page didn't get updated to `@5` from `@latest`.

### Checklist


SarahSoutoul (approved):
LGTM! Will update the dashboard onboarding appropriately!

PR #2625, opened 2025-09-25
Add section on handling missing requirements during custom SSO sign-up

beerose:
### 🔎 Previews:

-

### What does this solve?

Linear ticket: https://linear.app/clerk/issue/SUP-496/custom-flow-guide-how-to-handle-missing-requirements

### What changed?

- Added a new "Handle missing requirements during sign-up" section to the custom OAuth flow guide. An alternative would be to scope it under "Next.js" tab in "Create the sign-up and sign-in flow", but I assume that missing requirements are quite generic, and people using other frameworks could use it as a base, too. Also, I think, eventually, we'll add other tabs as well ("Expo", "iOS", "Android"). 
- Included a Next.js example of a continue page and updated SSO callback with `continueSignUpUrl`.
- Added notes on why this only applies to SignUp, and comments in code on handling different field types.

### Checklist

### DIFF
--- a/docs/guides/development/custom-flows/authentication/oauth-connections.mdx
+++ b/docs/guides/development/custom-flows/authentication/oauth-connections.mdx
@@ -39,7 +39,7 @@ You must configure your application instance through the Clerk Dashboard for the
             .authenticateWithRedirect({
               strategy,
               redirectUrl: '/sign-in/sso-callback',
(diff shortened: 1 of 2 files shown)


alexisintech:
tracking a bug here: https://clerkinc.slack.com/archives/C07JK3YT2G2/p1759355153141669

PR #2668, opened 2025-10-03
Document missing createInvitationBulk method on JS Backend SDK 

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-10973/reference/backend/invitations/create-invitation-bulk

### What does this solve?

In the Clerk Backend API Reference, there is a route to create multiple invitations in bulk, and the wrapper method (`createInvitationBulk`) for that route is also being documented in the `javascript` repo [here](https://github.com/clerk/javascript/blob/main/packages/backend/src/api/endpoints/InvitationApi.ts). However, we do not have it in our docs - this PR documents the missing method within our JS Backend SDK. 

### What changed?

- Adds the `createInvitationBulk` method in JS Backend SDK. 

**Note:** While doing this work, @manovotny and I noticed the `templateSlug` and `expiresInDays` properties were missing from the `createInvitation` method while being documents in both the `javascript` repo and the Backend API reference docs. So we added them there too as part of this PR. 

### Checklist

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -2445,6 +2445,10 @@
                                 "href": "/docs/reference/backend/invitations/create-invitation"
                               },
                               {
+                                "title": "`createInvitationBulk()`",
+                                "href": "/docs/reference/backend/invitations/create-invitation-bulk"
+                              },
+                              {
                                 "title": "`revokeInvitation()`",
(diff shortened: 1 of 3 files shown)


manovotny:
Let's wait for @alexisintech to review since I paired on this with you. 😊

PR #2703, opened 2025-10-14
Swap dashboard links: last-active?path= for /~

alexisintech:
update dashboard links to use `/~/<PATH>` instead of `last-active?path=<PATH>`

### DIFF
--- a/docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
+++ b/docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
@@ -41,7 +41,7 @@ Ensure you have setup a new connection within the third-party client or tool whe
 
 To create a Clerk OAuth application:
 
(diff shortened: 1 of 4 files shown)


docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
```diff
@@ -41,7 +41,7 @@ Ensure you have setup a new connection within the third-party client or tool whe
 
 To create a Clerk OAuth application:
 
-1. In the Clerk Dashboard, navigate to the [**OAuth Applications**](https://dashboard.clerk.com/last-active?path=user-authentication/oauth-applications) page.
+1. In the Clerk Dashboard, navigate to the [**OAuth Applications**](https://dashboard.clerk.com/~/user-authentication/oauth-applications) page.
```
SarahSoutoul: @alexisintech That link works fine as there is redirection anyway, but I can see this one: https://dashboard.clerk.com/~/oauth-applications, being used in other pages. But it doesn't live under user-authentication anymore it seems, so we may want to make them all have this format: https://dashboard.clerk.com/~/oauth-applications?

PR #2713, opened 2025-10-17
Fix checkout button image not updating

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-fix-image-checkout-button/react/reference/components/billing/checkout-button

### What does this solve?

See this Slack thread for context: https://clerkinc.slack.com/archives/C01QFRQNHSN/p1759264339893069.

To summarize, we updated the billing components images recently, but the one for CheckoutButton seems to remain the "old" one on the live site. See the issue here. 

Screenshot 2025-10-02 at 1 17 09 PM

This is due to the caching header for the public folder being one year. @NWylynko suggested either changing out the name of the image, so the system considers it brand new or add a v2. 

### What changed?

- Changed the name of the image. 

### Checklist

### DIFF
--- a/docs/guides/development/testing/test-emails-and-phones.mdx
+++ b/docs/guides/development/testing/test-emails-and-phones.mdx
@@ -9,19 +9,18 @@ Verification messages are used during sign-up, sign-in, and when adding an email
 
 ## Limitations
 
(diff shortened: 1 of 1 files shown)


NWylynko (approved):
Not a perfect solution but it works

PR #2739, opened 2025-10-23
Add that webhook handlers can access user private metadata

jfoshee:
Clarified that private metadata is accessible by webhook handlers.

### What does this solve?

- It wasn't clear to me whether this should be true until I saw [this example](https://clerk.com/docs/guides/development/webhooks/overview#:~:text=%22private_metadata%22%3A%20%7B%7D%2C)

Related to USER-2719

### What changed?

- Made it explicit that webhooks can receive private metadata

### Checklist


alexisintech (approved):
cool 😸💖

PR #2748, opened 2025-10-28
chore(how-clerk-works): mention JWT

thiskevinwang:
This adds an explicitly mention of "JWT"... so that the following usage of the "JWT" acronym in the paragraph after the 4 numbered list items is less abrupt.


docs/guides/how-clerk-works/overview.mdx
thiskevinwang:
```suggestion
> How exactly does stateless authentication mitigate database scaling challenges?
```

PR #2800, opened 2025-11-17
Custom OAuth IdP: Removed PKCE limitation

jfoshee:
I tested against our own Public/PKCE provider and it worked fine. We have a toggle under SSO Connections to enable PKCE for the provider.

UPDATE: This [just shipped last week](https://clerk.com/changelog/2025-11-12-pkce-support-custom-oauth)! 🙌 

Screenshot of toggle to enable PKCE on custom SSO provider


SarahSoutoul:
This looks good! Did we want to add anything in the docs showing where this toggle is though / how to use it? @jfoshee

jfoshee:
> This looks good! Did we want to add anything in the docs showing where this toggle is though / how to use it? @jfoshee

@SarahSoutoul Yes, I think that's a good idea. But I recommend a [separate task](https://linear.app/clerk/issue/DOCS-11181/pkce-support-for-custom-oauth-sso-providers) so we can remove the misleading limitation in the mean time.

PR #2815, opened 2025-11-20
Add access token + oauth glossary tooltips

alexisintech:
Adds "access token" tooltip
Adds tooltips for /oauth related glossary terms


alexisintech:
> This is a great addition! One thing tho:
> 
> This would be rare I guess, but since some of the tooltips contain tooltips, it becomes a bit fiddly when trying to hover over those. One of the pages has a tooltip that contains a tooltip, which I hover over, and that one also contains a tooltip, which I can't access then. A bit hard to explain so here is a video of what I mean.
> 
>  Screen.Recording.2025-11-20.at.10.14.05.am.mov

> This is a great addition! One thing tho:
> 
> This would be rare I guess, but since some of the tooltips contain tooltips, it becomes a bit fiddly when trying to hover over those. One of the pages has a tooltip that contains a tooltip, which I hover over, and that one also contains a tooltip, which I can't access then. A bit hard to explain so here is a video of what I mean.
> 
>  Screen.Recording.2025-11-20.at.10.14.05.am.mov

ahh yes :( i've created a ticket here: https://linear.app/clerk/issue/DOCS-11211/tooltip-design-bug-tooltips-within-tooltips-arent-interactable
but let's not let that block this! going to merge it 🚢

PR #2823, opened 2025-11-22
Refresh customization docs

brkalow:
⚠️  **Typedoc PR: https://github.com/clerk/clerk-docs/pull/2870. We will need to combine these PRs to avoid missing pages for incorrect links.** 

✅  Typedoc PR: https://github.com/clerk/javascript/pull/7384

### 🔎 Previews:

- https://clerk.com/docs/pr/brkfeat-customization-docs-tweaks/guides/customizing-clerk/overview
- https://clerk.com/docs/pr/brkfeat-customization-docs-tweaks/guides/customizing-clerk/bring-your-own-css

### What does this solve?

- Clarifies the different layers of Clerk customization so readers can choose the right approach instead of hunting through scattered content
- Moves the Account Portal guide out of the customization section

### What changed?

- Added a `Customizing Clerk` overview plus a dedicated Bring Your Own CSS guide, which moves existing content from the appearance prop overview page to a more use case focused guide
- Reworked the `appearance` prop reference to point toward the new guides and trim redundant explanations
- Relocated `account-portal/overview` to its own guide folder and updated the docs manifest and redirects to match the new structure

### Checklist

### DIFF
--- a/.github/workflows/typedoc-core-2.yml
+++ b/.github/workflows/typedoc-core-2.yml
@@ -1,4 +1,4 @@
-name: Generate Typedocs
+name: Generate Typedocs Core 2
 
 on:
   workflow_dispatch:
@@ -15,7 +15,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: clerk-docs
-          ref: core-3
           fetch-depth: 1
           show-progress: false
 
@@ -24,6 +23,7 @@ jobs:
         with:
           repository: clerk/javascript
           path: clerk-javascript
+          ref: release/core-2
           fetch-depth: 1
           show-progress: false
 
@@ -110,7 +110,7 @@ jobs:
           # 3. Check if PR already exists
           if gh pr list --head typedoc-${{ env.sha }} --json number --jq length | grep -q "^0$"; then
             echo "✅ Creating new PR for typedoc-${{ env.sha }}"
-            gh pr create --base core-3 --head typedoc-${{ env.sha }} --title "[ci] Update Typedoc (${{ env.sha }}) - Core 3" --body "Auto-generated PR to update Typedoc documentation from clerk/javascript@${{ env.sha }}"
+            gh pr create --base main --head typedoc-${{ env.sha }} --title "[ci] Update Typedoc (${{ env.sha }}) - Core 2" --body "Auto-generated PR to update Typedoc documentation from clerk/javascript@${{ env.sha }}"
           else
             echo "ℹ️  PR already exists for typedoc-${{ env.sha }}, skipping PR creation"
           fi
--- a/docs/guides/dashboard/dns-domains/satellite-domains.mdx
+++ b/docs/guides/dashboard/dns-domains/satellite-domains.mdx
@@ -3,6 +3,15 @@ title: Authentication across different domains
 description: Learn how to share sessions across different domains by adding satellite domains to your application.
 ---
 
+<TutorialHero
+  exampleRepo={[
+  {
+    title: "Satellite Domain demo using Turborepo with Clerk, Next.js, and React",
+    link: "https://github.com/clerk/clerk-multidomain-demo"
+  }
+]}
+/>
+
 <Include src="_partials/advanced-feature-callout" />
 
 > [!WARNING]


docs/manifest.json
```diff
@@ -426,11 +443,19 @@
                     "collapse": true
                   },
                   {
-                    "title": "Account Portal",
-                    "href": "/docs/guides/customizing-clerk/account-portal"
+                    "title": "Bring your own CSS",
+                    "href": "/docs/guides/customizing-clerk/bring-your-own-css"
+                  },
+                  {
+                    "title": "Theme Editor",
```
SarahSoutoul: This doesn't have a page assigned to it? @brkalow

SarahSoutoul:
Okay, I have left an initial [docs review](https://github.com/clerk/clerk-docs/pull/2823/commits/92702283744d5c3c2587eeb0d836e77d688d55c9) with mostly writing style / syntax consistency. 

There are some bigger restructuring changes I want to do, but I wanted to first suggest them here and get everyone's opinions as they're important discussion pts @brkalow @alexisintech @manovotny: 

1. As part of this PR, the account portal was moved into its own section. This change was based on the following rationale according to @brkalow: "it doesn't fit under customization IMO. customizing Clerk is mostly focused on customizing components / UI and other code-side settings". I do agree with this reasoning; only thing is it does look a bit odd on its own. One solution could be to destructure the Account portal overview into diff pages like we used to have, or we leave it as is. So I'd like to check opinions on this: 
  
    Screenshot 2025-12-03 at 1 52 03 pm
    
 2. To me, the **Bring your own CSS** page belongs within **Component customization**, so as a sub-sidebar option rather than its own sidebar option. It's still about customizing components, uses the `appearance` prop, and is mentioned within the `appearance` prop itself alongside the other props. So consistency wise, it makes more sense to have it as a sub-section. 
 
 3. I agree that **Build your own UI** should be under **Component customization**, as it's about custom flows. But, we may as well physically move the current **Build your own UI** section under **Development**, rather than link it to lead the user there. This would also align with what we had planned already: https://clerkinc.slack.com/archives/C01QFRQNHSN/p1764110358607419. 
 
 4. I'm not sure about the necessity of the new **Overview** - to me it just feels like a repetition of the **Appearance prop** page but in diff words. It only mentions the stuff under **Component customization**, and not the other stuff. If we want it as an actual overview of the entire **Customizing Clerk** section, I think it should mention other things like **localization**,  **custom menu items**, etc. Or, if it's just for everything under **Component customization**, then it should be moved within that section.

docs/manifest.json
brkalow: Yeah, hasn't shipped yet!

manovotny:
> 1. As part of this PR, the account portal was moved into its own section. This change was based on the following rationale according to @brkalow: "it doesn't fit under customization IMO. customizing Clerk is mostly focused on customizing components / UI and other code-side settings". I do agree with this reasoning; only thing is it does look a bit odd on its own. One solution could be to destructure the Account portal overview into diff pages like we used to have, or we leave it as is.

I understand @brkalow's argument. However, let me offer an alternative view. Although the amount of customization you can do via the dashboard is *extremely limited*, it technically is a layer of basic/easy customization that could be listed in the new **Overview**. It may be enough for some users without having to dip into variables, themes, and CSS.

> 2. To me, the **Bring your own CSS** page belongs within **Component customization**, so as a sub-sidebar option rather than its own sidebar option. 

+1, I agree with this. 
 
> 3. I agree that **Build your own UI** should be under **Component customization**, as it's about custom flows. But, we may as well physically move the current **Build your own UI** section under **Development**, rather than link it to lead the user there. 
 
I agree that having it take you to a completely different section in the sidebar is confusing. Additionally, it's selected in both places in the side bar (though they are far apart from each other), which is odd behavior.

Beyond that, I don't have a strong opinion on where the doc itself should actually live (i.e. in **Development** or **Customizing Clerk**
 
> 4. I'm not sure about the necessity of the new **Overview** - to me it just feels like a repetition of the **Appearance prop** page but in diff words. It only mentions the stuff under **Component customization**, and not the other stuff. If we want it as an actual overview of the entire **Customizing Clerk** section, I think it should mention other things like **localization**,  **custom menu items**, etc. Or, if it's just for everything under **Component customization**, then it should be moved within that section. 

I like having the **Overview** page. Mostly, I like that we talk about the ***layers of customization***. The deeper into the layers you go, you gain more control but incur greater implementation/maintenance costs. As a developer implementing Clerk, I’d like to understand the available options and their trade-offs. Ideally, this page should provide that information.

PR #2843, opened 2025-11-28
Add note about case-insensitivity of allowlist/blocklist identifiers

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-11182/guides/secure/restricting-access

### What does this solve?

The behavior around case-sensitivity for allowlist and blocklist identifiers was not documented, which could lead to confusion when blocking or allowing emails with different capitalization. This update clarifies that identifiers are normalized to lowercase.

### What changed?

Added a note for both allowlist and blocklist sections explaining that identifiers are case-insensitive.

### Checklist

### DIFF
--- a/docs/_tooltips/otp.mdx
+++ b/docs/_tooltips/otp.mdx
@@ -1,0 +1,1 @@
+An **OTP** stands for one-time password and is a code that is used to authenticate a user. The OTP is typically sent to a user's email address or phone number and must be entered within a certain time period to be valid.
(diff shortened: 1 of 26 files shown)


SarahSoutoul:
> Thoughts on making these a `[!NOTE]`?

Yup, I was hesitating btw adding it or not. So yeah happy to change it to a note.

manovotny (approved):

alexisintech (approved):
love that you added an example. thanks sarah!! 😸💖

PR #2909, opened 2025-12-18
feat: Add do not merge workflow

alexcarpenter:
### 🔎 Previews:

Saw this workflow in the clerk_go repo and thought it'd be handy here to. I want to mark a PR as ready for review but don't want to merge it, so we typically have to update the title with DO NOT MERGE. But imo, the label `do not merge` is a nice way to handle this flow.

-

### What does this solve?

-

### What changed?

-

### DIFF
--- a/.github/workflows/check-do-not-merge.yml
+++ b/.github/workflows/check-do-not-merge.yml
@@ -1,0 +1,20 @@
+name: Check for 'do not merge' label
+
+on:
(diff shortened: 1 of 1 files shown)


alexisintech (approved):
oooo sick idea, thank you 😸💖

manovotny (approved):

PR #2928, opened 2026-01-05
Correct DMARC claim in docs

jescalan:
The docs previously claimed that Clerk would have you set up DNS records for DMARC, which is not true. This small PR corrects that claim.

### 🔎 Previews:

- https://clerk.com/docs/pr/jedmarc-correction/guides/development/troubleshooting/email-deliverability


alexisintech (approved):
maybe something changed or maybe it was written wrong all along 🕵️  either way, nice catch and thank you 🙏

PR #2942, opened 2026-01-08
Change tab naming for MCP server guide

SarahSoutoul:
### 🔎 Previews:

-

### What does this solve?

While reviewing this PR, I noticed that the tab in Cursor settings for MCP had changed naming (**Tools & MCP**), and our docs was still referring to the old naming (**Tools & Integrations**). 

This PR makes sure we align with the new naming.


manovotny (approved):
Good catch! 👍

PR #2945, opened 2026-01-09
Add docs for Clerk.billing object

dstaley:
### 🔎 Previews:

- https://clerk.com/docs/pr/dsdocs-billing-namespace/reference/javascript/billing

### What does this solve?

- This PR adds documentation for the `Clerk.billing` namespace. This isn't something we expect very many customers to actually used, but given it's a public API it should be documented.

### What changed?

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 
 Clerk is more than authentication, and we want to convey every feature of Clerk as best as we can to make authentication and user management simpler for everyone. Explaining these concepts can sometimes be tricky, but with the help of a community, we can collaborate, share ideas, find the right words, and achieve greater clarity.
 
-We encourage you to contribute to Clerk's documentation! You can get started by following [this guide](./CONTRIBUTING.md).
+We encourage you to contribute to Clerk's documentation! You can get started by following [this guide](./contributing/CONTRIBUTING.md).
 
 > This repository contains the open-source documentation _content_ for [Clerk docs](https://clerk.com/docs). The _source code_ for the docs site is not yet open source, so if you have any questions or suggestions about _non-content_ related matters, please [reach out](https://clerk.com/support).
 
--- a/docs/_partials/authentication/social-connections/set-client-id-secret.mdx
+++ b/docs/_partials/authentication/social-connections/set-client-id-secret.mdx
@@ -1,5 +1,5 @@
-1. Navigate back to the Clerk Dashboard where the modal should still be open. Paste the **Client ID** and **Client Secret** values that you saved into the respective fields.
-1. Select **Add connection**.
+1. Navigate back to the Clerk Dashboard where the configuration page should still be open. Paste the **Client ID** and **Client Secret** values that you saved into the respective fields.
+1. Select **Save**.
 
 > [!NOTE]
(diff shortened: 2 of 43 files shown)


SarahSoutoul:
> Let me know your thoughts. There is one other thing I'd like to check with you as well. For the following methods: `getPaymentAttempt`, `getPlan`, and `getStatement`, I checked the `javascript` repo code and noticed they didn't take the params naming you passed in the actual docs. For e.g., with `getPaymentAttempt`, in the docs, it states that it takes `GetPaymentAttemptParams` (screenshot 1) but in the actual code (screenshot 2), it just directly takes `{ id: string; orgId?: string }`, so there is discrepancy there. This is not the case for the other methods.

Update here:
- @dstaley will fix the discrepancy on the `javascript` repo itself, as the params naming is correct, and it is the JS repo not having the right naming. 
- Have pushed a few more changes. 

The last thing I wanna check is whether or not we should be adding this new object to the [Clerk class page](https://clerk.com/docs/pr/dsdocs-billing-namespace/reference/javascript/clerk) given it is the main entry point for accessing the billing object. The same way other objects are referenced on the Clerk class page. Same applies for the APIkeys, which currently isn't displayed on the Clerk class page. This would allow for this sentence to be removed: `To use these methods, you must access them through the Clerk object: Clerk.billing.` as I think it causes confusion, and isn't sth we've got for other objects. @alexisintech @manovotny could I have your thoughts on this PR, the changes, and this discussion pt?

manovotny:
> whether or not we should be adding this new object to the [Clerk class page](https://clerk.com/docs/pr/dsdocs-billing-namespace/reference/javascript/clerk) given it is the main entry point for accessing the billing object. The same way other objects are referenced on the Clerk class page. Same applies for the APIkeys, which currently isn't displayed on the Clerk class page.

I feel like both Billing and APIKeys should be listed. Am I missing context as to why they wouldn't be added?

docs/reference/javascript/billing.mdx
```diff
@@ -0,0 +1,283 @@
...
+description: The Billing object provides methods for managing billing for a user or organization.
+sdk: js-frontend
+---
+
+<Include src="_partials/billing/billing-experimental" />
+
+The `Billing` object provides methods for managing billing for a user or organization.
+
+> [!NOTE]
+> If an `orgId` parameter is not provided, the methods will automatically use the current User ID.
```
alexisintech:
```suggestion
> If an `orgId` parameter is not provided, the methods will automatically use the current user's ID.
```

docs/reference/javascript/billing.mdx
```diff
@@ -0,0 +1,283 @@
...
+
+```js
+await clerk.billing.getSubscription({
+  orgId: 'org_123',
+})
+```
+
+### `startCheckout()`
+
+Creates a new billing checkout for the current user or supplied Organization. Returns a[`BillingCheckoutResource`](/docs/reference/javascript/types/billing-checkout-resource) object.
```
alexisintech:
```suggestion
Creates a new billing checkout for the current user or supplied Organization. Returns a [`BillingCheckoutResource`](/docs/reference/javascript/types/billing-checkout-resource) object.
```

PR #2990, opened 2026-01-20
Remove unnecessary icons

SarahSoutoul:
### 🔎 Previews:

-

### What does this solve?

[Linear](https://linear.app/clerk/issue/DOCS-11423/remove-all-icons-except-for-guides-and-reference)

@NWylynko pointed out the **Composables** icon looking out of place. We established that we'd remove all icons except for **Guides** and **Reference**.  This PR makes that change.


manovotny (approved):
👍

PR #2998, opened 2026-01-23
feat(docs): document satelliteAutoSync option for satellite domains

nikosdouvlis:
## Why

Core 3 changes satellite domain behavior — satellite apps no longer automatically redirect to the primary domain on first visit to sync auth state. This eliminates unnecessary latency for apps where most visitors are anonymous. Users need documentation on the new `satelliteAutoSync` option and a migration path from Core 2.

## What changed

- Added "How authentication syncing works" section explaining both sync modes (lazy vs automatic)
- Documented `satelliteAutoSync` prop in ClerkProvider, clerkMiddleware, and backend options
- Added "Migrating from Core 2" section with a comparison table and opt-in instructions
- Updated the "Ready to go" verification steps to reflect the new default behavior


SarahSoutoul:
LGTM @nikosdouvlis!

PR #3067, opened 2026-02-11
core 3: email link custom flows

dstaley:
### 🔎 Previews:

- https://clerk.com/docs/pr/dscore-3-email-links/guides/development/custom-flows/authentication/email-links

- Adds the docs for email link sign-in/sign-up with the new Core 3 hooks

### Deadline

- Feb 13

### DIFF
--- a/docs/_partials/custom-flows/email-link-verification.mdx
+++ b/docs/_partials/custom-flows/email-link-verification.mdx
@@ -1,6 +1,6 @@
 {/* Adds an email address to a user's account, and verifies it using an email link. Expo, iOS, and Android do not support email link verification flows. */}
 
-<If notSdk={["js-frontend", "nuxt"]}>
+<If notSdk={["js-frontend", "nuxt", "vue"]}>
   **This example is written for Next.js App Router but it can be adapted for any React-based framework, such as React Router or Tanstack React Start.**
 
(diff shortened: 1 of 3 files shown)


alexisintech:
sign up flow should be good to go
except when the user clicks the email link and hits `/sign-up/verify`, when the sign-up is complete, it redirects user to `/sign-up` instead of the homepage
per this part of the code
```
// If email link verification status is 'verified', check the `signUp.status`
          void (async () => {
            // If sign-up is complete, finalize the sign-up process
            if (signUp.status === 'complete') {
              const { error } = await signUp.finalize({
                navigate: ({ decorateUrl }) => {
                  const url = decorateUrl('/')
                  if (url.startsWith('http')) {
                    window.location.href = url
                  } else {
                    router.push(url)
                  }
                },
              })
              // If `finalize` fails, check the error
              if (error) {
                console.error(JSON.stringify(error, null, 2))
              }
            } else {
              // If sign-up is not complete, redirect to the sign-up page
              router.push('/sign-up')
            }
          })()
```

indicating that `signUp.status` is not complete - even though it was in this case.
we need to keep that `else` check for progressive sign-ups.

@dstaley is looking into this and how we can update the flow
i'm gonna move on to testing the sign-in flow

alexisintech:
and same thing is happening with sign-in flow 😅

PR #3074, opened 2026-02-12
feat(public-user-data): add username into public user data

Jibaru:
Update PublicUserData with username field.

### 🔎 Previews:

image

Add username field into PublicUserData object.

### DIFF
--- a/docs/reference/javascript/types/public-user-data.mdx
+++ b/docs/reference/javascript/types/public-user-data.mdx
@@ -48,4 +48,11 @@ Information about the user that's publicly available.
   - `string | null`
 
   The user's ID.
+
+  ---
+
+  - `username?`
+  - `string | null`
+
+  The user's username.
 </Properties>


Jibaru:
> Thanks for the PR! I've got a couple of questions:
> 
> 1. Is this Core 3 related at all? Just asking, because there's a separate brand for that.
> 2. I think it might also be missing from the `PublicUserData` interface in the JS repo too, although I could be wrong → [feat(shared): add username to PublicUserData interface javascript#7838](https://github.com/clerk/javascript/pull/7838)
> 3. I am not the most well-versed/experienced (yet) when it comes to how our repos are linked together. Is this something that should automatically happen through our TypeDoc linking @alexisintech?

Hi!
1. It is also core-2
2. Thanks! It was missing, I also added them here: 
https://github.com/clerk/javascript/pull/7837
https://github.com/clerk/javascript/pull/7832

PR #3082, opened 2026-02-13
add billing glossary

dstaley:
### 🔎 Previews:

- https://clerk.com/docs/pr/dsfeat-billing-glossary/guides/billing/glossary

- This PR adds a new Billing glossary page to explain all the various terms that we use and help customers better understand the technical details behind each term.

### Deadline

- No rush

### DIFF
--- a/docs/guides/billing/glossary.mdx
+++ b/docs/guides/billing/glossary.mdx
@@ -7,7 +7,7 @@ description: A collection of terms used in Clerk Billing and what they mean in t
 
 ## Amount
 
-A monetary value denominated in a specific currency. Currently, Clerk Billing currently only supports amounts in USD.
+A monetary value denominated in a specific currency. Currently, Clerk Billing only supports amounts in USD.
 
 ## Billing Period
 


docs/guides/billing/glossary.mdx
```diff
@@ -0,0 +1,82 @@
+---
+title: Clerk Billing Glossary
+description: A collection of terms used in Clerk Billing and what they mean in that context.
+---
+
+<Include src="_partials/billing/billing-experimental" />
+
+## Amount
+
+A monetary value denominated in a specific currency. Currently, Clerk Billing currently only supports amounts in USD.
```
mzhong9723:
```suggestion
A monetary value denominated in a specific currency. Currently, Clerk Billing only supports amounts in USD.
```

docs/guides/billing/glossary.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+The collection of successful payment attempts and applied credits that occurred on a single day in the UTC timezone.
+
+## Subscription
+
+The billing relationship established between a payer and a payee. A subscription is comprised of one or more subscription items. A payer will always have a subscription.
+
+## Subscription Item
+
+The record of the plan and price for a subscription for a given timespan. Changing the plan or price of a subscription always yields a new subscription item. Currently, there can only be one active subscription item at a time, though in the future it may be possible to have multiple active subscription items.
```
mzhong9723: >The record of the plan and price for a subscription for a given timespan.

In my mental model I don't directly associate plans and prices to the top level subscription. I think of a subscription item as what defines the plan and price for a payer during that time period.

mzhong9723: Might need to come back and re-read this next week 😅 

But maybe something like:

"The record of the plan and price for a payer for a given time period."

dstaley: > In my mental model I don't directly associate plans and prices to the top level subscription. I think of a subscription item as what defines the plan and price for a payer during that time period.

Unless I'm misunderstanding, it does appear that your comment and the definition here for subscription item are in alignment. Plans and prices aren't associated to the subscription, they're associated to the subscription item which is why the definition phrases it as "the record of the plan and price for the subscription".

> "The record of the plan and price for a payer for a given time period."

We're trying to be careful about restating relationships, and since a Subscription Item belongs to a Subscription which belongs to a payer, it by relationship is a record relating to a payer, so we don't technically need to re-state that relationship in its definition.

docs/guides/billing/glossary.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Include src="_partials/billing/billing-experimental" />
+
+## Amount
+
+A monetary value denominated in a specific currency. Currently, Clerk Billing currently only supports amounts in USD.
+
+## Billing Period
+
+The cadence in which a subscription is billed. Currently Clerk Billing supports monthly and annual billing periods.
```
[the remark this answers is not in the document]
dstaley: yeah agreed, I wonder if we split the definition? "In the context of a price, a billing period refers to the cadence  in which ... In the context of a subscription item, the billing period refers to the timespan between period_start and period_end." or something

PR #3108, opened 2026-02-19
Simplify docs manifest structure

manovotny:
> [!NOTE]
> This is only an experiment at this point. 
> https://github.com/clerk/clerk/pull/2144 is the other half of this PR.

### Previews

* https://clerk-git-manovotny-simplify-docs-nav-v2.clerkstage.dev/docs/pr/manovotny-simplify-docs-manifest

This PR refactors the documentation manifest system to use flat arrays and SDK-specific manifests instead of nested groups. It removes separators from the navigation structure and consolidates the architecture.

Key changes:
- Reorganized `docs/manifest.json` to use flat array structure
- Created `docs/manifest.android.json` and `docs/manifest.ios.json` for platform-specific navigation
- Updated `docs/manifest.schema.json` to reflect new structure
- Updated build scripts (`build-docs.ts`, `delete-doc.mjs`, `move-doc.ts`, `lib/manifest.ts`) to work with new structure
- Updated corresponding tests

### Deadline

No rush


NWylynko:
I like the direction of this, removing the nested arrays alone is a massive improvement.

Given this is a fundamental change to the manifest, I don't think a simple flags.json flag will be sufficient to get this live.  I think the best option is to have the build script downgrade the new manifest.json style to the existing manifest.json that clerk/clerk is expecting, and then also output to the `dist` a manifest.v2.json or something that is the new manifest syntax. This would allow a clerk/clerk pr to then pull in that new manifest and switch over seamlessly.

Additionally, given existing prs in clerk/clerk-docs probably have manifest.json changes, it would be great to have a `pnpm migrate-manifest` -> `scripts/migrate-manifest.ts` script that can be run to migrate the manifest over to the new syntax to reduce the chance of merge conflicts.

manovotny:
Closing this one out. It was the right experiment — the flat-array + per-SDK-manifest design proved out, and it's now landing for real through the monorepo, rebuilt against current code. This diff had drifted too far to rebase: roughly 90% of it was hand-converted manifest content that content churn kept invalidating.

What shipped matches this PR's shape: `navigationType` on every manifest, single flat arrays, `manifest.ios.json` / `manifest.android.json` discovered by convention, and dist navigation keyed by SDK. Two things changed from the experiment. The conversion runs as a deterministic migration script instead of a hand-edited diff, so it can't go stale the way this one did. And a parity checker proves the rendered nav is byte-for-byte unchanged across every SDK view before anything merges.

The new format will show up here when the mirror syncs after it lands.

PR #3113, opened 2026-02-20
Update old openapi redocly link to scalar

NWylynko:
### 🔎 Previews:

- https://clerk.com/docs/pr/nick-update-redocly-link-to-scalar/guides/development/migrating/overview

Note that reference docs pages are currently broken in preview environments (fixed here https://github.com/clerk/clerk/pull/2146)

- Old link to bapi reference docs using old redocly pathname syntax, updated to scalar pathname

### Deadline

-


manovotny:
Closing this. The link it set out to fix has since been fixed on main, and the redirect it was working around got fixed too.

`docs/guides/development/migrating/overview.mdx` now points at `/docs/reference/backend-api/tag/users/POST/users`, landed in #3381 back in May. Separately, clerk/clerk#2876 fixed the Redocly-to-Scalar redirect shim in July, so the original link resolves to that same URL on its own.

The remaining delta here would make things worse rather than better. Pinning `2026-05-12` opts out of the version injection that `parseApiReferencePath` does for unversioned reference paths, so the link would freeze on one spec version while main's tracks the latest. And the `users.body.password_hasher` deep link isn't a pattern we use anywhere in the docs, which lines up with what Alexis and I both hit when we tried the preview.

Deep-linking to a request body field is still a nice idea for that sentence, since it's specifically about `password_hasher`. Whoever picks that up should start from where main is today.

PR #3175, opened 2026-03-04
fix: remove unexported EmailCodeFactor type predicate from code examples

chriscanin:
Previews:                                                                                                                   
   
  - N/A (no visible rendering change — only code snippet updates)                                                                
                                                                    
  What does this solve? What changed?

  - The EmailCodeFactor type is not exported from @clerk/expo or @clerk/nextjs/types public entrypoints, so code examples using
  (factor): factor is EmailCodeFactor => ... would fail with a TypeScript error when users copy-paste them into their apps.
  - Removed the EmailCodeFactor import and type predicate from all non-legacy code examples, replacing with a simple (factor) =>
  factor.strategy === 'email_code' — which works identically at runtime and doesn't require any type import.
  - Legacy examples are unchanged since they import from @clerk/shared/types (a valid transitive dependency) and need the type
  narrowing to access .emailAddressId.
  - Addresses review feedback from @manovotny on the dashboard Expo quickstart.

  Deadline

  - ASAP — this affects all core-3 custom flow code examples.

  Other resources

  - Dashboard PR review comment from @manovotny identifying the issue

### DIFF
--- a/docs/_partials/custom-flows/email-password-sign-in.mdx
+++ b/docs/_partials/custom-flows/email-password-sign-in.mdx
@@ -2,7 +2,6 @@
 'use client'
 
 import { useSignIn } from '@clerk/nextjs'
-import { EmailCodeFactor } from '@clerk/nextjs/types'
 import { useRouter } from 'next/navigation'
 
 export default function Page() {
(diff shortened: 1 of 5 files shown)


manovotny (approved):
LGTM based on current state. 

However, I'll raise the reverse question: Should the type actually be exported?

PR #3176, opened 2026-03-04
feat: Agent Tasks

tmilewski:
### 🔎 Previews:

- https://clerk.com/docs/pr/tom-user-4863-add-to-clerk-docs/guides/development/testing/agent-tasks
- https://clerk.com/docs/pr/tom-user-4863-add-to-clerk-docs/reference/backend/agent-tasks/create
- https://clerk.com/docs/pr/tom-user-4863-add-to-clerk-docs/reference/backend/agent-tasks/revoke

- This PR adds overview and backend references to the new Agent Tasks feature.

### Deadline

-

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -1417,6 +1417,10 @@
                           "href": "/docs/guides/development/testing/test-emails-and-phones"
                         },
                         {
+                          "title": "Agent Tasks",
+                          "href": "/docs/guides/development/testing/agent-tasks"
+                        },
+                        {
                           "title": "Cypress",
                           "items": [
                             [
@@ -2890,6 +2894,24 @@
                               {
                                 "title": "`revokeSignInToken()`",
                                 "href": "/docs/reference/backend/sign-in-tokens/revoke-sign-in-token"
+                              }
+                            ]
+                          ]
+                        },
+                        {
+                          "title": "Agent Tasks",
+                          "tag": "(Beta)",
+                          "items": [
+                            [
+                              {
+                                "title": "`create()`",
+                                "wrap": false,
+                                "href": "/docs/reference/backend/agent-tasks/create"
+                              },
+                              {
+                                "title": "`revoke()`",
+                                "wrap": false,
+                                "href": "/docs/reference/backend/agent-tasks/revoke"
                               }
                             ]
                           ]
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -2900,7 +2900,6 @@
                         },
                         {
                           "title": "Agent Tasks",
-                          "tag": "(Beta)",
                           "items": [
                             [
                               {
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -2900,6 +2900,7 @@
                         },
                         {
                           "title": "Agent Tasks",
+                          "tag": "(Beta)",
                           "items": [
                             [
                               {
--- a/docs/reference/backend/agent-tasks/revoke.mdx
+++ b/docs/reference/backend/agent-tasks/revoke.mdx
@@ -7,7 +7,7 @@ sdk: js-backend
 Revokes an existing agent task, preventing it from being used to create a session.
 
 ```ts
-function revoke(agentTaskId: string): Promise<AgentTask>
+function revoke(agentTaskId: string): Promise<Omit<AgentTask, 'url'>>
 ```
 
 ## Parameters
--- a/docs/reference/backend/agent-tasks/revoke.mdx
+++ b/docs/reference/backend/agent-tasks/revoke.mdx
@@ -1,0 +1,28 @@
+---
+title: '`revoke()`'
+description: Use Clerk's JS Backend SDK to revoke an agent task.
+sdk: js-backend
+---
+
+Revokes an existing agent task, preventing it from being used to create a session.
+
+```ts
+function revoke(agentTaskId: string): Promise<AgentTask>
+```
+
+## Parameters
+
+<Properties>
+  - `agentTaskId`
+  - `string`
+
+  The ID of the agent task to revoke.
+</Properties>
+
+## Example
+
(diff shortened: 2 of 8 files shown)


docs/reference/backend/agent-tasks/revoke.mdx
```diff
@@ -0,0 +1,28 @@
...
+Revokes an existing agent task, preventing it from being used to create a session.
+
+```ts
+function revoke(agentTaskId: string): Promise<AgentTask>
+```
+
+## Parameters
+
+<Properties>
+  - `agentTaskId`
```
manovotny: This is documented correctly, but calling out that `create()` uses `taskID` and `revoke()` uses `agentTaskId`. It'd probably be better if they were consistently named the same.

docs/reference/backend/overview.mdx
```diff
@@ -79,6 +79,10 @@ The **SAML Connection** resource allows you to manage SAML connections associate
 
 The **Sign-in Token** resource allows you to create and manage sign-in tokens for your application. Sign-in tokens are JWTs that can be used to sign in to an application without specifying any credentials. A sign-in token can be used at most once and can be consumed from the Frontend API using the `ticket` strategy.
 
+### Agent tasks
+
+The **Agent Task** resource allows you to create agent tasks that generate URLs for creating sessions on behalf of users. This is useful for automated testing or agent-driven flows where full authentication isn't practical. The `create()` operation returns an `AgentTask` object containing a `url` that, when visited, creates a session for the specified user.
+
```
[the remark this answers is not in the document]
brkalow: 👍 let's add to the AI Overview as well, good idea!

manovotny:
Pushed some changes directly in 020862d8.

- Fixed the revoke example ID prefix from `agtask_xxx` to `agttsk_xxx` to match the actual `IDPAgentTask` constant in BAPI
- Updated the `AgentTask` `create()` response shape to use `agentTaskId` instead of the deprecated `taskId`
- Switched the reference page examples from `clerk` to `clerkClient` so the guide and reference pages use the same variable
- Added the missing `## Backend API (BAPI) endpoint` sections to the create and revoke reference pages
- Dropped `console.log(agentTask.url)` from the guide and reference examples since the URL is a one-time sign-in link
- Tightened the `permissions` wording — BAPI currently rejects anything other than `'*'`, so the docs now say it's the only supported value

Let me know what you think -- specifically on the ID prefix and `'*'` permission messaging. I want to make sure I'm understanding what I'm seeing on other repos correctly.

docs/reference/backend/overview.mdx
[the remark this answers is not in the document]
manovotny: Added a section to the AI overview in 020862d8.

docs/reference/backend/agent-tasks/revoke.mdx
manovotny: https://github.com/clerk/javascript/pull/8013 is merged and docs here have been updated to reflect.

docs/manifest.json
```diff
@@ -2894,6 +2898,23 @@
                             ]
                           ]
                         },
+                        {
+                          "title": "Agent Tasks",
```
manovotny, line 2902: The `(Beta)` tag was removed here in 0a298a31, but the JS Backend SDK still marks `agentTasks` as `@experimental` ([factory.ts](https://github.com/clerk/javascript/blob/main/packages/backend/src/api/factory.ts#L51-L53) — "private beta", "API is subject to change"). Should we restore the beta tag or drop the experimental marker?

tmilewski: Probably restore the beta tag at the moment.

tmilewski:
@manovotny TY! My only call-out is the removal of the `console.log`. I'm less concerned about the `console.log` specifically, and more about indicating that `.url` can/should be used to authenticate.

> Pushed some changes directly in [020862d](https://github.com/clerk/clerk-docs/commit/020862d802a60873360cbfc44a9279aa2c6080a2).
> 
> * Fixed the revoke example ID prefix from `agtask_xxx` to `agttsk_xxx` to match the actual `IDPAgentTask` constant in BAPI
> * Updated the `AgentTask` `create()` response shape to use `agentTaskId` instead of the deprecated `taskId`
> * Switched the reference page examples from `clerk` to `clerkClient` so the guide and reference pages use the same variable
> * Added the missing `## Backend API (BAPI) endpoint` sections to the create and revoke reference pages
> * Dropped `console.log(agentTask.url)` from the guide and reference examples since the URL is a one-time sign-in link
> * Tightened the `permissions` wording — BAPI currently rejects anything other than `'*'`, so the docs now say it's the only supported value
> 
> Let me know what you think -- specifically on the ID prefix and `'*'` permission messaging. I want to make sure I'm understanding what I'm seeing on other repos correctly.

docs/manifest.json
manovotny: Restored in f534a3fd.

manovotny:
Good call. The reason I dropped the `console.log` is that `agentTask.url` is credential-equivalent — one visit creates an authenticated session, so logging it to stdout risks leaking it through terminal history or CI output. But you're right that the example still needs to surface `.url`.

Compromise in 54c27d75 — added a trailing comment naming `.url` so readers know what to use, without putting it in logs.

manovotny (approved):

tmilewski:
> Good call. The reason I dropped the `console.log` is that `agentTask.url` is credential-equivalent — one visit creates an authenticated session, so logging it to stdout risks leaking it through terminal history or CI output. But you're right that the example still needs to surface `.url`.
> 
> Compromise in [54c27d7](https://github.com/clerk/clerk-docs/commit/54c27d754e392d949d62a85a4d044c47cc95e71d) — added a trailing comment naming `.url` so readers know what to use, without putting it in logs.

LGTM!

PR #3184, opened 2026-03-05
docs: Add buildSignInUrl guidance for satelliteAutoSync: false

jacekradko:
## Summary

- Adds a new section to the satellite domains guide explaining how to use `buildSignInUrl()` / `buildSignUpUrl()` when `satelliteAutoSync` is set to `false`
- Includes code examples for both vanilla JS and React
- Adds a callout warning against hardcoding sign-in URLs, which would skip the sync trigger parameter

## Context

A user with a vanilla JS satellite implementation set `satelliteAutoSync: false` and found that sessions weren't syncing. The issue was that their sign-in links were hardcoded instead of using `Clerk.buildSignInUrl()`, which appends the `__clerk_synced=false` parameter needed to trigger the handshake on return.

## Test plan

- [ ] Verify docs render correctly on preview
- [ ] Verify code examples are accurate

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SarahSoutoul (approved):
LGTM!

PR #3186, opened 2026-03-05
chore: x-twitter update

thiskevinwang:
For X/Twitter SSO connection, not the older Twitter one, we're supporting

- Email retrieval (possible as of 2025)
- Shared dev creds


BrandonRomano (approved):
Let's see if my stamp works

PR #3226, opened 2026-03-18
Reduce Next Steps nb of cards to 4 maximum

SarahSoutoul:

We recently added click tracking to the `<Card>` component used in Next Steps sections. This is useful for the docs team to better understand post-quickstart or post-guide user behaviour and to help with the broader initiative of using data to inform what should be included in Next Steps. This will drive future PRs in improving our Next Steps experience. 

However, in the meantime, this PR reduces the numbers of `<Cards>` in all Next Steps section to not go above 4. Having 6 cards can lead to an overwhelming user experience + we want to make the Next Steps as consistent as possible aka keeping it to 3/4 cards for each section. 

My reasoning behind some of the changes:

1. Redundancy: if two cards pointed to overlapping concepts, I kept the one that matched the page's focus and removed the more generic one.
2. Depth over breadth: I prioritized a tighter path that gets users to their next concrete step, rather than offering every possible path.
3. Consistency: I kept each Next steps block to 3–4 cards and pushed any “Reference” link to be last (or removed if the block was already at 4).
4. Relevance: cards that felt less directly related to the page’s topic were removed first.
 
### Deadline

No rush.


docs/guides/organizations/add-members/invitations.mdx
```diff
@@ -165,8 +165,8 @@ Use the following tabs to see examples for each method.
 Now that you understand how to invite users to your Organization, you can:
 
 <Cards>
```
alexisintech: I'd probably put the invitation metadata as the first card! As it'd be an extension to this guide

PR #3279, opened 2026-04-09
Update docs for SCIM GA (some features in beta)

im2nguyen:
### 🔎 Previews:

- 

- These docs changes update the SCIM and enterprise connection docs to support custom attribute mapping and group -> role mapping. This also updates SCIM to public beta, however calls out that folks should reach out to team-orgs@clerk.dev if they run into any compatibility issues.

### Deadline

-


docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -52,7 +49,7 @@ Once Directory Sync is enabled in Clerk, configure your IdP to point to Clerk's
 1. Set the **Unique identifier field for users** to `userName`.
 1. Enter the **Bearer token** from Clerk into the **API Token** field.
 1. Select **Test API Credentials** to verify the connection, then select **Save**.
-1. Under **To App**, enable the operations you want Okta to push to Clerk: **Create Users**, **Update User Attributes**, and **Deactivate Users**.
+1. Under **To App**, enable the operations you want Okta to push to Clerk: **Create Users**, **Update User Attributes**, **Push Groups**, and **Deactivate Users**.
```
manovotny: Is `Push Groups` actually a toggle inside the **To App** section in Okta? 

My understanding is that Create/Update/Deactivate live under **To App** provisioning, but Group Push is configured on its own dedicated tab within the app's Provisioning section. 

If that's right, readers who follow this step literally won't find a `Push Groups` checkbox here.

Maybe we should split this into two steps? One for the **To App** toggles and a separate one for enabling Push Groups on its own tab?

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -81,6 +78,57 @@ The following user attributes are synced from your IdP to Clerk via SCIM. The ex
...
+> [!TIP]
+> IdP groups are only available to map after they've been pushed to Clerk at least once. If you don't see a group, make sure **Push Groups** is enabled in your IdP and that at least one sync has occurred.
+
+### Multi-group precedence
+
+If a user belongs to multiple groups that each map to different Clerk roles, Clerk applies the role according to a configurable **precedence order**. The role with the highest precedence is applied. You can configure the precedence order in the Clerk Dashboard under **Role mapping** → **Precedence**.
+
+### Group deletion behavior
+
+When a group is deleted in your IdP, Clerk removes the group and reverts all affected users to the organization's default role. When a role mapping changes, a background job queues role reassignment for all affected group members.
```
manovotny: This seems to conflict with the **Multi-group precedence** section just above. 

If a user belongs to groups A (role X) and B (role Y), and group A is deleted, do they get role Y (re-evaluated against their remaining mappings) or fall back to the org's default role? 

The precedence model would suggest the former, but this sentence suggests the latter.

Could we clarify the actual behavior here?

Separately, the trailing sentence about role mapping changes triggering a background reassignment isn't really about group deletion. Should that live in its own subsection or under the **How it works** so the section title matches its contents?

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
```diff
@@ -100,6 +100,15 @@ This guide shows you how to set up a SAML connection with a custom IdP in Clerk.
 
   Learn more about [how to access the metadata from our APIs](/docs/guides/users/extending).
 
+  #### Map custom attributes (beta)
```
manovotny, line 103: Small consistency question.

The OIDC version of this same section is `## Map custom attributes (optional)`, and here it's `#### Map custom attributes (beta)`. 

Just double checking the should the different and not the same (ie. they're in different stages of maturity/availability).

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
```diff
@@ -100,6 +100,15 @@ This guide shows you how to set up a SAML connection with a custom IdP in Clerk.
...
+
+  In addition to the standard attribute mappings, you can sync custom attributes from your IdP (such as `department` or `employee_id`) into Clerk's `publicMetadata`. Custom attribute definitions are configured at the enterprise connection level and are shared with your SCIM/Directory Sync connection if one is enabled.
```
manovotny, line 107: Just above this, the **Map other claims** section explains how to land IdP claims in `publicMetadata` by prepending `public_metadata_` to the claim name. 

Now we're also introducing the shared custom-attribute flow, which also writes to `publicMetadata`. 

From a reader's perspective it's not obvious which mechanism they should reach for first. Are they equivalent? Is the prefix approach the ad-hoc option and this one the recommended path? Or is `public_metadata_*` being phased out?

A one-line clarifier at the top of this subsection distinguishing the two would probably save readers (and support) a round trip.

manovotny (commented):
Thanks for the PR @im2nguyen!

I made a direct commit for minor fixes. 94557f8

- `directory-sync.mdx:7` — Replaced "Our" / "We" with "Clerk's" / "Clerk has" in the public beta callout
- `custom-attribute-mapping.mdx:19` — Added missing Oxford comma in the standard attributes list.
- `directory-sync.mdx:83` — Reworded "When SCIM is enabled, it is the exclusive source..." to "When Directory Sync is enabled, SCIM becomes the exclusive source..." to match the phrasing in the SAML/OIDC callouts and resolve the ambiguous "it".
- `directory-sync.mdx:95` — Changed "click" → "select" (styleguide) and fixed "teh" → "the" typo.
- `directory-sync.mdx:105` — Fixed subject-verb agreement ("role mappings uses" → "role mapping uses"), lowercased "Enterprise Connection" to match the rest of the doc, and linked "role set" properly.
- `directory-sync.mdx:153` — Fixed "investiaging" → "investigating" typo.

I've left a couple other comments for my outstanding questions.

manovotny:
I also fixed build and lint errors. 54afdc00b77d5124bf05bf61778af32aa07c8082

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -81,6 +78,57 @@ The following user attributes are synced from your IdP to Clerk via SCIM. The ex
...
+
+For example, you might map the shared `department` attribute to `urn:ietf:params:scim:schemas:extension:enterprise:2.0:User:department` in Okta.
+
+## Role mapping
+
+Role mapping lets you automatically assign Clerk roles to users based on their IdP group memberships. This eliminates manual role management in Clerk and lets your IdP serve as the single source of truth for access levels.
+
+You must link the enterprise connection to an organization to use Directory Sync role mapping. The role mapping uses the linked organization's [role set](/docs/guides/organizations/control-access/role-sets). If you delete a role that's associated with a role map, Clerk will [remap the organization member roles](/docs/guides/organizations/control-access/role-sets#remapping-flow).
+
+When SCIM is enabled, roles managed through SCIM cannot be manually modified in Clerk. They are controlled entirely by group membership in your IdP. Roles become manually editable again when SCIM is disabled.
```
kalafut: >roles managed through SCIM cannot be manually modified in Clerk

Does this refer specifically role _assignments_? As written is sort of sound like I couldn't even, say, rename a role in Clerk.

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -81,6 +78,57 @@ The following user attributes are synced from your IdP to Clerk via SCIM. The ex
...
+1. Select the **Directory Sync** tab and navigate to **Role mapping**.
+1. For each IdP group that has been synced, select the Clerk role it should map to.
+1. Select **Save**.
+
+> [!TIP]
+> IdP groups are only available to map after they've been pushed to Clerk at least once. If you don't see a group, make sure **Push Groups** is enabled in your IdP and that at least one sync has occurred.
+
+### Multi-group precedence
+
+If a user belongs to multiple groups that each map to different Clerk roles, Clerk applies the role according to a configurable **precedence order**. The role with the highest precedence is applied. You can configure the precedence order in the Clerk Dashboard under **Role mapping** → **Precedence**.
```
gabrielmeloc22: we might need to update this since we don't have a section solely for the precedence. and instead specify that the user can drag the mappings and change the order.

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
gabrielmeloc22: I agree, the actual behavior is to resolve to next role based on the precedence and mapping configure. We would only fallback to the org's default role if there are no more matching mappings available.

gabrielmeloc22: I think we can have a section specify some of the behaviors we should expect when operating over mappings and associated entities. For example, when a mapping is updated, created or deleted, a background will resolve the roles of all affected members. As well as when the group gets deleted on the IdP or the role is removed on Clerk.

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
im2nguyen: I believe they should be in the same state. @kalafut and @gabrielmeloc22 were checking if oidc custom attribute mapping behaved the same

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
im2nguyen: will clarify, this is less about "roles" but more about the roles attached to a group membership

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
im2nguyen: yeah that makes sense. we're supporting both methods. the `public_metadata_` prefix is how you would do it through saml, while this is for scim. will add a clarifying sentence

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
iagodahlem: I took that section off from OIDC since we still don't have custom attributes for OIDC connections; it will be available only for SAML for now.

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,84 @@
...
+## Map attributes for SCIM (Directory Sync)
+
+When Directory Sync is enabled, SCIM is the **exclusive source of truth** for attribute values. SCIM-managed attributes are read-only in Clerk and cannot be edited from the Dashboard or by the user in `<UserProfile />`. When SCIM is disabled, attributes become editable again.
+
+Clerk's SCIM implementation exposes a `/Schemas` endpoint that advertises a custom schema extension (`urn:clerk:scim:schemas:extension:custom:2.0:User`). This tells your IdP that Clerk accepts arbitrary custom attributes, which are mapped directly to `publicMetadata`.
+
+For each custom attribute, you map the shared attribute to the SCIM schema attribute name your IdP sends:
+
+1. In the Clerk Dashboard, navigate to the **Directory Sync** tab on your enterprise connection.
+1. Under **Attribute mapping**, use the dropdown (populated from your IdP's `/Schemas` endpoint) to select the SCIM attribute for each custom attribute.
```
manovotny: This says "your IdP's `/Schemas` endpoint" but Clerk is the SCIM service provider exposing `/Schemas`. Saying "your" could make the user think it's something they need to implement. 

`directory-sync.mdx` already has the correct phrasing. We should match it:

```suggestion
1. Under **Attribute mapping**, use the dropdown (populated from the SCIM `/Schemas` endpoint) to select the SCIM attribute for each custom attribute.
```

manovotny (commented):
Small suggested fix on a "your" phrase, but otherwise, LGTM! 

Proactively approving.

manovotny (approved):

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,84 @@
...
+Attribute configuration happens in two layers:
+
+1. **Shared attribute pool** — You define the attributes you care about once, at the enterprise connection level. Each attribute maps to a key in the user's `publicMetadata`. These definitions are shared across SSO and SCIM for that connection.
+1. **Connection-specific mappings** — For each connection type (SAML, OIDC, or SCIM), you map the shared attribute to the IdP's specific field name or schema attribute.
+
+This keeps your schema consistent while allowing for the different naming conventions each IdP uses.
+
+## Common attributes
+
+Standard attributes (for example, username, first name, last name, and email addresses) are pre-configured for every enterprise connection. Clerk controls the mapping logic for these and they can be enabled or disabled per connection, but you don't need to define them manually.
```
iagodahlem: Common attributes can't be enabled/disabled:

```suggestion
Standard attributes (for example, username, first name, last name, and email addresses) are pre-configured for every enterprise connection. Clerk controls the mapping logic for these, but you don't need to define them manually.
```

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+### Define custom attributes
+
+1. In the Clerk Dashboard, navigate to the [**SSO connections**](https://dashboard.clerk.com/~/user-authentication/sso-connections) page.
+1. Select the enterprise connection you want to configure.
+1. Select the **Attribute mapping** tab.
+1. Under **Custom attributes**, select **Add attribute**.
+1. Enter a **Label** for the attribute (e.g., `Department`).
+1. Enter the `publicMetadata` key where this value will be stored (e.g., `department`).
+1. Select **Save**.
```
iagodahlem:
```suggestion
1. In the Clerk Dashboard, navigate to the [**SSO connections**](https://dashboard.clerk.com/~/user-authentication/sso-connections) page.
1. Select the enterprise connection you want to configure.
1. In the **Overview** tab, go to **Identity provider attributes**section.
1. Under **Custom attributes**, select **Create custom attribute**.
1. Enter a **Display name** for the attribute (e.g., `Department`).
1. Enter the `publicMetadata` **Key** where this value will be stored (e.g., `department`).
1. Select **Create attribute**.
```

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+### OIDC
+
+In OIDC, attributes are returned as claims in the ID token or userinfo endpoint. For each custom attribute:
+
+1. In the Clerk Dashboard, select the **Attribute mapping** tab on your OIDC connection.
+1. For each custom attribute, enter the claim name your IdP returns (e.g., `department`).
+1. In your IdP, ensure the claim is included in the token and mapped correctly.
+
+For a full walkthrough, see the [custom OIDC provider guide](/docs/guides/configure/auth-strategies/enterprise-connections/oidc/custom-provider).
```
iagodahlem: No custom attributes for OIDC yet.

```suggestion
```

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -81,6 +78,60 @@ The following user attributes are synced from your IdP to Clerk via SCIM. The ex
 > [!NOTE]
 > Attributes managed by Directory Sync are read-only in Clerk — they cannot be edited from the Clerk Dashboard or by the user in `<UserProfile />`. Changes must be made in the IdP and will sync automatically.
 
+### Custom attribute mapping
+
```
iagodahlem, line 80:
```suggestion
> [!NOTE]
> Custom attribute mapping is in public beta.
```

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -91,9 +142,17 @@ This is useful as a first debugging step. If a user isn't appearing in your app
...
+- `scim.provisioning_failed` — SCIM provisioning failed
+
+**Group events:**
+
+- `scim.group.created` — a group was pushed from the IdP
+- `scim.group.patched` — a group's name or membership was updated
+- `scim.group.deleted` — a group was deleted in the IdP
 
-Selecting an event shows its full details: event\_id, timestamp, and the complete event payload including the data received from the IdP. This is helpful for verifying exactly what your IdP sent. For example, confirming that the right attributes were included in a `scim.user.patched` event or investigating why a user wasn't created as expected.
+Selecting an event shows its full details: event\_id, timestamp, and the complete event payload including the data received from the IdP. This is helpful for verifying exactly what your IdP sent. For example, confirming that the right attributes were included in a `scim.user.patched` event, or investigating why a user wasn't created as expected.
```
iagodahlem: The events feature won't be released as part of SCIM, so we need to take this section off for now.

```suggestion
```

docs/guides/configure/auth-strategies/enterprise-connections/saml/custom-provider.mdx
```diff
@@ -100,6 +100,17 @@ This guide shows you how to set up a SAML connection with a custom IdP in Clerk.
 
   Learn more about [how to access the metadata from our APIs](/docs/guides/users/extending).
 
+  #### Map custom attributes (beta)
+
+  The `public_metadata_*` claim-name prefix described above is the SAML-only way to land IdP claims in `publicMetadata`. The shared custom attribute flow below is the recommended path when Directory Sync is also enabled on the connection, because the same attribute definitions apply to both SSO and SCIM.
```
kalafut:
```suggestion
  The `public_metadata_*` claim-name prefix described above is the SAML-only way to add IdP claims to `publicMetadata`. The shared custom attribute flow below is the recommended path when Directory Sync is also enabled on the connection, because the same attribute definitions apply to both SSO and SCIM.
```

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: Custom Attribute Mapping
+description: Learn how to sync custom attributes from your Identity Provider to Clerk's public metadata, shared across SSO and Directory Sync connections.
+---
+
+Custom attribute mapping lets you sync additional user data from your Identity Provider (IdP) (such as department, employee ID, or cost center) into Clerk's `publicMetadata`. Attribute definitions are configured at the **enterprise connection level** and are shared between your SSO connection (SAML or OIDC) and your Directory Sync (SCIM) connection, so the same attributes are available regardless of how a user authenticates or is provisioned.
```
manovotny: The dedicated `### OIDC` subsection was removed from this page, but the intro on line here still says the flow is `shared between your SSO connection (SAML or OIDC)` and the "how it works" list on line 13 says `(SAML, OIDC, or SCIM)`.

So the page claims OIDC is supported but never walks the reader through how to map an OIDC claim.

Are OIDC custom attributes actually supported? If yes, should we restore a short OIDC subsection (or cross-link). If no, should these two references drop OIDC so the page scopes to SAML + SCIM only?

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -15,16 +15,13 @@ This is especially useful for organizations that manage large user bases through
...
-- **Update**: When common user attributes (such as name or email) are updated in the IdP, Clerk syncs those changes automatically. A `user.updated` webhook event is fired.
+- **Update**: When user attributes (such as name, email, or custom attributes) are updated in the IdP, Clerk syncs those changes automatically. A `user.updated` webhook event is fired.
 - **Delete / Disable**: When a user is removed or deactivated in the IdP, Clerk deactivates the corresponding Clerk user and immediately revokes all of their active sessions. A `user.updated` webhook event is fired.
 
-These webhook events are separate from the [Directory Sync events](#debug-with-directory-events) logged in the Dashboard, which provide more detailed SCIM-specific debugging information. Webhook payloads for SCIM-triggered events aren't finalized yet.
+These webhook events are separate from the Directory Sync events logged in the Dashboard, which provide more detailed SCIM-specific debugging information. Webhook payloads for SCIM-triggered events aren't finalized yet.
```
manovotny: A few commits ago the `## Debug with directory events` section and its anchor were both removed, but this sentence still tells the reader that "Directory Sync events logged in the Dashboard" provide "more detailed SCIM-specific debugging information."

Is the Directory Events surface shipping as part of this release or not? If not, this sentence could point readers at a UI that doesn't exist. Should we drop it, or rewrite to talk about webhooks only?

manovotny (commented):
Saw more changes come in. A few more questions.

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
iagodahlem: Yeah, OIDC won't be supported yet, removing the references:

```suggestion
Custom attribute mapping lets you sync additional user data from your Identity Provider (IdP) (such as department, employee ID, or cost center) into Clerk's `publicMetadata`. Attribute definitions are configured at the **enterprise connection level** and are shared between your SSO connection (SAML) and your Directory Sync (SCIM) connection, so the same attributes are available regardless of how a user authenticates or is provisioned.
```

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## How it works
+
+Attribute configuration happens in two layers:
+
+1. **Shared attribute pool** — You define the attributes you care about once, at the enterprise connection level. Each attribute maps to a key in the user's `publicMetadata`. These definitions are shared across SSO and SCIM for that connection.
+1. **Connection-specific mappings** — For each connection type (SAML, OIDC, or SCIM), you map the shared attribute to the IdP's specific field name or schema attribute.
```
iagodahlem:
```suggestion
1. **Connection-specific mappings** — For each connection type (SAML or SCIM), you map the shared attribute to the IdP's specific field name or schema attribute.
```

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
iagodahlem: Not yet too, I'll drop it.
```suggestion
```

PR #3280, opened 2026-04-09
Add new create props to sign-in/up future

dstaley:
### 🔎 Previews:

- https://clerk.com/docs/pr/dsfix-future-tickets/nextjs/reference/objects/sign-in-future
- https://clerk.com/docs/pr/dsfix-future-tickets/nextjs/reference/objects/sign-up-future
- https://clerk.com/docs/pr/dsfix-future-tickets/guides/development/custom-flows/organizations/accept-organization-invitations

- This PR updates the types for `signUp.create()` and `signIn.create()` to reflect new parameters added in a recent SDK release.
- This PR also updates the docs for accepting organization invitations to reflect the new Core 3 APIs (this was missed in our initial Core 3 docs work).

### Deadline

- No rush!

### DIFF
--- a/docs/guides/development/custom-flows/organizations/accept-organization-invitations.mdx
+++ b/docs/guides/development/custom-flows/organizations/accept-organization-invitations.mdx
@@ -23,14 +23,14 @@ The `__clerk_ticket` query parameter contains the ticket token, which is essenti
 
 The `__clerk_status` query parameter is the outcome of the ticket verification and will contain one of three values:
 
-- `sign_in` indicates the user already exists in your application. You should create a sign-in flow using the invitation token by extracting the token from the URL and passing it to the [`signIn.create()`](/docs/reference/objects/sign-in#create) method.
-- `sign_up` indicates the user doesn't already exist in your application. You should create a sign-up flow using the invitation token by extracting the token from the URL and passing it to the [`signUp.create()`](/docs/reference/objects/sign-up#create) method.
+- `sign_in` indicates the user already exists in your application. You should create a sign-in flow using the invitation token by extracting the token from the URL and passing it to the [`signIn.ticket()`](/docs/reference/objects/sign-in-future#ticket) method.
+- `sign_up` indicates the user doesn't already exist in your application. You should create a sign-up flow using the invitation token by extracting the token from the URL and passing it to the [`signUp.create()`](/docs/reference/objects/sign-up-future#create) method.
 - `complete` indicates the user already exists in your application, and was signed in. The flow has been completed and no further actions are required.
 
 The following example demonstrates how to handle both sign-up and sign-in flows using the invitation token:
 
 1. It extracts the token from the URL.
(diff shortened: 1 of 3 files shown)


manovotny:
Did a review pass on this and pushed a handful of fixes in 8d5dafb0f. Summary:

**`accept-organization-invitations.mdx`**

- Replaced undefined `err` with `error` in the 4 error handlers (Next.js + Expo, sign-in + sign-up) — the `try/catch` was removed but `console.error` still referenced the old `err` variable.
- Added `useRouter` import and `const router = useRouter()` in both the Next.js and Expo samples — `router.push()` was being called without `router` ever being defined.
- Dropped the invalid `router.push(url as Href)` cast from the Next.js sample (`Href` is an `expo-router` type, and Next.js `router.push()` takes a plain string).
- Imported `type Href, useRouter` from `expo-router` in the Expo sample so the existing `router.push(url as Href)` cast resolves.
- Finished the Future API migration in the Expo sample — dropped the leftover `isLoaded` / `setActive: setActiveSignUp` / `setActive: setActiveSignIn` destructures and the `!signIn || !setActiveSignIn` guard that no longer match the Signal-based hook shape (verified against `packages/shared/src/types/state.ts`).
- Dropped the unused `errors: signUpErrors` / `errors: signInErrors` renames from the Next.js destructures.
- Fixed both sign-up `else` branches that incorrectly said `// If the sign-in attempt is not complete, check why.` — they now say `// Check why the sign-up is not complete.`.
- Reworded "`signUp.create()` will throw an error" to "`signUp.create()` will return an error" now that the Future API surfaces errors via `{ error }` instead of throwing.
- Added `token`, `organization`, and `accountStatus` to both `useEffect` dep arrays — the guard and body read all three, so `[signIn]` alone was missing deps.

**`sign-in-future.mdx`**

- Added the `signUpIfMissing?: boolean` property to `SignInFutureCreateParams`. It's present on the type (`packages/shared/src/types/signInFuture.ts:49`) but wasn't documented here.

**`sign-up-future.mdx`**

- Fixed the intro on line 11, which referenced `SignInFuture` instead of `SignUpFuture`.
- Moved the new `strategy?` and `password?` params into their correct alphabetical positions inside `SignUpFutureCreateParams`.
- Updated the `strategy?` type union to use `<code>[OAuthStrategy](/docs/reference/types/sso#o-auth-strategy) | ...</code>` to match the convention used in `sign-in-future.mdx`, `sign-in.mdx`, `sign-up.mdx`, etc.
- Added the missing `'oauth_token_apple'` and `'phone_code'` strategy values (both are on the type at `packages/shared/src/types/signUpFuture.ts:58-59`) with short descriptions.

That sounds like a lot more than it actually is, I promise. 😊 Let me know if anything here doesn't land the way you expected — happy to iterate.

dstaley:
@manovotny that all looks great to me!

manovotny (approved):

PR #3286, opened 2026-04-13
Generate object reference docs

alexisintech:
⚠️  In conjunction with https://github.com/clerk/javascript/pull/8276
⚠️  Pulled in Typedoc changes [[ci] Update Typedoc (8e1bd48) - Core 3](https://github.com/clerk/clerk-docs/pull/3416#top)

To test:
- pull https://github.com/clerk/javascript/pull/8276 
- run `pnpm run typedoc:generate:skip-build` (as long as you have `clerk-docs` and `clerk/javascript` symlinked; see the [`clerk/clerk README.md`](https://github.com/clerk/clerk))
- run the docs (in `clerk/clerk`, run `pnpm run dev`)
- preview the docs locally

It's time to get our `/objects` reference docs generated using typedoc!

### Related resources

[`clerk/javascript` PR](https://github.com/clerk/javascript/pull/8276)
https://linear.app/clerk/issue/DOCS-11789/generate-objects-docs


SarahSoutoul:
Okay, have pushed a minor [docs review - clerk-options](https://github.com/clerk/clerk-docs/pull/3286/commits/e37011ce66dd8898e261db1e8ff6a92efcc748bb) after reviewing the Typedoc output of the `shared/clerk-options`. Only does two things:

- Adds a missing sentence which is present on the current docs.
- Fixes the weird layout caused by <If> wrapping. 

Screenshot 2026-04-15 at 3 03 00 pm

Will continue reviewing and push a second docs review with the rest.

SarahSoutoul:
@alexisintech Here is a second [docs review pt 2](https://github.com/clerk/clerk-docs/pull/3286/commits/0c54c2d1c081c118526743afd7f00471f2638be8) to just add the Typedoc outputs for the intros of each object. 

Noticed they were there, looked good, but weren't being used. Again, if you don't love it, feel free to revert. This is in conjunction with the [docs review](https://github.com/clerk/javascript/pull/8276/commits/6fef09a71a46663f417ba6c65c627a89356fe583) in `javascript` repo which documents the intro of the Clerk object better.

alexisintech:
> @alexisintech Here is a second [docs review pt 2](https://github.com/clerk/clerk-docs/pull/3286/commits/0c54c2d1c081c118526743afd7f00471f2638be8) to just add the Typedoc outputs for the intros of each object.
> 
> Noticed they were there, looked good, but weren't being used. Again, if you don't love it, feel free to revert. This is in conjunction with the [docs review](https://github.com/clerk/javascript/pull/8276/commits/6fef09a71a46663f417ba6c65c627a89356fe583) in `javascript` repo which documents the intro of the Clerk object better.

me personally, I liked having the intro text in the mdx but your way of using typedoc for the intro text is fine too. I prefer the first way because if we ever want to update copy, we don't have to rely on a `clerk/javascript` PR (its fine for generated reference content because that's the point of it! but in the name of trying to minimize generating prose). however, either way we decide, I'm impartial ^~^

SarahSoutoul:
> me personally, I liked having the intro text in the mdx but your way of using typedoc for the intro text is fine too. I prefer the first way because if we ever want to update copy, we don't have to rely on a `clerk/javascript` PR (its fine for generated reference content because that's the point of it! but in the name of trying to minimize generating prose). however, either way we decide, I'm impartial ^~^

@alexisintech Yeah that's very _fair_ feedback. Reverted changes: [Revert 2nd docs review](https://github.com/clerk/clerk-docs/pull/3286/commits/0061ddacffaeef2527eef19a9ad9711198ce4dbf).

docs/reference/types/handle-o-auth-callback-params.mdx
```diff
@@ -0,0 +1,9 @@
+---
+title: '`HandleOAuthCallbackParams`'
+description: A type representing the parameters for an OAuth-related callback.
+sdk: js-frontend, astro, chrome-extension, expo, nextjs, react, react-router, tanstack-react-start, nuxt, vue
+---
+
+A type representing the parameters for a callback that handles an OAuth-related callback.
+
+<Typedoc src="shared/handle-o-auth-callback-params" />
```
SarahSoutoul, line 9: Two links are broken in this partial - the opaque token and first factor verifications links should turn into tooltips in the rendered output, just looking at the `javascript` JSDoc. So when clicking on the links, we've got 404s as a result. 

**Example** 

Image

Image

Image

docs/reference/components/control/clerk-loaded.mdx
```diff
@@ -93,7 +93,9 @@ The `<ClerkLoaded>` component guarantees that the Clerk object has loaded (the `
 
   #### `ClerkOptions`
 
-  <Include src="_partials/javascript/clerk-options" />
+  The `load()` method accepts an optional object that accepts the following props. All props are optional.
+
+  <Typedoc src="shared/clerk-options" />
```
SarahSoutoul, line 98: Checked this and found two missing properties in the generated Typedoc compared to the current docs:

- routerPush?
- routerReplace?

Image

From what I can tell, both should be included as part of `ClerkOptionsNavigation`.

Image

I think the issue is that `ClerkOptionsNavigation` is a union type, so the current flattening logic in `.typedoc/custom-theme.mjs` is not picking up those properties when `ClerkOptions` is rendered. It also seems like `ClerkOptionsNavigation` needs to be exported so TypeDoc can resolve it consistently during generation.

After poking at `.typedoc/custom-theme.mjs`, it looks like we may need to handle union object arms when merging intersection properties, otherwise `routerPush` / `routerReplace` get dropped from the final output.

docs/reference/types/billing-initialized-payment-method-resource.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: '`BillingInitializedPaymentMethodResource`'
+description: The BillingInitializedPaymentMethodResource type represents a payment method that has been initialized for a checkout session.
+sdk: js-frontend, astro, chrome-extension, expo, nextjs, react, react-router, tanstack-react-start, nuxt, vue
+---
+
+<Typedoc src="shared/billing-initialized-payment-method-resource" />
```
SarahSoutoul, line 7: Do we want to link the `<PaymentElement/>` component?

docs/reference/types/handle-o-auth-callback-params.mdx
SarahSoutoul: Seems I had mentioned it on the other PR already, so might already be on your agenda to fix: https://github.com/clerk/javascript/pull/8276#issuecomment-4263637346.

docs/reference/objects/api-keys.mdx
SarahSoutoul, line 1: Reviewed this comparing with current docs and noticed two things:

- Broken **Active Organization** tooltip (already flagged before) but making sure we have a paper trail to check against later. 
- When comparing the intros for `create()`, `getAll()` and `revoke()` btw this PR and current docs, found that we have missing info that is on current docs. This might be intended so didn't want to mess with it but I think for `create()` specifically, some of the info is key to have. 

**Example with `create()`** 

**Current docs**

Image

Vs. **this PR**

Image

As an isolated note, will review the objects one by one and prob leave a general comment like this on each file corresponding to each object if there is anything I notice.

docs/reference/types/handle-o-auth-callback-params.mdx
alexisintech: yeah we already talked about that in the clerk javascript PR - tooltips need to be fixed and is on the agenda. let's focus on the overall review for now

docs/reference/components/control/clerk-loaded.mdx
alexisintech: nice catch ! I'll get this fixed

docs/reference/objects/api-keys.mdx
alexisintech: this is a good catch! let me add that information back

docs/reference/types/billing-initialized-payment-method-resource.mdx
alexisintech: sure!

docs/reference/components/control/clerk-loaded.mdx
alexisintech: done [d5c6cd0](https://github.com/clerk/javascript/pull/8276/commits/d5c6cd0ccd324d41e641094808e7f16fe08472ef)

docs/reference/types/billing-initialized-payment-method-resource.mdx
alexisintech: done [4c39d12](https://github.com/clerk/javascript/pull/8276/commits/4c39d1289c53c8554b026041654408abad342c92)

docs/reference/objects/api-keys.mdx
alexisintech: done [365ca33](https://github.com/clerk/javascript/pull/8276/commits/365ca336b04b994d1f5c741d2eed0227ab31b2e7)

docs/reference/objects/billing.mdx
SarahSoutoul, line 1: On to Billing! Pushed a very small [docs review - billing](https://github.com/clerk/clerk-docs/pull/3286/commits/181294d34f8d83a7d0a698029142717cdf4b93c9) making the capitalization of certain words consistent across the page and its sister one on `javascript` repo: [docs review - billing](https://github.com/clerk/javascript/pull/8276/commits/2e98fbf359a40ffe3df8ba3dd07e6c35030dde4e). 

Other than that, this is looking good. I have one small opinion on the wording - I don't love the use of "gets" (example below), and I think "retrieves" reads a bit better, if the goal is to avoid repeating "returns". But I won't die in this hill :joy:

docs/reference/objects/api-keys.mdx
SarahSoutoul: The extracted Typedoc method pages for `getAll()` and `create()` were showing duplicate returns content.

Screenshot 2026-05-12 at 2 34 16 pm

Pushed a fix [here](https://github.com/clerk/javascript/pull/8276/commits/98fd223e499fee62518ddfb869aafe82797560bc) on `javascript` repo where I update `.typedoc/extract-methods.mjs` so if the method description already includes the `@returns` text, we don't add it again. Also the `revoke()` method was still missing the returns bit so added that.

docs/reference/types/handle-o-auth-callback-params.mdx
SarahSoutoul: Np. I do think this is still relevant to the overall review, and useful to track here for the paper trail/check against later though.

docs/reference/objects/billing.mdx
alexisintech: I went with "get" because a lot of the methods are "getInvitations" "getResource" so "get" is a more colloquial term that's already being used in the API.
however, if you want to go more formal, I'm not set/tied to anything - feel free to make the updates!

SarahSoutoul: That's a good pt! After looking at other objects, I'm happy to leave it as "gets" because that approach is consistent across them with any methods starting with "get"

docs/reference/objects/session.mdx
SarahSoutoul, line 1: On to Session! Pushed a very small [docs review - session](https://github.com/clerk/clerk-docs/pull/3286/commits/2494cfd7f0440433f534c9dd071c46b0dabe9c8e) fixing a broken Client link and fixing the format of `SessionVerifyCreateParams` to stay consistent with how other params are displayed on this page (see screenshot) + its sister [PR on the `javascript` repo](https://github.com/clerk/javascript/pull/8276/commits/3fc800984f625743afb36eec11ed6cc3a639d624) adding changes to help with `SessionVerifyCreateParams` displaying correctly + added missing sentence for `prepareSecondFactorVerification`. 

Image

Other bits I noticed that I didn't want to mess with as it might be on purpose:

- A bit of mismatch with `prepareSecondFactorVerification` where the `SessionVerifyPrepareSecondFactorParams` params aren't displayed. It has the `@skipParametersSection` tho, so might be on purpose. 

  Image

- For the `SessionVerifyPrepareFirstFactorParams`, some missing comments (for `actionCompleteRedirectUrl` and `strategy` for Passkey for e.g.)

  Image
  
  Image

- The Agent Task link leading to 404, but normal until the agent task stuff is fixed.

docs/reference/objects/user.mdx
SarahSoutoul, line 1: For User, pushed a [docs review - user](https://github.com/clerk/javascript/pull/8276/commits/914e689e4b957007131d959d559ea520fe8e2e9f) on the `javascript` repo with a tiny update to the `legalAccepted` description that was missing. And here are the things I noticed:

- Some missing properties when comparing this PR with current docs: `verifiedExternalAccounts`, `unverifiedExternalAccounts`, `verifiedWeb3Wallets`, `hasVerifiedEmailAddress`, `hasVerifiedPhoneNumber` and `privateMetadata`.

  Image
  Image
  Image

- Bit of a weird formatting on this one

  Image

- The placement of "You only need..." paragraph looks a bit odd to me. Could we include it as part of the javascript generation so it sits below the first paragraph? 

Image

alexisintech: 1. those properties don't show because they are not properties on the user object if you check out the javascript repo! I'm not sure if they were there before and got removed, or what happened, but they're not there anymore.
2. let me look into this! great catch.
3. if we add the partial to the JSDoc comment (which generates that first paragraph), then it appears for `reload` for every single resource because every resource shares the `reload` method (`reload` exists on `ClerkResource` and every resource is an extension of `ClerkResource`). we would want this callout to only show for `User`. so the solution here would be: leave it where it is, remove it fully, or manually write the `reload()` section so that we can have more fine-grained control over the description/intro paragraph that gets generated.

alexisintech: fixed the second issue! [ab06e81](https://github.com/clerk/javascript/pull/8276/commits/ab06e81b83c0436d6f13f8cb4b2b7e130f3e2c79)

let me know what you're thinking for the third issue :) in my opinion, we can manually write the `reload` section 🤷‍♀️

SarahSoutoul: 1. Yeah they were there before, so wanted to flag. But that makes sense. 
2. Fix looks good! 
3. I'm fine with either leaving it as is or manually rewrite the `reload` section for this specific scenario

SarahSoutoul: @alexisintech One more thing I just saw - if. we got by the "gets" policy, this should be "Gets a list..". This is the only method being written manually, so that was prob why. 

Screenshot 2026-05-13 at 3 17 16 pm

docs/reference/objects/sign-in-future.mdx
SarahSoutoul, line 1: Okay have reviewed both `signInFuture` and `signUpFuture` together and pushed [docs review - signInFuture & signUpFuture](https://github.com/clerk/javascript/pull/8276/commits/d58f52f7693b4aab8da51d0729795badbe76ef07) on `javascript` repo adding some missing info in the descriptions of `SignInFutureEmailCodeSendParams` + linking `useUser` hook for `signUpFuture` to be consistent with `signInFuture`.

Other than this, here is what I noticed:

- There is some missing info in the intro of the `create()` method for both `signInFuture` and `signUpFuture` when comparing with current docs, which might be on purpose, but flagging in case.
  
  Image

**All of the below if for `SignUpFuture` only**

- `SignUpFutureResource` in the `update()` method intro gets autolinked, so when clicking on it, nothing happens. This is because we have a [`'sign-up-future-resource', '/docs/reference/objects/sign-up-future']` in  `LINK_REPLACEMENTS` causing this. So we may want to just change it to `SignUpFuture` instance in the paragraph to avoid this, as the link replacement is needed for other typedoc outputs. 

Image

- Think the current bullet points we have for `emailLinkVerification` are pretty useful to have - but are removed in this PR's output

Image

- This is more of a consistency thing but for `SignUpFuture`, there is this notation of "verifications.sendEmailCode()" so "object.method()" but `SignInFuture` displayed as just "method()". Either notation works but, we may want to make this consistent across the two objects.
  
  **How it is for `SignUpFuture`**
  
  Image
  
  **How it is for `SignInFuture`**
  
  Image

alexisintech: 1. I'm not seeing what information is missing, could you point it out?
2. I agree, we can just change it back: done here [ee0ceaf](https://github.com/clerk/javascript/pull/8276/commits/ee0ceaf5defd77fcc128fc7d799f54e85f08c75c)
3. I thought the values were self-explanatory from their names so I thought it was maybe extra visual noise, but if you prefer having that addition, its not harmful to add it back! done here [ee0ceaf](https://github.com/clerk/javascript/pull/8276/commits/ee0ceaf5defd77fcc128fc7d799f54e85f08c75c)
4. I've updated all the methods to use the dot notation! done here [ee0ceaf](https://github.com/clerk/javascript/pull/8276/commits/ee0ceaf5defd77fcc128fc7d799f54e85f08c75c)

SarahSoutoul: 1. These two things
    
  Screenshot 2026-05-14 at 10 16 29 am
  Screenshot 2026-05-14 at 10 16 22 am


Rest looks good!

alexisintech: hmm I'm confused, they are in both docs so what is missing? 🤔

SarahSoutoul: Ah wait, didn't realize the callout text was moved into the paragraph. Ignore me lol

docs/reference/objects/user.mdx
SarahSoutoul: @alexisintech just went ahead and fixed [here](https://github.com/clerk/clerk-docs/pull/3286/commits/896f2c41786ef3124c83df71e431f93e326f2879).

docs/reference/objects/organization.mdx
SarahSoutoul, line 1: For Org, it looked good on the all. Made a tiny commit [docs review - org](https://github.com/clerk/javascript/pull/8276/commits/481b67dbb0211a92ac310a2f6ffbe388279e5417) on `javascript` repo with syntax change.

And noticed one thing - in the current docs, `create()` and `get()` are documented for the Organization page:

Image

Image

But not in this PR - looking at the `javascript` PR, they def exist and should be documented imo, but might be wrong. They are static methods in the `Organization.ts` file. 

Image

alexisintech: I think those are internal - they don't exist on the `OrganizationResource` ([source](https://github.com/clerk/javascript/blob/aa/generate-object-docs/packages/shared/src/types/organization.ts)). when you access `Organization` through hooks, those aren't available on it because the returned `Organization` is of type `OrganizationResource`.

docs/reference/objects/user.mdx
alexisintech: just seeing this! thanks sarah

docs/reference/objects/organization.mdx
SarahSoutoul: Okay, yeah I wasn't sure. We can close then!

docs/reference/objects/clerk.mdx
SarahSoutoul, line 1: Have reviewed both `Clerk` and `Client` objects. No notes for `Client` - looks good. For `Clerk`, have pushed this mall [docs review - clerk](https://github.com/clerk/javascript/pull/8276/commits/87215e7204597f20f7ef8bbea1c27be76c6cad5f) on `javascript` repo, doing the following:

- Linking correctly `APIKeysNamespace` as clicking on it would lead to 404. 
- Changed `afterSignInUrl` -> `afterSignUpUrl` for `buildAfterSignUpUrl` method description. 

Noticed two things:

- Should this link be external as it's directing to the Frontend API? 

  Image

- This links to 404, which makes sense as we don't have an `OAuthApplication` type created but this [PR](https://github.com/clerk/clerk-docs/pull/3315/) does. So we may want to bring some of those changes in here or we can wait for the PR to go and merge in the changes. 

  Image

SarahSoutoul:
@alexisintech okay have now reviewed everything in this PR + pushed my changes! Last comment/ suggestion is for `Clerk` object. 

@manovotny you're good to give it a look now!

docs/reference/objects/clerk.mdx
alexisintech: 1. yes I can update that to be external! done [fe17dca](https://github.com/clerk/javascript/pull/8276/commits/fe17dcafe0aaa9bd9aa5203a36f341a5be0ce1b7)
2. the best case scenario is we let that ship and then merge it in. do we think it's gonna ship soon? have you heard anything from rob/mind checking in with him?

SarahSoutoul: Yeah already checked in with him. Think pretty soon, but no specific ETA.

docs/reference/objects/session.mdx
```diff
@@ -413,89 +188,104 @@ type SessionVerifyPrepareFirstFactorParams =
...
-### `prepareSecondFactorVerification()`
-
-Initiates the [second factor verification](!second-factor-verification) process. This is a required step to complete a reverification flow when using a preparable factor. Returns a [`SessionVerification`](/docs/reference/types/session-verification) instance with its status and supported factors.
-
-```typescript
-function prepareSecondFactorVerification(
-  params: SessionVerifyPrepareSecondFactorParams,
-): Promise<SessionVerificationResource>
-```
+<Typedoc src="shared/session-resource/methods/prepare-second-factor-verification" />
```
SarahSoutoul, line 267: @alexisintech One more thing - shouldn't the params be `SessionVerifyPrepareSecondFactorParams` here?

Image

docs/reference/objects/user.mdx
SarahSoutoul: @alexisintech for number 3, what do we want to do?

docs/reference/objects/session.mdx
alexisintech: yes great catch - unfortunately since the types are equal, it resolves to the final type. I was working on a fix for it but it was getting convoluted just to fix one small edge case (and I didn't want it to start breaking all the other generated param types). if someone was viewing the types in the code, they'd see these are equal types - its only the function signature in the docs. if you think it's worth tackling, I can see if the new codex stuff has a better solution or maybe you or michael can find a decently simple solution for it (without breaking a bunch of other things)

the other solution is to change the heading name (since its manually written) and then just remove that first line in the code block

SarahSoutoul: Yeah I do think it's quite a niche issue / small edge case. So I like this solution:

`the other solution is to change the heading name (since its manually written) and then just remove that first line in the code block`

Think it's worth implementing! @alexisintech

alexisintech: done ! [69f5bb0](https://github.com/clerk/clerk-docs/pull/3286/commits/69f5bb03aed90ba8ae666747dc3c1598ec165300)

manovotny:
Pushed some changes directly in fb691ec89.

- Fixed `load()` signature in `clerk.mdx` — was `function load(...) => Promise<void>`, which is invalid TS; now `function load(...): Promise<void>`. Dropped the `{{ prettier: false }}` directive since it's no longer needed.
- Updated the inline `SessionVerifyPrepareFirstFactorParams` snippet in `session.mdx` to match source: `PasskeyConfig` → `PassKeyConfig` (renamed at both the union site and the inline type definition), and the enterprise branch is now `Omit<EnterpriseSSOConfig, 'actionCompleteRedirectUrl'>` to mirror `packages/shared/src/types/session.ts:539`.
- Swapped `<Typedoc src=".../on" />` and `<Typedoc src=".../off" />` in `clerk.mdx` so the order is alphabetical, matching the rest of the file.
- Restored the intro sentence "Information about the user that's publicly available." to `public-user-data.mdx`, since `handle-o-auth-callback-params.mdx` sets that as the precedent among the newly-added type pages.

LMK if these look good to you. Leaveing a few comments for things I felt less confident on.

docs/reference/objects/clerk.mdx
```diff
@@ -834,474 +122,63 @@ The method handles the verification outcome as follows:
...
 
-A method used to set the current session and/or Organization for the client. Accepts a [`SetActiveParams`](/docs/reference/types/set-active-params) object.
+<Typedoc src="shared/clerk/methods/redirect-with-auth" />
 
-```typescript
-function setActive(params: SetActiveParams): Promise<void>
-```
+<Typedoc src="shared/clerk/methods/set-active" />
 
 #### Using the `navigate()` parameter
```
manovotny, line 186: This whole section reads as part of `setActive()` only because it sits under the typedoc-rendered h3 heading from line 181. Two similar pockets live higher up — lines 99-125 right after `handle-email-link-verification`, and the lone line 131 after `handle-redirect-callback`.

It renders correctly today, but the structure is fragile: if typedoc ever appends a footer/separator below the method body, these blocks drift out of context. 

Two ways to prevent future hiccups:

1. Promote each block into a JSDoc continuation (e.g. `@remarks`) on the JS side so it ships inside the method body that typedoc generates.
2. Give each block its own explicit subheading so the hierarchy is declared rather than implied by placement — the `setActive` one already has `#### Using the navigate() parameter`, but the other two don't.

Either works; 2 is the smaller change if 1 is too disruptive. Or feel free to dismiss if you feel like it's not a problem.

docs/reference/objects/api-keys.mdx
```diff
@@ -214,109 +214,14 @@ The `APIKeys` object provides methods for managing API keys that allow your appl
 
 The `APIKeys` object is of type [`APIKeyResource`](/docs/reference/types/api-key-resource).
```
manovotny: Verified against `packages/shared/src/types/clerk.ts:1166` — `apiKeys: APIKeysNamespace`. The `APIKeys` object is `APIKeysNamespace`; `APIKeyResource` is what `create()` / `getAll()` return.

The Properties section below this line is `APIKeyResource` properties, and the Methods section mixes `APIKeysNamespace` methods (`create`, `getAll`, `revoke` — called via `clerk.apiKeys.create(...)`) with `APIKeyResource` instance methods (`reload` — called on a single key like `apiKey.reload()`). Readers can't tell which kind of call goes with which method.

Pre-existed before this migration, but the typedoc switch makes it more visible. Maybe rephrase to something like:

```suggestion
Methods on the `APIKeys` object return [`APIKeyResource`](/docs/reference/types/api-key-resource) instances. The properties below describe `APIKeyResource`.
```

A fuller fix would split namespace methods and resource methods into distinct sections — happy to leave that for a follow-up.

docs/reference/objects/user.mdx
```diff
@@ -538,205 +93,33 @@ function getOrganizationMemberships(
...
 #### `UpdateUserParams`
 
-<Properties>
-  - `username?`
-  - `string`
-
-  The user's [username](/docs/guides/configure/auth-strategies/sign-up-sign-in-options#username). Only supported if username is enabled in the instance settings.
-
-  ---
+{/* DOCS NOTE: This is a complex type that `extract-methods.mjs` cannot parse through, so we manually add it here. */}
```
manovotny, line 123: The DOCS NOTE is honest, but a hand-maintained block here is a drift risk — `UpdateUserParams` in source can grow new fields and this table won't catch them. Same shape applies to `getOrganizationMemberships` at lines 70-94 (`GetUserOrganizationMembershipParams = ClerkPaginationParams`).

Worth considering whether `extract-methods.mjs` can be taught to parse these. The two cases have very different complexity — sketch below for whoever picks this up next.

<details>
<summary>Starter plan for fixing this in <code>extract-methods.mjs</code></summary>

### Case A — `GetUserOrganizationMembershipParams = ClerkPaginationParams` (likely easy)

This is a bare alias `type Foo = Bar` with no `typeArguments`. The existing `resolveDeclarationWithObjectMembers` chain at `.typedoc/extract-methods.mjs:953-958` should fall through to the underlying type, but probably doesn't because TypeDoc doesn't emit `typeArguments` for the unspecified generic default on `ClerkPaginationParams<T = object>`. The fix is to descend into the intersection arms of the alias when `typeArguments.length === 0`. Worth testing whether removing the manual block at user.mdx:70-94 "just works" first — it might already.

### Case B — `UpdateUserParams = Partial<SnakeToCamel<Pick<UserJSON, ...>>>` (focused half-day refactor)

The blocker is `SnakeToCamel<T>` — a template-literal mapped type with an `as` rename clause that TypeDoc represents opaquely. The pattern repeats elsewhere (`UpdatePasskeyParams` is the same shape, and there are `DeepSnakeToCamel` / `DeepCamelToSnake` siblings), so a real solution scales.

Resolver steps:

1. In `resolveDeclarationWithObjectMembers`, recognize the wrapper utilities by name on `ReferenceType` — `Partial`, `Pick`, `SnakeToCamel` / `DeepSnakeToCamel`, `DeepCamelToSnake`.
2. Resolve `Pick<UserJSON, K1 | K2 | ...>` to a property subset (TypeDoc resolves `Pick` natively in some cases; otherwise enumerate the literal type args and filter `UserJSON.children`).
3. Walk the resolved children and synthesize new property reflections with names run through a small `snakeToCamel()` JS helper, copying `comment` / `type` from the source property so JSDoc carries over.
4. Apply `Partial<>` by setting `optional: true` on every synthesized child (or wrap in `OptionalType`).

Sources to look at:

- `SnakeToCamel` definition: `packages/shared/src/types/utils.ts:9-19`
- `UpdateUserParams` chain: `packages/shared/src/types/user.ts:442-453`
- Existing resolver to extend: `.typedoc/extract-methods.mjs:886-1026`

Estimated effort: ~100-200 lines of mjs, plus a recursive helper for the `Deep*` variants.

</details>

docs/reference/objects/session.mdx
```diff
@@ -56,129 +56,13 @@ All sessions that are **expired**, **removed**, **replaced**, **ended** or **aba
...
-  - [`User`](/docs/reference/objects/user) | null
-
-  The user associated with the session, or `null` if none.
-</Properties>
+<Typedoc src="shared/session-resource/properties" />
 
 ## Methods
 
-### `attemptFirstFactorVerification()`
+{/* DOCS NOTE: Some of these parameters sections are manually written because as you will see, they are complex types that are difficult to generate with Typedoc. For example, many of these are union types where each of the union members share a property like `code`. So if we were to generate a single parameters table for all of the values, `code` would be duplicated for each of the union members, and the table wouldn't really make sense. */}
```
manovotny, line 64: Same maintenance/drift concern as the user.mdx DOCS NOTE — five manually-written param sections in this file (`SessionVerifyAttemptFirstFactorParams`, `SessionVerifyAttemptSecondFactorParams`, `CheckAuthorizationParams`, `SessionVerifyPrepareFirstFactorParams`, `PhoneCodeSecondFactorConfig`) will silently drift if the JS source unions add or change arms.

Worth flagging that the technical case here is harder than `user.mdx`'s `UpdateUserParams`. The rationale in this DOCS NOTE — "if we were to generate a single parameters table for all of the values, `code` would be duplicated" — points at a genuinely harder problem: discriminated unions can't be cleanly flattened, because flattening loses the strategy-to-fields correspondence.

<details>
<summary>Why this is harder than the <code>SnakeToCamel</code> case</summary>

The existing union handler at `.typedoc/extract-methods.mjs:987-1021` already merges union members into a single `__unionMerged` holder and filters `never` discriminators. It works for unions like `CheckAuthorizationParams` where each arm uses `never` markers (`permission?: never; feature?: never; ...`) to discriminate — the `never` filter strips the discriminator noise and the merged table is usable.

It doesn't work for unions like `SessionVerifyAttemptFirstFactorParams`:

```ts
type EmailCodeAttempt = { strategy: 'email_code'; code: string }
type PhoneCodeAttempt = { strategy: 'phone_code'; code: string }
type PasswordAttempt  = { strategy: 'password';   password: string }
type PasskeyAttempt   = { strategy: 'passkey';    publicKeyCredential: ... }
```

Flatten that and you get four rows (`strategy`, `code`, `password`, `publicKeyCredential`) but lose the rule that `code` only applies when `strategy ∈ { 'email_code', 'phone_code' }`. The author's workaround — show the raw TS union in a code block — sidesteps the issue rather than solving it.

### Possible path forward

Teach the union resolver to detect "discriminated unions" (heuristic: all arms share a property whose type is a unique string literal across arms) and render each arm as its own labeled subsection rather than merging. Sketch:

1. In the `union` branch of `resolveDeclarationWithObjectMembers`, scan members for a shared literal-string discriminator property.
2. If found, return a synthetic holder that the table renderer treats specially — emit `<h5>EmailCodeAttempt</h5>` + table per arm, instead of a single merged table.
3. Keep the existing merge path for non-discriminated unions (e.g. `CheckAuthorizationParams` with `never` markers).

### Caveats

- This is a meaningful page-structure change, not a pure parser fix. Worth a design pass on what the rendered output should look like before coding it.
- Some of the manual blocks here (e.g. the inline `WithReverification<...>` definition for `CheckAuthorizationParams`) carry JSDoc that lives only in the docs. Those JSDoc lines would need to move to source before this could be automated.
- Effort estimate: probably 1-2 days for the resolver + renderer, plus the design conversation. Larger than the `SnakeToCamel` case.

</details>

I'm not meaning to make a big deal out of these. I see the work you've done solving a lot of problems, and I _want_ to remove _all_ the manual maintenance here, but it just might not be possible or worth it.

alexisintech:
waiting on [docs(shared): Fix Typedoc links](https://github.com/clerk/javascript/pull/8746#top) to get merged
which will fix these build errors 🙏

docs/reference/objects/clerk.mdx
alexisintech: I'm going to dismiss this as there isn't a case where Typedoc would append anything below the method body. The extract-methods script outputs the same structure for all of the methods:

```mdx
## `methodName()`

Description from JSDoc comment.

### `Parameters` 

Parameter table.
```

docs/reference/objects/user.mdx
alexisintech: I'm opening to trying this out in a followup! I've created a ticket to track https://linear.app/clerk/issue/DOCS-11825/generate-updateuserparams

docs/reference/objects/session.mdx
alexisintech: I've made a followup ticket! https://linear.app/clerk/issue/DOCS-11826/typedoc-handle-complex-unions

manovotny:
Pushed a small follow-up in b3803bbaa.

- Tightened the `APIKeys` properties intro. The wording I'd suggested earlier ("Methods on the `APIKeys` object return `APIKeyResource` instances") was slightly off — `getAll()` returns a `ClerkPaginatedResponse<APIKeyResource>`, not a bare instance, and `reload()` is a method on `APIKeyResource` itself. Reworded to "The `APIKeys` methods create, list, and revoke `APIKeyResource` objects," which maps cleanly to the three namespace methods.

LMK if this looks good to you. Posting one more question.

docs/reference/objects/clerk.mdx
manovotny: Makes sense — if extract-methods emits a fixed shape and nothing renders below the method body, the drift I was picturing can't happen. Happy to dismiss. Thanks for laying out the structure.

docs/reference/objects/user.mdx
manovotny: Thanks for ticketing it. The Case A / Case B split above should give whoever picks it up a head start — Case A (the `ClerkPaginationParams` alias) is the cheap win worth trying first.

docs/reference/objects/session.mdx
manovotny: Appreciate the ticket — and no worries at all, agreed some of these may not be worth automating. Discriminated-union detection is the crux if someone takes it on; the caveats above flag what needs a design call first.

clerk-typedoc/shared/clerk/properties.mdx
```diff
@@ -0,0 +1,21 @@
...
+| <a id="loaded"></a> `loaded`                       | `boolean`                                                                                            | Indicates whether the `Clerk` object is ready for use. Set to `false` when the `status` is `"loading"`. Set to `true` when the `status` is `"ready"` or `"degraded"`.                                                                                                                                                                    |
+| <a id="oauthapplication"></a> `oauthApplication`   | [`OAuthApplicationNamespace`](../o-auth-application-namespace.mdx)                                   | OAuth application helpers (e.g. consent metadata for custom consent UIs).                                                                                                                                                                                                                                                                |
```
manovotny, line 12: The `oauthApplication` row renders as `[`OAuthApplicationNamespace`](../o-auth-application-namespace.mdx)` — a raw relative `.mdx` path that leaks into all 11 SDK-scoped rendered `clerk.mdx` pages (its table siblings `organization`/`session`/`user` resolve to real `/docs/...` routes via `SDKLink`).

Root cause is upstream: `o-auth-application-namespace` isn't in `LINK_REPLACEMENTS` in `clerk/javascript`'s `.typedoc/custom-plugin.mjs` (siblings like `o-auth-consent-info` and `organization-resource` are). 

Heads up that `build:tsx` doesn't catch this — the link validator skips relative `.mdx` links inside embedded Typedoc tables, so the clean build isn't proof here.

Fix would be adding the mapping there + likely a routed `docs/reference/types/oauth-application-namespace.mdx` page (parallel to the `image-resource` page this PR adds), then regenerate. Worth a ticket like the others?

alexisintech: yes we can update this as a followup based on this PR: https://github.com/clerk/clerk-docs/pull/3315/changes#diff-4758545ec3ecda5309aeef6572743a8c3613c42b0f0e40128f2a93a6eb4c9bbc
if we update the link now, there's nothing to point it to because the oauth stuff doesn't exist yet.

I've created a ticket https://linear.app/clerk/issue/DOCS-11827/typedoc-oauthapplicationnamespace-needs-a-replacement

PR #3265, opened 2026-04-02
Add custom flows to mobile sidebar

alexisintech:
### 🔎 Previews:

-

### What does this solve? What changed?

iOS/Android SDKs offer three prebuilt views, leaving users to rely on mostly custom flows to build apps. This PR serves to add the appropriate custom flows to the mobile sidebar.

- Adds final missing ios/android examples
- Updates copy for docs so they are tailored for ios/android
- Scopes docs that are not for ios/android

### Deadline

- No rush

### Other resources

- Ask for sign-in-or-up examples: https://clerkinc.slack.com/archives/C065VEK864A/p1775155462792539
- Question about waitlist for native: https://clerkinc.slack.com/archives/C065VEK864A/p1775155506450539
- Question about `AuthView` including session task handling: https://clerkinc.slack.com/archives/C065VEK864A/p1775158294090309
- Clarity regarding bot protection with native apps: https://clerkinc.slack.com/archives/C065VEK864A/p1775596951998429
- Ask for manage-mfa examples: https://clerkinc.slack.com/archives/C065VEK864A/p1775597470502319

PR #3289, opened 2026-04-14
Fix <If notSdk> false positives and skip partial scope checks

manovotny:
### 🔎 Previews:

- N/A (build script changes only)

The `<If>` component validator had two sources of false positives that caused us to disable it entirely with `⚠️ TEMPORARILY DISABLED` warnings:

**1. `notSdk` false positives** — The validator treated `notSdk` props identically to `sdk` props. When a page used `<If notSdk={["ios", "android"]}>` but wasn't scoped for ios/android, the validator flagged it. But `notSdk` means "show for everything EXCEPT these" — the excluded SDKs don't need to be in scope. Fixed by having `extractSDKsFromIfComponent` return `undefined` for `notSdk` (SDK names are still validated, just not scope-checked).

**2. Partial false positives** — Shared partials like `_partials/hooks/use-user.mdx` contain `<If sdk="react">`, `<If sdk="nextjs">`, etc. for multiple SDKs. When included on a page scoped to only one SDK, the validator flagged the others. Fixed by marking embedded partial nodes with `data.fromPartial` in `checkPartials.ts` and skipping scope checks for those subtrees in the validator.

**What's NOT re-enabled yet:** The `safeFail` calls remain as `console.warn` because ~79 remaining warnings are ios/android manifest issues that #3265 will resolve. After #3265 merges, only 4 files / 8 warnings remain (tested by merging #3265 locally):

| Issue | File | Count |
|-------|------|-------|
| Frontmatter missing ios/android | `embedded-email-links.mdx` | 2 |
| Manifest missing ios/android | `embedded-email-links.mdx` | 2 |
| Manifest missing ios/android | `sign-up-sign-in-options` | 2 |
| Manifest missing android | `social-connections/google` | 1 |
| Manifest missing android | `legacy/passkeys` | 1 |

Updated the TODO comments to reference #3265.

**Out of scope / follow-up:**
- **Dead `<If>` detection in partials** — This PR skips scope checks for `<If>` inside partials (to eliminate false positives), but doesn't yet detect truly dead content in partials (e.g., `<If sdk="X">` in a partial where no including page is scoped for X). That requires an aggregate check across all including pages — a separate feature.

**Changes:**
- `validateIfComponents.ts` — Return `undefined` for `notSdk` props (skip scope check), import `SKIP` and skip `fromPartial` subtrees
- `checkPartials.ts` — Mark embedded partial root nodes with `data.fromPartial`
- `build-docs.test.ts` — Add regression test for `notSdk` with out-of-scope SDKs, update TODO comments to reference #3265

### Deadline

- No rush

### Other resources

- #3265 — Add custom flows to mobile sidebar (resolves 79 of the 81 remaining warnings)

### DIFF
--- a/scripts/build-docs.test.ts
+++ b/scripts/build-docs.test.ts
@@ -1241,9 +1241,7 @@ Testing with a simple page.`,
     expect(output).toContain(`warning sdk \"astro\" in <If /> is not a valid SDK`)
   })
 
-  // TODO: Temporarily disabled due to large-scale docs/SDK changes (Core 3, native mobile sidebar, and Development SDK-specificity.
-  // Change back when `safeFail` is restored in validateIfComponents.ts
-  console.warn('⚠️  TEMPORARILY DISABLED: <If> SDK not in frontmatter test skipped')
+  // TODO: Re-enable after clerk/clerk-docs#3265 (mobile custom flows manifest) merges and safeFail is restored in validateIfComponents.ts
   test.skip('<If> SDK not in frontmatter fails the build', async () => {
     const { tempDir } = await createTempFiles([
       {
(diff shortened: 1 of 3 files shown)


manovotny:
Consolidated with #3290.

PR #3292, opened 2026-04-15
(billing): clarify Stripe account linking requirements

mauricioabreu:
### 🔎 Previews:

https://clerk-docs-git-mauricio-antunes-update-stripe-account-wi-ae58be.clerkstage.dev/

Customers were confused by the statement that existing Stripe accounts must not be linked to another platform, since in practice some accounts can still be connected to multiple apps.

This change clarifies that the restriction only applies to accounts that are controlled by another platform (e.g created via a platform's Connect setup), which cannot be connected via OAuth with read/write access

### Other resources
Slack thread: https://clerkinc.slack.com/archives/C08QHV76VA8/p1776201504373949

### DIFF
--- a/docs/guides/billing/overview.mdx
+++ b/docs/guides/billing/overview.mdx
@@ -32,7 +32,8 @@ Clerk Billing allows your customers to purchase recurring Subscriptions to your 
 ### Can I use an existing Stripe account with Clerk Billing?
 
 Yes, you can use an existing Stripe account, as long as it isn't controlled by another platform.
-Accounts created under a platform's Connect setup must be disconnected from that platform before they can be linked to Clerk.
+
+Disconnect accounts created under a platform's Stripe Connect setup from that platform before linking them to Clerk.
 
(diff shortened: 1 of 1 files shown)


manovotny:
Pushed some changes directly in 6660f4b72.

- **Separated paragraphs**: Added a blank line between the general answer and the Stripe Connect caveat so they render as distinct paragraphs instead of running together.
- **Clarified "Stripe Connect"**: Changed "a platform's Connect setup" to "a platform's Stripe Connect setup" so readers unfamiliar with Stripe's product naming aren't left guessing.
- **Active voice**: Rewrote "Accounts created under a platform's Connect setup must be disconnected from that platform before they can be linked to Clerk" → "Disconnect accounts created under a platform's Stripe Connect setup from that platform before linking them to Clerk."

Give them a look and let me know — happy to revise or revert if they're off.

manovotny (approved):

mauricioabreu:
@manovotny much better! I appreciate the improvements 🙏🏼

SarahSoutoul (approved):

PR #3296, opened 2026-04-16
Simplify production guide setup

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-simplify-mobile-setup/guides/development/deployment/production (need to test for iOS, Android, and any other non mobile SDK) 

Following the release of this [PR](https://github.com/clerk/clerk/pull/2402) fixing the TOC headings, I wanted to simplify the current setup we had for the duplicated production guide (web vs mobile). Since there are very few differences between web and mobile for the deployment process, it didn't make much sense to have them duplicated. We did this in the past because when wrapped into one using the <If> components, it would cause issues with the TOC. 

However, now that the TOC issues are fixed, I think it's better to keep it as one and use the <If> components for the parts that are proper to ios and android. 

### Deadline

No rush.


SarahSoutoul:
@manovotny have spotted a bug with the TOC while working on this PR. Hopefully, the video talks for itself but there seems to be a weird behaviour happening where the headings don't appear for Android, but when switching the order in the <If>, then they do appear but are gone from iOS. 

https://github.com/user-attachments/assets/f28a7114-e5ba-4c9f-bbf5-14727a796a75

I did a test with `nextjs` and `react`, and the same thing happens. 

https://github.com/user-attachments/assets/56685001-cefd-4e65-99d8-561f98dd2873

For it to work properly, I need to have it as such:

```
<If sdk={["react", "nextjs"]}>
  <If sdk="nextjs">
    ## Testing TOC 

    I AM A HEADING NEXTJS
  </If>

  <If sdk="react">
    ## Testing TOC 

    I AM A HEADING REACT
  </If>
</If>
```

instead of

```
<If sdk={["react", "nextjs"]}>
   ## Testing TOC 

  <If sdk="nextjs">
    I AM A HEADING NEXTJS
  </If>

  <If sdk="react">
    I AM A HEADING REACT
  </If>
</If>
```

SarahSoutoul:
@manovotny And another issue... on that same page, with the `<If notSdk=..>` (see code below). That heading appears for Next.js, but not for the rest of the SDKs as it should.


Screenshot 2026-04-16 at 10 54 53 am


https://github.com/user-attachments/assets/6e684465-5831-4b87-a514-3365e031efb5


From what I can tell, both of these issues happen when several SDKs are passed to the `sdk` prop or `notSdk` prop.

manovotny:
Pushed some changes directly in 27f764826.

- Added static redirects for the old native-mobile production URLs (`/docs/reference/native-mobile/production`, plus the `/docs/ios/` and `/docs/android/` SDK-prefixed variants) → `/docs/guides/development/deployment/production`, so old bookmarks and search results don't 404.
- Swapped "Make sure" → "Ensure" on the new iOS/Android "app is registered" lines to match the styleguide.

docs/guides/development/deployment/production.mdx
```diff
@@ -72,36 +72,77 @@ When you set a root domain for your production deployment, Clerk's authenticatio
...
+
+    Ensure your iOS app exists in the production instance under [**Native Applications**](https://dashboard.clerk.com/~/native-applications).
+    You will need your **App ID Prefix** and **Bundle ID**.
+
+    ## Update associated domains
+
+    In Xcode, update **Associated Domains** to use your production [Frontend API URL](!frontend-api-url):
+
+    1. Under **Signing & Capabilities**, add or update the entry to `webcredentials:{YOUR_FRONTEND_API_URL}`.
+    1. Use the production instance's Frontend API URL.
```
manovotny: Pre-existing copy, but since this PR is a simplification pass...

Does step 2 add anything step 1 doesn't already convey? The `{YOUR_FRONTEND_API_URL}` placeholder in step 1 already signals that the production Frontend API URL is what goes there.

Could collapse to a single step:

```suggestion
    1. Under **Signing & Capabilities**, add or update the entry to `webcredentials:{YOUR_FRONTEND_API_URL}`, using the production instance's Frontend API URL.
```

manovotny (approved):

SarahSoutoul:
Thanks for the approval @manovotny! I want to hold off merging this, mainly for the issues raised above.

manovotny:
@SarahSoutoul yep, digging into those this afternoon. 👍

redirects/static/docs.json
```diff
@@ -4386,5 +4386,17 @@
...
+    "source": "/docs/reference/native-mobile/production",
+    "destination": "/docs/guides/development/deployment/production"
+  },
+  {
+    "source": "/docs/ios/reference/native-mobile/production",
+    "destination": "/docs/guides/development/deployment/production"
+  },
+  {
+    "source": "/docs/android/reference/native-mobile/production",
+    "destination": "/docs/guides/development/deployment/production"
```
alexisintech, line 4404: I'm curious as to why all three are necessary 💭

alexisintech (approved):
one question but its not blocking! this is a great optimization - awesome job remembering the context behind this, sarah 😸💖

redirects/static/docs.json
manovotny: The deleted file had `sdk: ios, android` in its frontmatter, so three URLs were routable: the base path plus the `/docs/ios/` and `/docs/android/` prefixed variants. Each needs its own redirect to avoid 404s. At least I think that's right. 

All of these are currently accessible:

* https://clerk.com/docs/reference/native-mobile/production
* https://clerk.com/docs/ios/reference/native-mobile/production
* https://clerk.com/docs/android/reference/native-mobile/production

PR #3303, opened 2026-04-17
feat: Added remark-gfm to support checklists in docs

royanger:
### 🔎 Previews:

- no visible changes in this PR to link to

- `clerk` PR - https://github.com/clerk/clerk/pull/2361
- This PR enables checklists in docs

Markdown example:
```markdown
1. Once you make the change to your domain, you will need to update the following:
   - [ ] Update DNS records.
   - [ ] Generate new SSL certificates.
   - [ ] [Update your Publishable Key](#update-your-publishable-key).
   - [ ] If using social connections, update the settings with your social connections so that the redirect URL they are using is correct.
   - [ ] If using JWT templates, update JWT issuer and JWKS Endpoint in external JWT SSO services.
```
Example from [Changing Domains doc](https://clerk.com/docs/guides/development/deployment/changing-domains) where I made the above change in local dev, for reference

image

image

### Deadline

- n/a, and just a proposal is the Docs team wants to enable this


manovotny:
Pushed some changes directly in a457ff012.

- Added remark-gfm to `scripts/lib/partials.ts`. `checkPartials` embeds `partial.node` directly, and that node was still being parsed without GFM — so a checklist inside a partial would get stripped back to plain text at embed time. This makes the feature work for partials too.
- Added two regression tests under a new `GFM Support` describe block: one for a checklist in a doc, one for a checklist inside an `<Include />`'d partial.

Drop either if the scope feels wider than you wanted.

manovotny (approved):

royanger:
> Pushed some changes directly in [a457ff0](https://github.com/clerk/clerk-docs/commit/a457ff01210aa230701464364e063fb80944df12).
> 
> Drop either if the scope feels wider than you wanted.

This look great. I totally missed partials. Thanks !

PR #3304, opened 2026-04-17
Align optional prop notation in docs

SarahSoutoul:

Found this old Linear ticket, which I decided to look into as I've been having this exact thought for a few months. Prop descriptions don't need "optional" because they should have a `? `marker clearly showing they're optional. This PR removes all instances of "optional" in prop descriptions. 

Through this, I also noticed the following:
- Some component reference pages said “All props are optional” or “all of which are optional”, but not every prop name in the list reflected that convention. So, I updated optional prop lists so prop names consistently use a trailing `?` where appropriate.

I will look at the `javascript` repo next as some of the Typedoc outputs contain this too. 

### Deadline

No rush. Docs improvement only. 

### Other resources

Linear: https://linear.app/clerk/issue/DOCS-11113/chore-prop-descriptions-dont-need-optional-because-they-should-have


manovotny:
Pushed some changes directly in ba829f40.

- Restored the dropped `fallback?` prop name in `organization-profile.mdx` — the previous diff left a blank `-` row, which would have rendered an empty row in the `<Properties>` block.
- Added `?` to `localization` in `clerk-provider/properties.mdx` and to `appearance` + `localization` in `update-clerk-options-properties.mdx`, where descriptions had been de-"Optionalized" but the prop names hadn't been updated. Verified against the source types in the `javascript` repo.
- Added `?` to `supportEmail` in `clerk-provider/properties.mdx` and dropped the leading "Optional" from the `supportEmail` descriptions in both that partial and `javascript/clerk-options.mdx`.
- Dropped "Optional" from five prop descriptions in `objects/clerk.mdx` (`addListener` options, two `buildWaitlistUrl`/sign-up URL `options?.params`, waitlist `options?.initialValues`, and `setNavigate` options) to match the rest of the pass.

Also noticed a pre-existing broken markdown link in `organization-switcher.mdx` — `` `<CreateOrganization />`][createorg-ref] `` was missing the opening `[`, so the reference link wasn't rendering. Fixed in 4dae9910.

LMK if you disagree with any of the edits.

manovotny (approved):
Looks so much better. 🫶

PR #3308, opened 2026-04-20
Emphasize email requirement in Directory docs

kalafut:
We continue to regularly see SCIM provisioning failures due to lack of an email field in the payload. This doc update calls out the requirement explicitly.

### DIFF
--- a/docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
+++ b/docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
@@ -40,7 +40,7 @@ Directory Sync is configured per enterprise connection. To set it up:
 Once Directory Sync is enabled in Clerk, configure your IdP to point to Clerk's SCIM endpoint.
 
 > [!IMPORTANT]
-> Clerk requires an `email` attribute to be sent for every user. Even if your `userName` attribute contains an email address, your IdP must also be configured to send the email in the SCIM `email` attribute — Clerk will not fall back to `userName` for email. Users whose SCIM payloads omit `email` will fail to provision.
+> Clerk requires an `email` attribute to be sent for every user. Even if your `userName` attribute contains an email address, you must also configure your IdP to send the email in the SCIM `email` attribute — Clerk will not fall back to `userName` for email. Users whose SCIM payloads omit `email` will fail to provision.
 
 ### Okta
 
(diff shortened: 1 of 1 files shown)


iagodahlem (approved):

manovotny:
Pushed a couple of small tweaks directly in 631a3c8.

- Reworded the callout to active voice: "you must also configure your IdP to send the email…"
- Dropped the "being" gerund from the Okta step: "mapped and sent to Clerk"

Left a separate comment on the `email` vs `emails` SCIM field name — want your read on that one.

docs/guides/configure/auth-strategies/enterprise-connections/directory-sync.mdx
```diff
@@ -39,6 +39,9 @@ Directory Sync is configured per enterprise connection. To set it up:
 
 Once Directory Sync is enabled in Clerk, configure your IdP to point to Clerk's SCIM endpoint.
 
+> [!IMPORTANT]
+> Clerk requires an `email` attribute to be sent for every user. Even if your `userName` attribute contains an email address, you must also configure your IdP to send the email in the SCIM `email` attribute — Clerk will not fall back to `userName` for email. Users whose SCIM payloads omit `email` will fail to provision.
```
manovotny: I am not an SCIM expert, so I'm triple checking...

[RFC 7643](https://www.rfc-editor.org/rfc/rfc7643) defines the SCIM user email field as `emails` (array) with nested `.value`, and the existing attribute table on line 75 uses `email.value`. This callout and the new Okta step on line 54 both describe the payload field as a top-level singular `email`.

Is Clerk's receiver actually accepting that, or does the failing IdP config need to populate `emails[].value`?

If the Clerk Dashboard label for the mapping is literally `email` (singular), it would help to distinguish the UI label from the SCIM payload field here and align the terminology with the table below so readers don't map the wrong SCIM path.

kalafut: Good callout. The data structure is an array, but IdPs sometimes mask that aspect. Does my [latest update](https://github.com/clerk/clerk-docs/pull/3308/commits/8fabc966301050279213297396b6d90161ea0a97) at least avoid giving the wrong impression? I don't really want to spend a lot of words on this. The most important takeaway should be that we need an email for the user.

manovotny: Yep, that reads right. Callout now points at the correct SCIM field, and keeping the Okta step at `email` / `user.email` matches Okta's UI labels — no reason to over-explain it.

manovotny (approved):

SarahSoutoul (approved):

PR #3322, opened 2026-04-27
Fix Account Portal direct links redirect docs

f-westergren:
### 🔎 Previews:
- https://clerk.com/docs/pr/folke-fix-account-portal-direct-links/guides/account-portal/direct-links

- Updates the Account Portal direct links guide so fallback redirect configuration points users to the Clerk Dashboard instead of SDK/component props.
- Previously, the fallback redirects section told users to configure sign-in/sign-up redirects with environment variables and sign-out/organization redirects through component props like `<ClerkProvider>`, `<OrganizationProfile />`, and `<OrganizationSwitcher />`. That guidance does not match where Account Portal fallback redirects are configured.
- The updated docs now direct users to the **Account Portal** page in the Clerk Dashboard and the **Redirects** tab, where fallback redirect URLs for sign-in, sign-up, and other Account Portal flows can be managed.
- Adds a note clarifying that `redirect_url` must use the instance domain or one of its subdomains, with `localhost` allowed in development.
- Fixes the example explanation to use `https://accounts.myapp.com/sign-in` instead of `http://accounts.myapp.com/sign-in`.

### Deadline
- No rush

### Other resources
- https://clerkinc.slack.com/archives/C01QFRQNHSN/p1776762949745409

### DIFF
--- a/docs/guides/account-portal/direct-links.mdx
+++ b/docs/guides/account-portal/direct-links.mdx
@@ -5,7 +5,7 @@ description: Learn how to share direct links to your Account Portal pages, and h
 
 ## Redirect URL
 
-If a user accesses an Account Portal page _directly_, the `redirect_url` query param will not be present, so the user cannot be redirected back to your application once they are finished with their Account Portal flow. To prevent this, it is recommend that you always specify the redirect in the link when sharing it.
+When a user accesses an Account Portal page directly, there is no `redirect_url` query param present, so Clerk has no way to redirect the user back to your application after their flow is complete. To ensure users are redirected back, include the `redirect_url` query param in any direct links you share.
 
 You can use the following format for your direct links:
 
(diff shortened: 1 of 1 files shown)


manovotny:
Pushed some changes directly in 246efe46.

- Added a cross-link to [Customize your redirect URLs](/docs/guides/development/customize-redirect-urls) so self-hosted component users still have a path off this page — addresses the concern about losing the prop/env-var details without re-introducing the original Account Portal confusion.
- Tightened the `redirect_url` `[!NOTE]`. Tracing `isAllowedRedirect` in `clerk/javascript`, localhost works in dev because of the same-origin check, not because dev mode unlocks it. New wording reflects that.
- "they are still sent" → "they're still sent" and dropped the gerund ("after completing their flow" → "after they complete their flow") to match the styleguide.
- "select the **Redirects** tab" → "open the **Redirects** tab" to match the styleguide's worked example for this exact dashboard section.
- Fixed a pre-existing "it is recommend" → "it is recommended" typo in the first paragraph while we were in there.

Let me know if you agree with these changes or not.

f-westergren:
🚀 LGTM

manovotny (approved):

PR #3327, opened 2026-04-28
Update Docs for SDK Satellite Domain API and Script Tag Usage

SarahSoutoul:
### 🔎 Previews:

- https://clerk.com/docs/pr/ss-docs-11525/guides/dashboard/dns-domains/satellite-domains
- https://clerk.com/docs/pr/ss-docs-11525/js-frontend/getting-started/quickstart

A customer integrating satellite domains with ClerkJS was confused because the required configuration is split across two places:

- `domain` only works when set during initialization
- `isSatellite` is configured during `load()`

That split was not obvious in the docs, especially for script-tag users, and it led to trial-and-error around where each option belongs. The docs also did not clearly state that script-tag attributes must be prefixed with `data-clerk-*`, which made the CDN integration path harder to discover and easier to misconfigure. These updates make the configuration model explicit and add examples for the non-Next.js cases that were missing.

In addition to solving this issue, I also restructured a lot of the page to make it look for UI friendly and easier to read, using `<Tabs>` and `<CodeBlockTabs>`. The changes also add/ fix anything that was wrong or confusing when testing this myself. 

***What changed***

- Clarifies that `domain` must be set during initialization, while `isSatellite` is set in `Clerk.load()`
- Documents that script-tag attributes must use the `data-clerk-*` prefix
- Expands the satellite domains guide beyond Next.js to cover direct `@clerk/clerk-js` and raw `<script>` integrations
- Lots of restructure changes. 

### Deadline

No rush. 

### Other resources

Linear: https://linear.app/clerk/issue/DOCS-11525/update-docs-for-sdk-satellite-domain-api-and-script-tag-usage


manovotny:
Pushed some changes directly in 53fc0b8a.

- Fixed the `data-clerk-*` callout link — was pointing at `#properties` (Clerk instance read-only props), now points at `#constructor` where the constructor options live.
- In the satellite `proxy.ts` example, dropped the `http://` from the commented-out dev `domain` (`'http://localhost:3001'` → `'localhost:3001'`) so it matches every other dev `domain` in the file.
- Reworded the `isSatellite` description in `clerk-options.mdx` to contrast with `domain` (constructor vs `load()`) instead of saying "set this in `load()`" inside the `load()` options section.
- Trimmed the dev comment block in the NPM-module and `<script>` primary-app examples — `signInUrl`/`signUpUrl` are identical in dev and prod for the primary, so only `allowedRedirectOrigins` needs to be flagged as different.
- Tightened the `allowedRedirectOrigins` row in the properties table to "A list of origins that the primary domain is allowed to redirect back to."

Let me know if I got anything wrong.

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,459 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
   - Using environment variables
   - Using properties
 
-  Use the following tabs to select your preferred method. Clerk recommends using environment variables.
+  Use the following tabs to select your preferred method. Clerk recommends using **environment variables**.
 
   <Tabs items={["Environment variables", "Properties"]}>
     <Tab>
-      You can configure your satellite application by setting the following environment variables:
+      To configure satellite domains with environment variables, you need to set values in both your primary app and your satellite app:
```
manovotny: Since this PR adds plain `@clerk/clerk-js` and `<script>` paths to the Properties tab, this tab feels like the odd one out. All the env var names are `NEXT_PUBLIC_*` and the follow-up is `<ClerkProvider>`. A non-Next.js reader who picks "Environment variables" gets Next.js examples.

Would it be worth labelling this tab Next.js-only, or splitting it with framework sub-tabs to match the Properties side?

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,459 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                      <Component {...pageProps} />
+                    </ClerkProvider>
+                  )
+                }
+                ```
+              </CodeBlockTabs>
+            </Tab>
+
+            <Tab>
+              In the Next project associated with your satellite domain, configure your `<ClerkProvider>` as shown in the following example:
```
manovotny: Within the same step the satellite app shows dev/prod two different ways: 

1. Inline `// Or, in development:` comments here and in the NPM module example
2. `<CodeBlockTabs options={["Development", "Production"]}>` for the env vars and the script-tag example. 

Is there a reason to mix them, or should we pick one pattern across the page?

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,459 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                  ```
+
+                  > [!NOTE]
+                  > If you enable `satelliteAutoSync: true`, hardcoded primary sign-in or sign-up URLs can still work because the satellite app syncs automatically on first load.
+                </Tab>
+              </Tabs>
+            </Tab>
+
+            <Tab>
+              For apps that initialize Clerk via a `<script>` tag, configure constructor options on the ClerkJS script and runtime options in Clerk's [`load()`](/docs/reference/objects/clerk#load) method.
```
manovotny: The satellite `<script>` block below shows the full ClerkJS tag with `data-clerk-domain` added, but drops the `@clerk/ui` script and the `ui: { ClerkUI: window.__internal_ClerkUICtor }` option that the JS quickstart sets up. A reader who copies this whole block as-is will lose the prebuilt UI components.

Should we either frame these as additive deltas ("add `data-clerk-domain` to your existing ClerkJS tag") or include the UI bundle so the snippet is self-contained? 

The NPM examples have a similar gap.

SarahSoutoul: Decided to go the route of including the UI bundle, and using `+`  markers to show the additions. Changes here: [Changes post feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/c097ad0ae647ee591926bf095d6f6a31331b483f). @manovotny

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Good pt! Made them all consistent by using the dev/prod diff ways rather than using inline comments. 

Changes here: [Changes post feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/c097ad0ae647ee591926bf095d6f6a31331b483f)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Decided to go the route of labeling it Next.js and add a sentence highlighting it can be reworked to be used with other frameworks. Changes here: [Changes post feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/c097ad0ae647ee591926bf095d6f6a31331b483f)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+  There are two ways to configure your primary and satellite apps to work together:
 
   - Using environment variables
   - Using properties
 
-  Use the following tabs to select your preferred method. Clerk recommends using environment variables.
+  Use the following tabs to select your preferred method. Clerk recommends using **environment variables**.
 
-  <Tabs items={["Environment variables", "Properties"]}>
+  <Tabs items={["Environment variables (Next.js)", "Properties"]}>
```
alexisintech: I actually would not add "Next.js" to the tab name, as it reads like: if you are on a different SDK, then that tab is not for you. 
but! I do agree with adding the
**This example is written for Next.js App Router but can be adapted for any framework/language.**

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
     <Tab>
-      You can configure your satellite application by setting the following environment variables:
+      To configure satellite domains with environment variables in Next.js, you need to set values in both your primary app and your satellite app. This example is written for Next.js App Router **but can be adapted for any framework/language**.
```
alexisintech: I'd bold the entire thing, I know there's an instance where only half is bolded, but in a majority of other cases, we have the adapt line fully bolded.


```suggestion
      To configure satellite domains with environment variables, you need to set values in both your primary app and your satellite app. **This example is written for Next.js App Router but can be adapted for any framework/language**. Next.js prefixes environment variables with `NEXT_PUBLIC_` if they need to be exposed to the frontend. You must update this prefix for the framework or environment variable loader that you are using.
```

we could even throw it in a callout so that it doesn't get missed:

```suggestion
      To configure satellite domains with environment variables, you need to set values in both your primary app and your satellite app.
      > [!WARNING]
      > This example is written for Next.js App Router but can be adapted for any framework/language. Next.js prefixes environment variables with `NEXT_PUBLIC_` if they need to be exposed to the frontend. You must update this prefix for the framework or environment variable loader that you are using.
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
-                <body>
-                  <ClerkProvider allowedRedirectOrigins={['http://localhost:3001']}>{children}</ClerkProvider>
-                </body>
-              </html>
-            )
-          }
-          ```
+      <Tabs items={["Primary app", "Satellite app"]}>
+        <Tab>
+          Add the following environment variables to your primary domain application:
```
alexisintech: incase users are not signed in to the docs/dashboard, let's let them know where they can get their api keys

```suggestion
          Add the following environment variables to your primary domain application. Your [Publishable Key](!publishable-key) and [Secret Key](!secret-key) can always be retrieved from the [**API keys**](https://dashboard.clerk.com/~/api-keys) page in the Clerk Dashboard.
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
-          }
+          ```env {{ filename: '.env' }}
+          NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY={{pub_key}}
+          CLERK_SECRET_KEY={{secret}}
+          NEXT_PUBLIC_CLERK_SIGN_IN_URL=/sign-in
+          NEXT_PUBLIC_CLERK_SIGN_UP_URL=/sign-up
           ```
-        </CodeBlockTabs>
+
+          You will also need to add the `allowedRedirectOrigins` property to `<ClerkProvider>` on your _primary domain app_ to ensure that the redirect back from primary to satellite domain works correctly. For example:
```
alexisintech: I'd argue this can get removed as it's in the "Primary app" tab.
If the consensus is to keep it, we should bold instead of italicize as that's the common pattern in the docs to emphasize things.

```suggestion
          You will also need to add the `allowedRedirectOrigins` property to `<ClerkProvider>` to ensure that the redirect back from primary to satellite domain works correctly. For example:
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+              )
+            }
+            ```
+          </CodeBlockTabs>
+        </Tab>
+
+        <Tab>
+          Add the following environment variables to your satellite domain application:
+
+          <CodeBlockTabs options={["Development", "Production"]}>
```
alexisintech, line 197: it's tough that the development publishable and secret key show in the production tab:

Image

I have two ideas:
1. get the injection to support injecting production keys. implication: security reasons? the secret key is hidden in the docs  so this should satisfy that implication?
2. we have the callout above:
> [!NOTE]
      > In development, your [Publishable Key](!publishable-key) and [Secret Key](!secret-key) will start with `pk_test_` and `sk_test` respectively. In production, they will start with `pk_live_` and `sk_live` respectively.

but I'm wondering if we should instead have comments **in the code examples** so that readers don't miss that information, and get that information right in the spots where they need it.
something like:
```env
# In development, your API keys will start with with `pk_test_` and `sk_test` respectively. 
# In production, they will start with `pk_live_` and `sk_live` respectively.
# Ensure you load the correct keys
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
alexisintech: also, I know we're only writing this guide for next.js, but what if their framework doesn't use clerkprovider? should we add a bit about "if your framework doesn't use clerkprovider, pass this property to your clerk integration. refer to the quickstart guide to find where to pass this property."
that's off the top of my head, it of course can be reworked!

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+      | Property | Description |
+      | - | - |
+      | `isSatellite` | Defines the app as a satellite app when `true`. |
+      | `domain` | Sets the domain of the satellite application. This is required because the [Publishable Key](!publishable-key) is shared across your multi-domain apps, so Clerk cannot infer the satellite domain from it. |
+      | `signInUrl` | The URL used when signing in on your satellite application. It needs to point to your primary application. This option is optional for production instances and required for development instances. |
+      | `signUpUrl` | The URL used when signing up on your satellite application. It needs to point to your primary application. This option is optional for production instances and required for development instances. |
+      | `allowedRedirectOrigins` | A list of origins that the primary domain is allowed to redirect back to. |
 
       > [!TIP]
       > The `URL` parameter that can be passed to `isSatellite` and `domain` is the request url for server-side usage or the current location for client usage.
```
alexisintech:
```suggestion
      > The `URL` parameter that can be passed to `isSatellite` and `domain` is the request URL for server-side usage or the current location for client usage.
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
-            // Always run for Clerk-specific frontend API routes
-            '/__clerk/(.*)',
-          ],
-        }
-        ```
+      Use the following tabs to see the setup for your app type:
+
+      <Tabs items={["Next.js", "JavaScript"]}>
+        <Tab>
+          In a Next.js setup, configure [`<ClerkProvider>`](/docs/reference/components/clerk-provider) in both your primary and satellite apps, then update [`clerkMiddleware()`](/docs/reference/nextjs/clerk-middleware) in the satellite app.
```
alexisintech: it's required to update clerkmiddleware to do this property option - so i'm assuming other sdk's cannot do the property option? if so, we should declare that somewhere.

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+          > [!IMPORTANT]
+          > You should set your `CLERK_PUBLISHABLE_KEY` and `CLERK_SECRET_KEY` in your environment variables even if you're using props to configure satellite domains.
+
+          <Tabs items={["Primary app", "Satellite app"]}>
+            <Tab>
+              In the Next.js project associated with your primary domain, configure your `<ClerkProvider>` as shown in the following example:
+
+              <CodeBlockTabs options={["Development", "Production"]}>
+                <CodeBlockTabs options={["App Router", "Pages Router"]}>
+                  ```tsx {{ filename: 'app/layout.tsx' }}
```
alexisintech: let's get highlights on what we're adding, the variables and the props added to clerkprovider. all these code examples should have highlights of what the user is adding because their app/layout.tsx may have a lot more code added and they need to understand what exactly they're adding to their code

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+              And the middleware associated with your satellite domain should look like this:
+
+              <Include src="_partials/nextjs/nextjs-15-callout" />
+
+              <CodeBlockTabs options={["Development", "Production"]}>
+                ```ts {{ filename: 'proxy.ts' }}
+                import { clerkMiddleware, createRouteMatcher } from '@clerk/nextjs/server'
+
+                // Set the homepage as a public route
+                const isPublicRoute = createRouteMatcher(['/'])
```
alexisintech: we don't force auth on all routes anymore. let's remove this pattern - its opt in now.

```suggestion
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                  domain: 'localhost:3001',
+
+                  // satelliteAutoSync defaults to false. Uncomment below to automatically sync auth state on first load.
+                  // This adds a redirect on every first visit, which comes with a performance cost.
+                  // satelliteAutoSync: true,
+                }
+
+                export default clerkMiddleware(async (auth, req) => {
+                  if (isPublicRoute(req)) return // if it's a public route, do nothing
+                  await auth.protect() // for any other route, require auth
```
alexisintech: same comment as above - removing force auth for all routes

```suggestion
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                    '/__clerk/(.*)',
+                  ],
+                }
+                ```
+
+                ```ts {{ filename: 'proxy.ts' }}
+                import { clerkMiddleware, createRouteMatcher } from '@clerk/nextjs/server'
+
+                // Set the homepage as a public route
+                const isPublicRoute = createRouteMatcher(['/'])
```
alexisintech: same comment as before
```suggestion
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                  domain: 'satellite.dev',
+
+                  // satelliteAutoSync defaults to false. Uncomment below to automatically sync auth state on first load.
+                  // This adds a redirect on every first visit, which comes with a performance cost.
+                  // satelliteAutoSync: true,
+                }
+
+                export default clerkMiddleware(async (auth, req) => {
+                  if (isPublicRoute(req)) return // if it's a public route, do nothing
+                  await auth.protect() // for any other route, require auth
```
alexisintech: same as before
```suggestion
```

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                    +   signInUrl: '/sign-in',
+                    +   signUpUrl: '/sign-up',
+                    +   allowedRedirectOrigins: ['http://localhost:3001'],
+                      })
+                    ```
+
+                    ```js {{ filename: 'src/main.js' }}
+                      import { Clerk } from '@clerk/clerk-js'
+
+                      const clerk = new Clerk('{{pub_key}}')
```
alexisintech: same idea as before, we should add a comment about injecting the correct key pk_test vs pk_live

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -120,256 +120,652 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+
+                        // satelliteAutoSync defaults to false. Uncomment below to automatically sync auth state on first load.
+                        // satelliteAutoSync: true,
+                      })
+                    ```
+
+                    ```js {{ filename: 'src/main.js' }}
+                      import { Clerk } from '@clerk/clerk-js'
+
+                      const clerk = new Clerk('{{pub_key}}', {
```
alexisintech: same idea as before, we should add a comment about injecting the correct key pk_test vs pk_live

SarahSoutoul: Okay changes added here: [Changes to feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/223a4b668ac1008dbedd7c8b1da06b736dd002cf)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Okay changes added here: [Changes to feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/223a4b668ac1008dbedd7c8b1da06b736dd002cf)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Okay changes added here: [Changes to feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/223a4b668ac1008dbedd7c8b1da06b736dd002cf). 

Let me know what you think!

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Okay changes added here: [Changes to feedback](https://github.com/clerk/clerk-docs/pull/3327/commits/223a4b668ac1008dbedd7c8b1da06b736dd002cf)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Changes added here: [Add highlights and note on middelware](https://github.com/clerk/clerk-docs/pull/3327/commits/0486f9a1286184b8d508cce26ed32b93982becc0)

docs/guides/dashboard/dns-domains/satellite-domains.mdx
SarahSoutoul: Changes added here: [Add highlights and note on middelware](https://github.com/clerk/clerk-docs/pull/3327/commits/0486f9a1286184b8d508cce26ed32b93982becc0)

SarahSoutoul:
@alexisintech @manovotny okay think I've addressed all comments and feedback on this! Let me know thoughts

alexisintech:
I've posted my docs review:

- In the warning callout at the top of the doc, re-ordered the information to keep the flow of information sequential
- Updated the heading for the "Using `satelliteAutoSync: false` (default) with sign-in/sign-up links" section to "Build sign-in and sign-up links" to be more clear on the action the user is taking
  - Moved the callout above the example to keep the order of information sequential
  - Replaced `<Tabs>` with `<If>` to be more SDK-specific
- Removed the callout about development vs. production keys in favor of comments in the code example itself to remove visual noise, keep the information close to the source, and so that the user doesn't miss a very important piece of information
- Replaced `<Tabs>` with `<If>` to be more SDK-specific. 

I think this is fantastic work for the scope of this PR - very nice job Sarah!!
Once this is shipped, there should be a follow-up PR to add more SDK-specificity. Right now, it's very specific to Next.js which is a problem in some areas for users that are using other SDK's:
- There's a middleware section that is Next.js specific
- The examples rely on ClerkProvider, and some SDK's have different integrations

We can tag team the follow up work if that interests you, we can pair on it :) There is an example in the demo repo of how to do satellite apps for React SDK apps: https://github.com/clerk/clerk-multidomain-demo
And we can utilize `<If>` to make it more specific for apps that aren't using ClerkProvider
I think we can get this guide more SDK-specific all around! I've created a ticket here: https://linear.app/clerk/issue/DOCS-11843/make-the-satellite-domains-guide-more-sdk-specific

Nice work on this Sarah 😸💖

alexisintech:
another question!

do we want to keep pages router support in this? we haven't been adding pages router examples to our guides, and imo, I think we can remove it! the examples are not much different.

SarahSoutoul:
> another question!
> 
> do we want to keep pages router support in this? we haven't been adding pages router examples to our guides, and imo, I think we can remove it! the examples are not much different.

Yeah that's a fair statement! Let me remove them. 

In terms of the rest, thanks for creating a ticket - happy to work on this with you and pair. Def think we could make this guide even better by making it SDK-specific, but as you said, not the scope of this PR/ ticket.

SarahSoutoul:
Okay @alexisintech - have done the following:

- Removed pages router: [Remove pages router](https://github.com/clerk/clerk-docs/pull/3327/commits/7748499d8b89ba0d31696f23dd38048dcea3901f)
- Removed an unnecessary line that didn't make sense after switching to <If>: [Remove unnecessary line](https://github.com/clerk/clerk-docs/pull/3327/commits/be20710ec8f2d53551dd9e736de4a3293624ef63)
- Created a [JS PR](https://github.com/clerk/javascript/pull/8845) to update the description of `isSatellite` prop, whcih was part of the Linear ticket. That description change was included in this PR initially but with the /object/ reference docs changes, the partial containing the prop got removed and is now coming from Typedoc. 

I'll look at the [ticket](https://linear.app/clerk/issue/DOCS-11843/make-the-satellite-domains-guide-more-sdk-specific) you created next and come up with a proposal, and we can jump on a call sometime next week to discuss.

manovotny:
Pushed some changes directly in 112443ba.

- Reworded the `data-clerk-*` callout — the script tag reads exactly three attributes (`data-clerk-publishable-key`, `data-clerk-proxy-url`, `data-clerk-domain`, verified in `clerk-js/src/index.browser.ts`), so the callout now names them instead of saying "any Clerk option," and points everything else at `load()`.
- Added a sentence to the npm-module intro noting the examples build on the JavaScript Quickstart setup — without the quickstart's `@clerk/ui` loader, `window.__internal_ClerkUICtor` is undefined and the prebuilt components silently won't mount.
- Simplified `clerkMiddleware(async (auth, req) => {}, options)` to `clerkMiddleware(options)` — it's a documented overload, and the empty callback left unused params behind after the force-auth removal.

docs/guides/dashboard/dns-domains/satellite-domains.mdx
```diff
@@ -124,256 +128,535 @@ When `satelliteAutoSync` is `false` (the default), you must ensure your sign-in
...
+                    import { Clerk } from '@clerk/clerk-js'
+
+                    // In development, your API keys will start with `pk_test_` and `sk_test_` respectively. Ensure you load the correct keys.
+                    const clerk = new Clerk('{{pub_key}}', {
+                  +   domain: 'localhost:3001',
+                    })
+
+                    await clerk.load({
+                      ui: { ClerkUI: window.__internal_ClerkUICtor },
+                  +   isSatellite: true,
```
manovotny, line 487: Heads up: the regenerated reference this guide links to shows `load()` typed as `Without<ClerkOptions, 'isSatellite'>` — while the `ClerkOptions` table on the same page lists `isSatellite`. 

The guide is right at runtime: clerk-js's `load()` accepts full `ClerkOptions`, the `isSatellite` getter reads it from there, and the constructor only takes `domain`/`proxyUrl`, so `load()` is the only path for plain JS. 

clerk/javascript#8845 fixes the description and clerk/javascript#8846 fixes the type — once those release, the Typedoc sync resolves the contradiction. Nothing to change in this PR.

manovotny (approved):

PR #3338, opened 2026-05-02
docs: flutter v1

MarkOSullivan94:
### 🔎 Previews:

* https://clerk-git-nick-add-flutter-sdk.clerkstage.dev/docs/pr/fork-markosullivan94-pr-flutter-docs/flutter/getting-started/quickstart

> [!NOTE]
> This is the V1 of the Flutter docs so it doesn't have a "Build your own UI" section which both Android and iOS SDKs have in their documentation. This will be submitted at a later date and will go into further detail on how to use `clerk_auth` instead of `clerk_flutter`.

This pull request introduces documentation for integrating Clerk authentication into Flutter applications using the [`clerk_flutter`](https://pub.dev/packages/clerk_flutter) package.

**Flutter Quickstart and Social Sign-In Integration:**

- Added a new `Flutter Quickstart` guide that walks users through setting up Clerk authentication in a Flutter app, including installation, Android configuration, and usage of prebuilt Clerk widgets.
- Introduced detailed guides for implementing native "Sign in with Apple" and "Sign in with Google" in Flutter, covering both prebuilt UI and custom authentication flows, platform-specific requirements, and troubleshooting tips. [[1]](diffhunk://#diff-7a027609840e4cc472d50fc92296663e223475430e1543bedd4f2f80dc07232cR1-R117) [[2]](diffhunk://#diff-2601eacfe091a87467f55d23b115b45ac63979b52b137f4c8dd6726dc0615ff9R1-R204)
- Updated the getting started overview to include Flutter as a supported SDK, with navigation and icon support.

**Clerk Widget Theming for Flutter:**

- Added a new reference page for `ClerkThemeExtension`, explaining how to customize Clerk widget colors and text styles using Flutter's theming system, with code samples for global and scoped theming, custom text styles, and light/dark mode support.

### Deadline

- No rush

### Other resources

- https://pub.dev/packages/clerk_flutter
- https://pub.dev/packages/clerk_auth
- PR for adding Flutter to the docs sidebar → https://github.com/clerk/clerk/pull/2579


manovotny:
Pushed fixes directly across two commits — 39326ea (the co-review pass) and 355e828 (Google guide follow-up). A run-down:

- **Compile-breaking samples:** `ClerkAuth` now takes `config: ClerkAuthConfig(...)` (not a top-level `publishableKey`); `oauthSignIn(strategy: Strategy.oauth_google)` → `ssoSignIn(context, Strategy.oauthGoogle)`; `ClerkThemeExtension` is `final` (it has no const constructor).
- **Links:** fixed the broken `ClerkAuth` / auth-flows / widgets-overview links and the `clerk-flutter` → `clerk-sdk-flutter` GitHub URLs; renamed `user-profile-views.mdx` → `user-profile.mdx` so the sidenav entry resolves (the page was unreachable).
- **Navigation/IA:** grouped Apple/Google under a **Social connections** section (matching iOS/Android) and backticked `ClerkAuth` in the nav.
- **Google social connection:** swapped the hand-rolled "Enable Google" / "Set the Client ID and Secret" steps for the shared `enable-google` / `set-client-id-secret` partials — the hand-rolled version was missing the required **Use custom credentials** toggle (verified against the Dashboard).
- **Styleguide:** American spelling, `sh {{ filename: 'terminal' }}` on terminal blocks, sentence-case headings, "sidenav", canonical Dashboard-nav phrasing, plus a few callout/heading fixes.
- **Misc:** added `flutter` to the `schemas.ts` icon enum, converted the image placeholders to MDX comments, and ran Prettier across the changed files.

`pnpm run build:tsx` passes with no warnings (it was failing on 13 broken-link/manifest warnings before).

Please lmk if anything looks incorrect. I am not a Flutter expert, by any means. I am leaving comments for the remaining things I have questions about.

docs/reference/flutter/auth.mdx
```diff
@@ -0,0 +1,410 @@
...
+  child: ClerkAuthBuilder(
+    signedInBuilder: (context, auth) => const HomeScreen(),
+    signedOutBuilder: (context, auth) => const ClerkAuthentication(),
+  ),
+)
+```
+
+Other pre-built widgets:
+
+- `ClerkSignInPanel` — sign-in form only
```
manovotny: `ClerkSignInPanel`, `ClerkSignUpPanel`, `ClerkOAuthPanel`, and `ClerkForgottenPasswordPanel` aren't exported from `package:clerk_flutter/clerk_flutter.dart` — they live in `src/` but aren't in the public barrel, so a reader importing the package can't actually use them. 

Are these meant to be public? If so they probably need exporting in the SDK first; if not, maybe drop them here. Same question applies to the `ClerkOAuthPanel` mentions in the two social sign-in guides.

docs/reference/flutter/widgets/user/user-profile.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final user = ClerkAuth.userOf(context);
+
+    if (user == null) {
+      return const SizedBox.shrink();
+    }
+
+    return const ClerkUserProfile();
```
manovotny: Same export question as the panels in `auth.mdx`: `ClerkUserProfile` is declared in `src/` but isn't exported from `clerk_flutter.dart`, so `const ClerkUserProfile()` won't resolve for a reader using the public package. 

`ClerkAuthentication`, `ClerkUserButton`, and `ClerkOrganizationList` are exported and fine, so this looks like an SDK oversight. Worth confirming whether to export it in the SDK or hold this page (plus its nav entry and the mention in widgets/overview) until it's public.

docs/getting-started/quickstart.flutter.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+  ```xml
+  <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+            xmlns:tools="http://schemas.android.com/tools">
+      <uses-permission android:name="android.permission.INTERNET"/>
+      ...
+  </manifest>
+  ```
+
+  ## Use Clerk's prebuilt widgets
```
manovotny: This is the only place the quickstart introduces `ClerkAuth`, and it links out to the example app rather than showing the setup. 

For a quickstart, could we show the actual wiring here — wrapping the app in `ClerkAuth(config: ClerkAuthConfig(publishableKey: '{{pub_key}}'), child: ...)` with `ClerkAuthBuilder` — so a reader reaches a running app without leaving the page.

docs/reference/flutter/user.mdx
```diff
@@ -0,0 +1,207 @@
...
+final auth = ClerkAuth.of(context);
+
+await auth.updateUser(
+  username: 'janedoe',
+  firstName: 'Jane',
+  lastName: 'Doe',
+);
+```
+
+### Set profile image
```
manovotny, line 34: Most H3s here jump straight from heading to code block. The styleguide wants a short lead-in before each example ("The following example demonstrates…"). 

Same pattern in `auth.mdx` and `organizations.mdx` — low priority, but worth a pass for consistency with the rest of the docs.

docs/reference/flutter/organizations.mdx
```diff
@@ -0,0 +1,173 @@
...
+final auth = ClerkAuth.of(context);
+final invitations = await auth.fetchOrganizationInvitations();
+```
+
+### Accept invitation
+
+```dart
+final auth = ClerkAuth.of(context);
+
+await auth.acceptOrganizationInvitation(invitation);
```
manovotny, line 159: `invitation` isn't defined in this snippet — it would read clearer paired with the `fetchOrganizationInvitations()` result just above (e.g. `final invitation = invitations.first;`). 

Same with the bare `session` in `clerk-auth.mdx`, and the `firstWhere(...)` calls in `user.mdx` will throw if there's no match (no `orElse`). 

All illustrative, so your call on how far to flesh them out.

docs/guides/configure/auth-strategies/sign-in-with-google.flutter.mdx
```diff
@@ -0,0 +1,203 @@
...
+    await authState.idTokenSignIn(
+      provider: clerk.IdTokenProvider.google,
+      token: token,
+    );
+
+    // If this is a new user, a sign-up object may be returned with missing fields
+    if (authState.signUp case clerk.SignUp signUp
+        when signUp.missingFields.isNotEmpty) {
+      final nameParts = account.displayName?.split(' ') ?? [];
+      await authState.attemptSignUp(
```
manovotny, line 165: The example app's custom sign-up handles `legalAccepted` when the instance requires legal acceptance — worth confirming whether this flow needs it too, or it could fail sign-up on instances that have it enabled.

docs/getting-started/quickstart/overview.mdx
```diff
@@ -74,6 +74,12 @@ description: See the getting started guides and tutorials.
 
   ---
 
+  - [Flutter](/docs/flutter/getting-started/quickstart)
+  - Use the Clerk Flutter SDK to authenticate users in your Flutter applications.
+  - <Icon name="flutter" />
```
manovotny, line 79: Heads up that `<Icon name="flutter" />` (and the `flutter` TutorialHero icon in the guides) needs the SVG registered in `clerk/clerk`. 

I can help land that in https://github.com/clerk/clerk/pull/2579 — just wanted to flag.

docs/reference/flutter/widgets/authentication/clerk-authentication.mdx
```diff
@@ -0,0 +1,78 @@
+---
+title: '`ClerkAuthentication`'
+description: Clerk's ClerkAuthentication widget renders a UI for signing in and signing up users in your Flutter app.
+sdk: flutter
+---
+
+{/* ![The ClerkAuthentication widget renders a comprehensive authentication interface that handles both user sign-in and sign-up flows.](/docs/images/ui-components/flutter-clerk-authentication.png){{ style: { maxWidth: '460px' } }} */}
```
manovotny: These pages stub a screenshot placeholder, but the `flutter-*` images don't exist in the repo yet — same on `user-button.mdx` and `user-profile.mdx`. 

Not a bug, just a decision worth making deliberately: iOS and Android ship real screenshots on the equivalent view pages (precedent for adding them), while Expo ships none (precedent for skipping). 

Both are fine, and screenshots drift as the UI changes. Want to add `flutter-clerk-*.png` assets (with dark variants, like iOS/Android) and uncomment these, or drop the placeholders for now?

docs/getting-started/quickstart/overview.mdx
manovotny: Confirmed we're good here. Resolving.

docs/reference/flutter/widgets/authentication/clerk-authentication.mdx
MarkOSullivan94: I think the screenshots are a good idea but perhaps we could include a caption to make them aware of what version they were taken from?

This would help readers understand what it should look like and if it does look a bit different they will be able to recognise it was from an earlier version.

docs/reference/flutter/auth.mdx
MarkOSullivan94: These are not meant to be public. 

I'll update this page and the social sign-in guides to remove them.

docs/reference/flutter/widgets/user/user-profile.mdx
MarkOSullivan94: I'm going to remove this page for now as it's not included in the current version so it wont block releasing the documentation

docs/reference/flutter/user.mdx
MarkOSullivan94: I'll make this change. 

Admittedly I didn't see the styleguide before writing the docs, instead I focused on using both the docs for Android and iOS as a guide and now that I just looked there was quite a few cases where neither of them had a short lead-in before many of the code examples.

docs/getting-started/quickstart.flutter.mdx
MarkOSullivan94: Done. 

I've replaced the section with a `main.dart` snippet.

docs/reference/flutter/user.mdx
MarkOSullivan94: I've added lead-in sentences to every H3/H4 section that went straight from heading to code block across `auth.mdx`, `user.mdx`, and `organizations.mdx`.

docs/reference/flutter/organizations.mdx
MarkOSullivan94: Fixed all three issues:

- `organizations.mdx`: the Accept invitation snippet now fetches the list inline and assigns `final invitation = invitations.first;` so the variable is defined when `acceptOrganizationInvitation` is called.

- `clerk-auth.mdx`: added `final session = auth.session;` immediately before `signOutOf(session!)` so the variable isn't bare.

- `user.mdx`: replaced all four `firstWhere` calls (two for email addresses, two for phone numbers) with `firstWhereOrNull` from `package:collection` (already a direct dependency of `clerk_flutter`). Removed the force-unwrap `!` in each case and added an explicit `if (x == null) return;` guard instead.

docs/guides/configure/auth-strategies/sign-in-with-google.flutter.mdx
MarkOSullivan94: I've added a `[!NOTE]` callout after the `attemptSignUp` snippet. 

It explains that instances with legal acceptance enabled will have `clerk.Field.legalAccepted` in `missingFields`, that `attemptSignUp` will fail without it, and gives the correct conditional pattern: `legalAccepted: signUp.missing(clerk.Field.legalAccepted) ? true : null`. 

Kept it as a callout rather than changing the snippet itself, since how consent is collected (checkbox, modal, etc.) is app-specific.

docs/reference/flutter/widgets/authentication/clerk-authentication.mdx
MarkOSullivan94: Added the screenshots now

manovotny:
Pushed a few follow-ups in 0de68ce.

- Repaired a rebase artifact in production.mdx — the Publishable Key bullet and TIP callout were duplicated and the new `<If is="llm">` block landed mid-list (this was also failing Prettier/CI Lint)
- Defined `session`/`organization` in the auth.mdx session-management snippets — same treatment you gave clerk-auth.mdx
- Mirrored your `legalAccepted` note into the Apple guide (same custom sign-up flow, same failure mode)
- Pointed the legal acceptance link at /docs/guides/secure/legal-compliance — the old `authentication/configuration/legal-acceptance` path 404s, and relative links get build-validated
- Reworded the quickstart `main.dart` lead-in to the styleguide form

`pnpm run build:tsx` and Prettier both pass clean. Everything else from the latest round looked great — nice work turning all of that around.

docs/reference/flutter/user.mdx
```diff
@@ -0,0 +1,246 @@
...
+  IdentifierType.emailAddress,
+);
+```
+
+### Verify email address
+
+The following example demonstrates how to verify an email address with the code sent to the user.
+
+```dart
+import 'package:collection/collection.dart';
```
manovotny: `firstWhereOrNull` comes from `package:collection`, which these snippets import, but the quickstart never adds as a direct dependency. It resolves transitively, but trips Dart's `depend_on_referenced_packages` lint in apps that enable it. 

Maybe worth adding `collection` to the `flutter pub add` line, or a half-sentence here? Thoughts?

docs/reference/flutter/widgets/authentication/clerk-authentication.mdx
manovotny: Screenshots look good. 

Markdown image titles render as captions here, eg. `alt`. 

iOS/Android don't do this today, so it'd be a new pattern, which is fine. We can update iOS/Android as a follow up improvement to their docs.

manovotny (commented):
Great improvements from the last round of questions. A couple more for you.

manovotny:
Ran the docs end-to-end on a clean machine today (fresh Flutter install, new Clerk app, iPhone simulator) — good news and one blocker.

**The good news: the quickstart works exactly as written once the package installs.** Followed it verbatim, then drove the full flow on the simulator: sign-up → email verification (a wrong code is correctly rejected, with the error surfacing through the `ClerkErrorListener` wiring) → "You are signed in!" → session persisted across a cold start. A factory-fresh dev instance also worked without touching the Native API toggle, so the "ensure it's enabled" phrasing is right.

**The blocker: clerk_flutter 0.0.15-beta on pub.dev is missing its `lib/generated/` localization files** — clerk/clerk-sdk-flutter#402, which you triaged. `flutter pub add clerk_flutter clerk_auth` resolves 0.0.15-beta, so the quickstart currently fails with ~50 compile errors for every reader at "Run your application". The files are absent from the tarball itself (0.0.14-beta's archive has them), so the pub-dev dartdoc issue won't fix it — it needs a republish.

I don't think this should merge until that's resolved — a broken first build would be a rough debut. A republish (0.0.16-beta) with `lib/generated/` included means the docs ship exactly as written. I've also added an inline suggestion on the install step covering the version constraint, with a stopgap option if the release takes a while.

(Small things noticed while running it, for whenever: the separate `flutter pub get` step is a no-op after `flutter pub add`, and the AndroidManifest snippet shows `xmlns:tools`, which the Flutter 3.44 scaffold doesn't have.)

docs/reference/flutter/installation.mdx
```diff
@@ -0,0 +1,36 @@
...
+  ```
+
+  **Manually via `pubspec.yaml`**
+
+  In your `pubspec.yaml`, add `clerk_flutter` and `clerk_auth` under `dependencies`, then run `flutter pub get`.
+
+  ```yaml
+  dependencies:
+    clerk_flutter: ^0.0.15-beta
+    clerk_auth: ^0.0.15-beta
```
manovotny: These hardcoded constraints go stale on every release — for pre-1.0 packages, `^0.0.15-beta` only allows `<0.0.16`. The Android installation doc uses a `<latest-version>` placeholder instead; suggest matching it:

```suggestion
    clerk_flutter: ^<latest-version>
    clerk_auth: ^<latest-version>
```

(If clerk/clerk-sdk-flutter#402 drags on, this is also where the docs could temporarily pin `0.0.14-beta` — I verified every API the quickstart uses exists in that release. The placeholder is the right end state either way.)

MarkOSullivan94:
> Ran the docs end-to-end on a clean machine today (fresh Flutter install, new Clerk app, iPhone simulator) — good news and one blocker.
> 
> **The good news: the quickstart works exactly as written once the package installs.** Followed it verbatim, then drove the full flow on the simulator: sign-up → email verification (a wrong code is correctly rejected, with the error surfacing through the `ClerkErrorListener` wiring) → "You are signed in!" → session persisted across a cold start. A factory-fresh dev instance also worked without touching the Native API toggle, so the "ensure it's enabled" phrasing is right.
> 
> **The blocker: clerk_flutter 0.0.15-beta on pub.dev is missing its `lib/generated/` localization files** — [clerk/clerk-sdk-flutter#402](https://github.com/clerk/clerk-sdk-flutter/issues/402), which you triaged. `flutter pub add clerk_flutter clerk_auth` resolves 0.0.15-beta, so the quickstart currently fails with ~50 compile errors for every reader at "Run your application". The files are absent from the tarball itself (0.0.14-beta's archive has them), so the pub-dev dartdoc issue won't fix it — it needs a republish.
> 
> I don't think this should merge until that's resolved — a broken first build would be a rough debut. A republish (0.0.16-beta) with `lib/generated/` included means the docs ship exactly as written. I've also added an inline suggestion on the install step covering the version constraint, with a stopgap option if the release takes a while.
> 
> (Small things noticed while running it, for whenever: the separate `flutter pub get` step is a no-op after `flutter pub add`, and the AndroidManifest snippet shows `xmlns:tools`, which the Flutter 3.44 scaffold doesn't have.)

Great news about the quickstart 🎉 

I need to get @slightfoot to do the new release

docs/reference/flutter/installation.mdx
MarkOSullivan94: I like this approach of using `^<latest-verison>`

docs/reference/flutter/user.mdx
MarkOSullivan94: It doesn't feel right to suggest adding `collection` package into this.

I've refactored it to replace all four `firstWhereOrNull` calls with `.where(...).firstOrNull`, this removes the need for the `collection` package. 

`.where(...).firstOrNull` is built into Dart 3.0+ (the project requires Dart ≥ 3.6.2), so no additional dependency is needed.

manovotny:
Re-ran the full verification against the published 0.0.16-beta. Two-part update: the docs verify clean — and the sign-in testing surfaced an SDK gap that I think should hold the merge.

**Everything documented works verbatim, unpatched:** `flutter pub add` resolves 0.0.16-beta with `lib/generated/` intact (clerk/clerk-sdk-flutter#402 confirmed fixed at the consumer level), the quickstart builds and launches clean on the iOS simulator, the example app fresh-resolves and compiles, the `main.dart` block matches the working app byte-for-byte, and the new `where(...).firstOrNull` snippets analyze clean with no extra imports. Nice turnaround on all three follow-up commits.

**The gap: clerk/clerk-sdk-flutter#418.** With [Client Trust](https://clerk.com/docs/guides/secure/client-trust) enabled — the default on new instances — signing out and signing back in verifies the password and then dead-ends in `ClerkAuthentication`: the panel has no UI for `needs_client_trust`, so it renders empty, with no error and no session. Reproduced on the same install (sign out → sign in), on fresh installs, and in the SDK's own example app; root cause confirmed from the persisted client state.

**Why I think this holds the merge**: a reader finishes the quickstart, wires `signOut()` per the auth reference, signs out, and can't sign back in — silently — on a default-configured instance. Same posture as #402: the fix belongs in the SDK (handling `needs_client_trust` in the prebuilt UI), and the docs ship cleanly once it lands.

manovotny:
No longer needed — Flutter docs now live in the SDK repo → https://github.com/clerk-community/clerk-sdk-flutter/pull/432

PR #3344, opened 2026-05-04
chore: Update Next.js middleware matcher to match on `/__clerk/` paths

brkalow:
### 🔎 Previews:

-

- Ensures our documented middleware matcher includes routes that start with `/__clerk/`. This is important because our new proxy behavior relies on the clerk middleware executing for the proxy path, which starts with `/__clerk/` by default.

### Deadline

- No rush

### DIFF
--- a/docs/guides/development/upgrading/upgrade-guides/core-2/nextjs.mdx
+++ b/docs/guides/development/upgrading/upgrade-guides/core-2/nextjs.mdx
@@ -92,6 +92,8 @@ export const config = {
     '/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)',
     // Always run for API routes
     '/(api|trpc)(.*)',
+    // Always run for Clerk-specific frontend API routes
+    '/__clerk/(.*)',
   ],
 }
 ```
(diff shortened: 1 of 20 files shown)


thiskevinwang (approved):

Railly (approved):

docs/guides/development/upgrading/upgrade-guides/core-2/nextjs.mdx
```diff
@@ -92,6 +92,8 @@ export const config = {
     '/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)',
     // Always run for API routes
     '/(api|trpc)(.*)',
+    // Always run for Clerk-specific frontend API routes
+    '/__clerk/(.*)',
```
alexisintech, line 96: I noticed you updated these examples in the /core-2 doc - but I think this middleware update would only be for core 3? is that correct?

brkalow: @alexisintech I did, it would be applicable for folks using the proxy in core 2 as well. I opted to just change every place in this PR, acknowledging that the core 2 upgrade guide is probably not referenced much anymore.

alexisintech (approved):
🚢 🚢 🚢

PR #3353, opened 2026-05-07
Include in build script output llms-full.txt per sdk

NWylynko:
## Summary

Companion to [clerk/clerk#2385](https://github.com/clerk/clerk/pull/2385). That PR replaces `/docs/llms-full.txt` with a small index that links out to per-SDK full-context dumps at `/docs/llms/{sdk}/llms-full.txt`. Those routes call `getFileContents('_llms/{sdk}/llms-full.txt', branch)` from this repo — so we need to actually produce those files in the build.

## What this changes

- New optional `llms.perSdkFullPath` config (template, defaults to `_llms/{sdk}/llms-full.txt`). When set, the build emits one bundle per canonical SDK in `validSdks` after the existing `_llms/llms-full.txt`.
- `filterOutputDocsForSdk(docs, targetSdk, validSdks)` decides what goes in each per-SDK bundle:
  - Files under `{targetSdk}/…` — kept (the real SDK-specific content).
  - Files under `{otherSdk}/…` — dropped.
  - Core (non-prefixed) files — dropped if shadowed by any `{anySdk}/<same path>` (those core files are landing-redirect pages).
  - Single-SDK docs that don't get their own variant directory (`sdkScoped: true` at the core path) — included only when `targetSdk` is in `availableSdks`.
- `listOutputDocsFiles` now also extracts `sdkScoped` + `availableSdks` from each doc's frontmatter so the filter has what it needs for single-SDK docs.
- The old `_llms/llms-full.txt` is preserved, it will still be needed until clerk/clerk is updated

The `{sdk}` keys in the path are the canonical 16 SDKs from `VALID_SDKS` — matching what `clerk/clerk` resolves migration aliases to via `sdkMap`.

## Verification

- `pnpm exec vitest run` — 191/191 tests pass, including:
  - `scripts/lib/llms.test.ts` — unit tests for `sdkPrefixOfPath` (incl. `react` vs `react-router` disambiguation) and `filterOutputDocsForSdk` (core/shadowed/single-SDK/other-SDK cases).
  - `scripts/build-docs.test.ts` — two new integration tests covering the per-SDK output (multi-SDK doc with `.react.mdx` variant + shared core doc; and skipping emission for SDKs with zero content).
- `pnpm run build` against the real docs:
  - Emits all 16 per-SDK bundles plus the existing combined one.
  - Sample sizes: combined 367k lines · `nextjs` 123k · `react` 119k · `js-frontend` 114k · `go` 96k.
  - Each per-SDK file has zero `redirectPage:` frontmatter and a single distinct `activeSdk:` value matching the bundle.

## Follow-ups

- Once this lands, the routes added in clerk/clerk#2385 will start serving real content. Until then those routes 404 (called out in that PR's testing notes).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/scripts/build-docs.test.ts
+++ b/scripts/build-docs.test.ts
@@ -7467,6 +7467,121 @@ sourceFile: /docs/api-doc.mdx
 # API Documentation
 `)
   })
+
+  test('Should output per-SDK llms-full.txt files containing the SDK variant for shadowed docs and core docs that are not SDK-scoped', async () => {
+    const { tempDir, readFile, listFiles } = await createTempFiles([
+      {
+        path: './docs/manifest.json',
+        content: JSON.stringify({
+          navigation: [
+            [
+              { title: 'Quickstart', href: '/docs/quickstart' },
+              { title: 'API Doc', href: '/docs/api-doc' },
+            ],
+          ],
+        }),
+      },
+      {
+        path: './docs/quickstart.mdx',
+        content: `---
+title: Quickstart
+description: Shared
+---
+
+Shared content
+`,
+      },
+      {
+        path: './docs/api-doc.mdx',
+        content: `---
+title: API Documentation
+description: x
+sdk: nextjs, react
+---
+
+Default content
+`,
+      },
+      {
+        path: './docs/api-doc.react.mdx',
+        content: `---
+title: API Documentation for React
+description: x
+---
+
+React content
+`,
+      },
+    ])
+
+    await build(
+      await createConfig({
+        ...baseConfig,
+        basePath: tempDir,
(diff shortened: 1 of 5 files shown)


NWylynko:
Going to not move ahead with this, I think https://github.com/clerk/clerk/pull/2388 is more important and will not work with this pr in the current iteration of it. Once the repos are merged could re-visit this idea but the code with be completely different.

This is because this pr splits the llm-full.txt's out by sdk, written out in their built state, but without the mdx to md conversion. The #2388 pr disreguards the llm-full.txt that the built script exports as it builds it itself using directory.json and loading the mdx files directly, so it can then transform this individually.

PR #3355, opened 2026-05-08
chore: Add docs for appearance options elevation

alexcarpenter:
### 🔎 Previews:

- https://clerk.com/docs/pr/carp-appearance-flush-option/nextjs/guides/customizing-clerk/appearance-prop/options

- Adds `elevation="raise|flush"` docs

### Deadline

-

### Other resources

- https://github.com/clerk/javascript/pull/8510

### DIFF
--- a/docs/guides/customizing-clerk/appearance-prop/captcha.mdx
+++ b/docs/guides/customizing-clerk/appearance-prop/captcha.mdx
@@ -4,7 +4,7 @@ description: Utilize Clerk's captcha prop in order to change the appearance of t
 sdk: astro, chrome-extension, expo, nextjs, nuxt, react, react-router, tanstack-react-start, vue, js-frontend, fastify, expressjs, go, ruby
 ---
 
-{/* JS file: https://github.com/clerk/javascript/blob/main/packages/types/src/appearance.ts#L538 */}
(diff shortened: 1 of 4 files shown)


manovotny:
Pushed some changes directly in e82ab2983.

- Added the session task components (`<TaskChooseOrganization />`, `<TaskResetPassword />`, `<TaskSetupMFA />`) to the `elevation` "applies to" list, since the option covers them too. Kept "such as" because `<OAuthConsent />` and `<UserVerification />` are also affected but have no public doc pages to link.
- Broadened the `options.mdx` meta `description` to match your new intro line ("layout and behavior of Clerk's prebuilt components") and added the missing Oxford comma.
- Fixed the `{/* JS file: ... */}` source pointers across the appearance docs (`options`, `overview`, `variables`, `captcha`) — they all pointed at `packages/types/src/appearance.ts`, which now 404s. Repointed them to `packages/ui/src/internal/appearance.ts` and dropped the line anchors so they don't silently drift.
- Alphabetized `animations` before `autoFocus`.

I know these slightly increase scope, but these are fixes I noticed while reviewing your changes. Figured we'd fix them while we're here.

manovotny (approved):

PR #3361, opened 2026-05-11
docs(expo): document EAS production caveat for Google Sign-In client IDs

### LINKED ISSUE #3256: @clerk/expo native Google Sign-In fails in production EAS builds when EXPO_PUBLIC_CLERK_GOOGLE_WEB_CLIENT_ID is not in expoConfig.extra (works in dev)
In Expo + EAS production builds, useSignInWithGoogle() can throw:

Google Sign-In credentials not found. Please set EXPO_PUBLIC_CLERK_GOOGLE_WEB_CLIENT_ID in your .env file. Even when the variable is correctly configured in EAS environment variables and visible in build logs.

This works in dev builds, but fails in production/TestFlight.

Environment
@clerk/expo: 3.1.2
Expo SDK: 55
EAS Build + EAS Update
iOS production build (TestFlight)
Custom flow using useSignInWithGoogle()
What I observed
EXPO_PUBLIC_CLERK_GOOGLE_WEB_CLIENT_ID is set in Expo/EAS env dashboard for the production environment.
Build logs show env vars are loaded.
Dev build works.
Production build fails with missing web client ID error from Clerk Google hook.
Root cause (likely)
In @clerk/expo, Google credentials are read from:

Constants.expoConfig.extra.EXPO_PUBLIC_CLERK_GOOGLE_WEB_CLIENT_ID
fallback: process.env.EXPO_PUBLIC_CLERK_GOOGLE_WEB_CLIENT_ID
process.env fallback is in Clerk code under node_modules. Expo does not inline EXPO_PUBLIC_* into node_modules in production bundle transforms, so this fallback can be undefined in production.

Relevant Expo behavior/source:

https://github.com/expo/expo/blob/952bd96c660b1f8df1a324adf1df8ca66f56a7a1/packages/babel-preset-expo/src/__tests__/inline-env-vars.test.ts#L145-L166
Relevant Clerk code path:

packages/expo/src/hooks/useSignInWithGoogle.shared.ts (compiled: dist/hooks/useSignInWithGoogle.shared.js)

Request
Could Clerk docs explicitly mention this production caveat for Expo/EAS native Google Sign-In?

Docs page:

https://clerk.com/docs/expo/guides/configure/auth-strategies/sign-in-with-google

chriscanin:
## Summary

- Document that `EXPO_PUBLIC_*` env vars are **not** inlined into `node_modules` (this is intentional Expo behavior, per [Expo docs](https://docs.expo.dev/guides/environment-variables/) — "Code inside node_modules is not affected for security purposes"), so production EAS builds of `useSignInWithGoogle()` need the Google client IDs mirrored into `expoConfig.extra`.
- Add an `IMPORTANT` callout + tabbed `app.json` / `app.config.ts` examples to the "Configure environment variables" step.
- Update the existing iOS URL scheme `extra` example to include the client IDs in the same block so users don't accidentally overwrite them.

## Why

The `@clerk/expo` hook reads `Constants.expoConfig.extra.EXPO_PUBLIC_CLERK_GOOGLE_*` first and falls back to `process.env.EXPO_PUBLIC_CLERK_GOOGLE_*`. Because Expo's babel preset doesn't inline `EXPO_PUBLIC_*` into `node_modules` for security reasons, the `process.env` fallback resolves to `undefined` in production EAS bundles, and the hook throws `Google Sign-In credentials not found...` even when the EAS env vars are correctly configured and visible in build logs. The docs previously only mentioned `.env`, leaving users no way to make the values reachable in production.

Fixes #3256

## Test plan

- [ ] Preview deploys render the new callout + tabs correctly
- [ ] iOS / Android tabs both show the new `extra` block
- [ ] iOS URL scheme section still works and now shows cumulative `extra` shape

### DIFF
--- a/docs/_partials/expo/authview-apple-callout.mdx
+++ b/docs/_partials/expo/authview-apple-callout.mdx
@@ -1,2 +1,0 @@
-> [!IMPORTANT]
-> If you're using [`<AuthView />`](/docs/reference/expo/native-components/auth-view) from `@clerk/expo/native`, you do **not** need to install `expo-apple-authentication`, `expo-crypto`, or use the `useSignInWithApple()` hook — `<AuthView />` handles the sign-in flow automatically.
--- a/docs/_partials/expo/authview-google-callout.mdx
+++ b/docs/_partials/expo/authview-google-callout.mdx
@@ -1,2 +1,0 @@
-> [!IMPORTANT]
-> If you're using [`<AuthView />`](/docs/reference/expo/native-components/auth-view) from `@clerk/expo/native`, you do **not** need to install `expo-crypto` or use the `useSignInWithGoogle()` hook — `<AuthView />` handles the sign-in flow automatically.
--- a/docs/getting-started/quickstart.expo.mdx
+++ b/docs/getting-started/quickstart.expo.mdx
@@ -428,13 +428,7 @@ Use the following tabs to choose your preferred approach:
 
       ## Native sign-in with Google and Apple (optional)
 
-      If you want to add native Sign in with Google and Sign in with Apple buttons that authenticate without opening a browser, you'll need to install `expo-crypto`:
-
-      ```bash
-      npx expo install expo-crypto
-      ```
-
-      Then, refer to the [Sign in with Google](/docs/guides/configure/auth-strategies/sign-in-with-google) and [Sign in with Apple](/docs/guides/configure/auth-strategies/sign-in-with-apple) guides for full setup instructions, including any additional dependencies specific to each provider. **This approach requires a [development build](https://docs.expo.dev/develop/development-builds/introduction/) because it uses native modules. It cannot run in Expo Go.**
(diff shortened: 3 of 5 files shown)


alexisintech:
hmm, but this info is already included in the following section: ## Set up native Sign in with Google (iOS only)
so we could change the title/description of that section to be tailored for "EAS / Production / iOS builds" or just make it required for everyone, so that no one ever runs into these issues 🤔

alexisintech (approved):
thank you for the updates chris, it looked much better!

I made a few adjustments to the copy, and changed the callout to be a quiz so it's clear on why the callout is there. 
while taking a look at the guide, I also made some other adjustments like removing the partial (it was only used in this one place, so a partial isn't necessary). this led me to updating the apple version of the partial, and thus led me to updating to the expo quickstart as well - there was some redundant instructions there about installing `expo-crypto` which is already included in both of the guides that that section links to!

I've approved but if you have any suggestions on the changes, feel free to discuss. if you're cool with the changes, feel free to merge 😸💖

PR #3364, opened 2026-05-12
Add documentation for Organizations integration with OAuth IdP

jfoshee:
### 🔎 Previews:

- [How Clerk Implements OAuth](https://clerk.com/docs/pr/cursor-oauth-orgs-integration-docs-76a1/guides/configure/auth-strategies/oauth/how-clerk-implements-oauth)
- [Use OAuth for Single Sign-On](https://clerk.com/docs/pr/cursor-oauth-orgs-integration-docs-76a1/guides/configure/auth-strategies/oauth/single-sign-on)

Adds documentation for the Organizations integration with OAuth Applications (OAuth IdP). When an instance has Organizations enabled, a new `user:org:read` scope becomes available for OAuth apps.

**Changes made:**

1. **`how-clerk-implements-oauth.mdx`:**
   - Added `user:org:read` to the scopes table with a description of its availability and purpose.
   - Added a new "Organizations and OAuth" subsection under "Consent screen management" explaining:
     - The Organization selector shown on the OAuth consent screen when this scope is requested.
     - The `org_id` claim included in access tokens and ID tokens, and at info endpoints
   - Updated the authorization server metadata example to include `user:org:read` in `scopes_supported` and `org_id` in `claims_supported`.

2. **`single-sign-on.mdx`:**
   - Added `org_id`, `org_name`, `org_slug` to the `/oauth/userinfo` response example and properties table.
   - Added `org_id` to the `/oauth/token_info` response example with a clarifying note.
   - Added `org_id` to the ID token additional claims table.

3. Added a `slug` tooltip

### Deadline

- ASAP if possible. (Sorry I released this feature thinking I already had a green light here.)

Part of USER-5273

### DIFF
--- a/docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
+++ b/docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
@@ -143,6 +143,9 @@ export async function GET() {
 }
 ```
 
+> [!NOTE]
+> The `org_id` property is only present in the response when Organizations are enabled and the `user:org:read` scope was granted. See [Organizations and OAuth](/docs/guides/configure/auth-strategies/oauth/how-clerk-implements-oauth#organizations-and-oauth).
+
 The `/oauth/userinfo` endpoint provides the following user properties, depending on the granted scopes:
 
 | Property | Description |
@@ -159,7 +162,7 @@ The `/oauth/userinfo` endpoint provides the following user properties, depending
 | `public_metadata` | The user's public metadata |
 | `private_metadata` | The user's private metadata |
 | `unsafe_metadata` | The user's unsafe metadata |
-| `org_id` | The ID of the Organization the user selected during the OAuth consent flow. Only present when the `user:org:read` scope is granted. |
+| `org_id` | The ID of an Organization to which the user belongs. See [Organizations and OAuth](/docs/guides/configure/auth-strategies/oauth/how-clerk-implements-oauth#organizations-and-oauth). |
 
 ### Get token information
 
@@ -222,9 +225,6 @@ export async function POST(req: NextRequest) {
   "org_id": "string"
 }
 ```
-
-> [!NOTE]
-> The `org_id` property is only present in the response when Organizations are enabled and the `user:org:read` scope was granted. See [Organizations and OAuth](/docs/guides/configure/auth-strategies/oauth/how-clerk-implements-oauth#organizations-and-oauth).
 
 ## Get additional user information with the OAuth 2.0 flow using the OpenID Connect (OIDC) protocol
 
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
looks good on my end!

jescalan (approved):
LGTM, assuming we are still planning to turn on org integration by default, which I know was under discussion. If its changing to be a thing you need to activate within oauth config, we may need to adjust wording a little bit

manovotny:
Pushed some changes directly in 9d713192f.

- userinfo: updated the note so it covers `org_name` and `org_slug` too (they're conditional on the same thing as `org_id`), and added that an Organization has to be associated with the token
- token_info: added the conditional note for `org_id` that the PR description mentioned but wasn't in the diff
- metadata example: added a note that `user:org:read` and `org_id` only show up when Organizations are enabled
- userinfo properties table: capitalized "Organization" in the `org_name`/`org_slug` rows to match the rest of the docs
- consent section: reworded so it's clear the `org_id` claim is in the access token always, but in the ID token only when `openid` is also requested (also un-split a paragraph)
- slug tooltip: sharable -> shareable

One question still incoming as a review comment.

Let me know if any of these look off to you.

docs/guides/configure/auth-strategies/oauth/single-sign-on.mdx
manovotny: This authenticates with a Bearer access token, but `/oauth/token_info` is protected by the OAuth client credentials. The WARNING above says Client ID and Client Secret, and that's what the endpoint expects: HTTP Basic auth, with the token to introspect in the `token` body field. As written, this returns a 401.

I think we need something like this for it to be correct, but could love a second opinion:

```ts
// clientId and clientSecret come from your Clerk OAuth app
const credentials = Buffer.from(`${clientId}:${clientSecret}`).toString('base64')

const response = await fetch(`${process.env.CLERK_FRONTEND_API_URL}/oauth/token_info`, {
  method: 'POST',
  headers: {
    Authorization: `Basic ${credentials}`,
    'Content-Type': 'application/x-www-form-urlencoded',
  },
  body: new URLSearchParams({ token: tokens.refreshToken }),
})
```

This is pre-existing (not introduced by this PR), but noticed it while review. Should we fix it while we're here (if you agree it needs changed)?

jfoshee: You are right; that's a good catch. I just verified w/ a request like this:

```
curl -Ss -X POST \
https://....clerk.accountsstage.dev/oauth/token_info \
-H 'Authorization: Basic abc...==' \
-H 'Content-Type: application/x-www-form-urlencoded' \
-d 'token_type_hint=access_token' \
-d 'token=eyJ...'
```

And the token type hint is optional.

manovotny:
Pushed one more in fabebc615.

- Defined `clientId` and `clientSecret` as `'YOUR_X_HERE'` placeholders in the `/oauth/token_info` example so the snippet actually compiles; the previous fix referenced them without declaring them. Also reworded the preamble to link to the Dashboard's OAuth apps page and nudge toward env vars in production.

manovotny (approved):

jfoshee:
Thanks for all the improvements @manovotny! 🚀

PR #3365, opened 2026-05-13
Add clarity surrounding user enumeration protection

alexisintech:
### 🔎 Previews:

-

`secure/user-enumeration-protection`: tested bulk versus strict user enumeration flows when signing up/signing in and updated the docs accordingly.

`development/custom-flows/authentication/sign-in-or-up`: updated the copy to more clearly explain how the two flow options relate to user enumeration protection settings.

### Deadline

No rush

### Other resources

https://linear.app/clerk/issue/DOCS-11693/clarify-sign-in-or-up-behavior-with-strict-user-enumeration-protection

### DIFF
--- a/docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
+++ b/docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
@@ -26,7 +26,9 @@ search:
 This guide demonstrates how to build a custom user interface that **allows users to sign up or sign in within a single flow**. There are two approaches:
 
 - [**Standard flow**](#standard-sign-in-or-up-flow): The sign-in attempt immediately tells you whether an account exists. If it doesn't exist yet, you start the sign-up flow. This is simple and gives users immediate feedback, but it reveals whether an account exists before any verification, making it susceptible to [user enumeration](/docs/guides/secure/user-enumeration-protection) attacks.
+  - Mimics the Account Portal/prebuilt components' behavior when [**bulk** user enumeration protection](/docs/guides/secure/user-enumeration-protection#bulk-user-enumeration-protection) is enabled.
 - [**`signUpIfMissing` flow**](#sign-in-or-up-with-sign-up-if-missing): The sign-in proceeds to verification regardless of whether an account exists. Only after verification does the backend reveal whether the account exists or needs to be created. This prevents user enumeration attacks.
+  - Mimics the Account Portal/prebuilt components' behavior when [**strict** user enumeration protection](/docs/guides/secure/user-enumeration-protection#strict-user-enumeration-protection) is enabled.
 
 ## Standard sign-in-or-up flow
 
@@ -37,6 +39,9 @@ This guide demonstrates how to build a custom user interface that **allows users
 <Include src="_partials/custom-flows/enable-email-password" />
 
 ### Build the flow
+
+> [!TIP]
+> This flow mimics the Account Portal/prebuilt components' flows when [**bulk** user enumeration protection](/docs/guides/secure/user-enumeration-protection#bulk-user-enumeration-protection) is enabled.
 
 To blend a sign-up and sign-in flow into a single flow, you must **treat it as a sign-in flow**, but with the ability to sign up a new user if they don't have an account. You can do this by **checking for the `form_identifier_not_found` error** if the sign-in process fails, and then starting the sign-up process.
 
@@ -1226,6 +1231,9 @@ To blend a sign-up and sign-in flow into a single flow, you must **treat it as a
 </If>
 
 ## Sign-in-or-up with `signUpIfMissing`
+
+> [!TIP]
+> This flow mimics the Account Portal/prebuilt components' flows when [**strict** user enumeration protection](/docs/guides/secure/user-enumeration-protection#strict-user-enumeration-protection) is enabled.
 
 <If notSdk={["ios", "android"]}>
   The `signUpIfMissing` option offers a privacy-preserving alternative to the standard sign-in-or-up flow. Unlike the standard flow, which discloses whether an account exists from the outset, this option proceeds to the verification step regardless of if the account exists. Only after the user has successfully completed the verification step does the flow reveal if an account already exists or if one needs to be created. Although it is **recommended** to pair the `signUpIfMissing` flow with [strict user enumeration protections](/docs/guides/secure/user-enumeration-protection) in the Clerk Dashboard for maximum security, **this option doesn't require that setting.**
--- a/docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
+++ b/docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
@@ -32,6 +32,9 @@ This guide demonstrates how to build a custom user interface that **allows users
 
 ## Standard sign-in-or-up flow
 
+> [!TIP]
+> This flow mimics the Account Portal/prebuilt components' behavior when [**bulk** user enumeration protection](/docs/guides/secure/user-enumeration-protection#bulk-user-enumeration-protection) is enabled.
+
 ### Enable email and password authentication
 
 **This example uses the [email and password sign-in custom flow](/docs/guides/development/custom-flows/authentication/email-password) as a base. However, you can modify this approach according to the settings you've configured for your application's instance in the Clerk Dashboard.**
@@ -39,9 +42,6 @@ This guide demonstrates how to build a custom user interface that **allows users
 <Include src="_partials/custom-flows/enable-email-password" />
 
 ### Build the flow
-
-> [!TIP]
-> This flow mimics the Account Portal/prebuilt components' flows when [**bulk** user enumeration protection](/docs/guides/secure/user-enumeration-protection#bulk-user-enumeration-protection) is enabled.
 
 To blend a sign-up and sign-in flow into a single flow, you must **treat it as a sign-in flow**, but with the ability to sign up a new user if they don't have an account. You can do this by **checking for the `form_identifier_not_found` error** if the sign-in process fails, and then starting the sign-up process.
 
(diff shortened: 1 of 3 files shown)


SarahSoutoul (approved):

manovotny:
Pushed a few small cleanups directly in 7c3e04a7.

- Moved the **Standard sign-in-or-up flow** TIP up to sit directly under its H2 so it mirrors the `signUpIfMissing` TIP's placement (the TIP describes the flow as a whole, not just the build step).
- Changed `flows` → `behavior` in both TIPs to match the intro sub-bullets and fix the singular-subject/plural-object mismatch.
- Tightened the strict email-code line in `user-enumeration-protection.mdx` to drop the doubled "there was": `… explaining that there was a sign-in attempt, but there was no account found.` → `… explaining that someone attempted to sign in to a non-existent account.`

Left a couple of pending review comments on broader questions I didn't have answers to.

Let me know what you think.

docs/guides/secure/user-enumeration-protection.mdx
```diff
@@ -35,7 +35,7 @@ Strict enumeration covers:
 - Sign up: If you sign up with an email or phone number that already exists, Clerk won't send a verification code. Instead, it sends a notification (email or SMS) letting the account owner know someone tried to sign up, and suggesting they sign in instead.
 - Sign in: When the account exists, sign-in proceeds normally. When it doesn't, Clerk's behavior depends on the strategy:
   - **Password or Web3 wallet**: The attempt is rejected without revealing whether the account exists.
-  - **Email code or email link**: Clerk sends a notification to the address explaining that someone tried to sign in, and suggesting they sign up instead.
+  - **Email code or email link**: Clerk shows the verification screen and sends a notification to the address explaining that someone attempted to sign in to a non-existent account.
```
manovotny, line 41: This says strict's email-code path sends a notification only (no real code, so the user is stuck on the verification screen), but the `signUpIfMissing` flow doc says it sends a real code that succeeds and transfers to sign-up. 

If prebuilt components use `signUpIfMissing` under the hood when strict is on and public sign-up is allowed, this line might be describing the restricted/waitlist case specifically. 

Should we narrow this to the scenario it actually covers, or update the `signUpIfMissing` flow doc to match?

docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
```diff
@@ -26,10 +26,15 @@ search:
 This guide demonstrates how to build a custom user interface that **allows users to sign up or sign in within a single flow**. There are two approaches:
 
 - [**Standard flow**](#standard-sign-in-or-up-flow): The sign-in attempt immediately tells you whether an account exists. If it doesn't exist yet, you start the sign-up flow. This is simple and gives users immediate feedback, but it reveals whether an account exists before any verification, making it susceptible to [user enumeration](/docs/guides/secure/user-enumeration-protection) attacks.
+  - Mimics the Account Portal/prebuilt components' behavior when [**bulk** user enumeration protection](/docs/guides/secure/user-enumeration-protection#bulk-user-enumeration-protection) is enabled.
```
manovotny: This sub-bullet and the TIP under the Standard H2 say nearly the same thing about 15 lines apart (same for the `signUpIfMissing` pair). Intentional redundancy for direct-navigation readers, or should one of the two go?

docs/guides/secure/user-enumeration-protection.mdx
alexisintech: oh you know what, it's because I was testing with the fresh nextjs-quickstart app and I assumed the `<SignIn />` component was a combined flow (both sign in and up) but I see you have to opt-in with a prop or env var

alexisintech: okay just tested with the combined flow, and this is still the case. its sending no account found emails, it never sends a code to sign up with... maybe I'm doing this incorrectly?

I have a fresh app, I have `withSignUp={true}` on the `<SignIn />` component so that I am opted in to the combined flow, and this does reflect in the rendered component. however, I try to enter an email for an account that doesn't exist, and I get a "Sign in attempted, no account found" email - I never a get a code to sign-up with.

Image

alexisintech: so this might be an API issue?

docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
alexisintech: I put it at the top as well to help users choose which section they want.
but included it also in the section incase users ignore the introduction section and skip to the section they're looking for.
so that either way, no matter where the users land (intro or specific section), they have that information.

what are your thoughts?

docs/guides/secure/user-enumeration-protection.mdx
manovotny: After pulling up `packages/ui/src/components/SignIn/SignInStart.tsx` in `clerk/javascript` and `api/shared/sign_in/service.go` in `clerk_go` —you're not doing anything wrong, and your test matches what line 38 now describes. I was wrong and didn't dig deep enough. 

The prebuilt `<SignIn withSignUp={true} />` combined flow calls `signIn.create(...)` without `signUpIfMissing` and only transfers to sign-up when FAPI returns `FORM_IDENTIFIER_NOT_FOUND` (`SignInStart.tsx` lines 444–467). When strict is on, FAPI suppresses that error and serves an enumeration-protection response (fake verification screen + notification email), so the transfer branch is unreachable. That's the dead end you're seeing.

`signUpIfMissing` is a different code path on the server. In `service.go` around line 1597, FAPI does:

```
enumerationProtectionEnabled := userSettings.EnumerationProtectionEnabled() || signIn.SignUpIfMissing
```

So when the client opts in, FAPI behaves as if strict were on regardless of the setting, but it also enables the `CreateTransferForSignUpIfMissing` path that turns a missing user into a real `sign_up_if_missing_transfer` after a successful verification. The prebuilt components never opt into this, so under strict they can't reach the transfer.

The cleaner framing:

- **Standard custom flow ≈ prebuilt `<SignIn withSignUp />`** = behavior under bulk / no protection ✅ — the "mimics bulk" line works.
- **`signUpIfMissing` ≠ prebuilt under strict.** It's more capable: under strict, prebuilt gets stuck; `signUpIfMissing` completes the transfer.

TL;DR: We could:

1. Leave this alone/as-is.
2. We could drop the "mimics strict" line on the `signUpIfMissing` flow (both the TIP and the intro sub-bullet) and reframe it as the recommended custom flow when strict is enabled (line 1239 already half-says that).

Totally your call.

docs/guides/development/custom-flows/authentication/sign-in-or-up.mdx
manovotny: Makes sense. Both reader paths land on the info. Leaving both is fine with me. 👍

docs/guides/secure/user-enumeration-protection.mdx
alexisintech: ooo this is great info. so we should also document that the sign-in-or-up flow with prebuilt components/account portal will not work (sign-in's will not transfer to sign-up's) if strict user enumeration is turned on.

and then as for fixing this issue at hand, let's drop the mimics stuff. i will update it now

alexisintech: actually, i readded the tips but changed the word 'mimic' to 'similar to'
because the whole idea behind adding that information is that if someone coming to build a sign-in-or-up custom flow, and they want to know how to recreate a clerk flow, they should know which option to pick in relation to which settings they have enabled in the clerk dashboard.
does that make sense? totally open to reworking it the execution of it, but that's the idea

manovotny: The "similar to" swap works for the **bulk** pairing — the Standard flow really does behave like the prebuilt components under bulk, so that one's fine as-is.

The **strict** one still doesn't quite hold, though. Since the prebuilt components / Account Portal can't complete a sign-in-or-up under strict (the sign-in never transfers to sign-up), `signUpIfMissing` isn't similar to what they do under strict — it does the thing they can't. And it'd sit right next to the caveat you mentioned adding about prebuilt not working under strict, so the two would read as a contradiction.

Your goal still works, I think it just needs a different framing for the strict case. Instead of "similar to prebuilt behavior," maybe anchor it as "use this flow when strict is enabled, because prebuilt can't." Something like:

> Use this flow if your application has strict user enumeration protection enabled. The Account Portal and prebuilt components can't complete a sign-in-or-up under strict — a sign-in for a non-existent account won't transfer to sign-up — so `signUpIfMissing` is how you offer sign-in-or-up with strict turned on.

Same idea for the strict intro sub-bullet. That folds your "prebuilt doesn't work under strict" note in as the reason, rather than a separate caveat.

Does that make sense, or am I not understanding it correctly?

alexisintech: I've updated the copy again to make it more clear - what do you think 👀

manovotny: This looks great! I think we're close. 

I pushed 67164fc8 with some of the last thoughts I had to save some back and forth. Let me know if you dislike any changes and I can remove them.

alexisintech: good idea on the callout! another idea.. we are adding a caveat to user enumeration in https://github.com/clerk/clerk-docs/pull/3368

this "sign-in-or-up doesn't work with strict" issue also feels like a caveat. leads me to think, do we just create a caveats/limitations section? do we rework this section altogether in light of these two PR's 🤔

manovotny: @alexisintech agree, the pattern's there. On naming: `Limitations` is the dominant repo convention (6 instances — bot-protection, client-trust, reverification, etc. — vs one `Caveats`). I'd probably lean into that for wording.

I see three sequencing options:

1. This PR (#3365) does the rework, #3368 rebases on top.
2. #3368 lands first as-is, this PR does the rework and folds @dmoerner's just-merged callout into the new section.
3. Merge both as-is, do the rework in a follow-up PR. Each PR ships unblocked and focused. Short interim window where the strict section has two stacked notes, which is fine as long as the follow-up happens quickly. Bonus: gives space to reconcile @dmoerner's `#### H4` callout style with the `[!NOTE]` style at the same time.

No strong preference from me. Your call.

dmoerner: Sorry, I completely missed this PR while looking at my other one.

alexisintech: merging with the callout for now - then will pull into the other PR, merge that to main and then see if its fine with the callout and edge-case section, or if the entire strict section can get reworked and maybe add a limitations section.

PR #3367, opened 2026-05-13
fix: Remove clerk elements reference from how clerk works

alexcarpenter:
### 🔎 Previews:

- https://clerk.com/docs/pr/carp-move-clerk-elements/guides/how-clerk-works/overview#levels-of-abstraction

- Removes clerk elements reference from how clerk works

### Deadline

no rush
-


manovotny (approved):
We do have a larger deprecation task, but happy to accept this.

PR #3370, opened 2026-05-13
(customizing-clerk/localization) Document how to use pipes and modifiers

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/aa-docs-6116/guides/customizing-clerk/localization#pipe-syntax-and-modifiers

Adds section to https://clerk.com/docs/guides/customizing-clerk/localization about using pipes/modifiers when performing custom localization.

### Deadline

- No rush

### Other resources

- [Relevant Linear ticket](https://linear.app/clerk/issue/DOCS-6116/restore-documentation-on-using-pipesmodifiers-with-the-localization)
- Used https://github.com/clerk/javascript/blob/main/packages/localizations/src/en-US.ts as a resource to write these docs

### DIFF
--- a/docs/guides/customizing-clerk/localization.mdx
+++ b/docs/guides/customizing-clerk/localization.mdx
@@ -383,6 +383,8 @@ Now that you know the key, you can pass it to the `localization` prop and set th
   ```
 </If>
 
+### Pass multiple keys
+
 You can also customize multiple entries by passing multiple keys. The following example updates the "to continue to" subtitles on the `<SignUp />` component to say "to access" instead.
 
 <If notSdk={["astro", "js-frontend", "vue", "nuxt", "fastify"]}>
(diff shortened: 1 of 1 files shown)


docs/guides/customizing-clerk/localization.mdx
```diff
@@ -702,3 +704,19 @@ The following example updates the `not_allowed_access` error message. This messa
   })
   ```
 </If>
+
+### Pipe syntax and modifiers
+
+Clerk localization strings support a small template language. Some values will use `{{ variable | modifier(...args) }}` to pass a runtime value through one or more named formatters. Here's what each part means:
+
+- `variable` — the runtime value (like in the example, a `Date` passed as `date`).
+- `|` — the pipe operator; it sends the value into the modifier to its right.
```
SarahSoutoul:
```suggestion
- `|` — the pipe operator; it passes the value to the modifier on its right.
```

Just a suggestion, but feels a bit smoother to me with "passes" as more common in technical terms.

docs/guides/customizing-clerk/localization.mdx
```diff
@@ -702,3 +704,19 @@ The following example updates the `not_allowed_access` error message. This messa
   })
   ```
 </If>
+
+### Pipe syntax and modifiers
+
+Clerk localization strings support a small template language. Some values will use `{{ variable | modifier(...args) }}` to pass a runtime value through one or more named formatters. Here's what each part means:
+
+- `variable` — the runtime value (like in the example, a `Date` passed as `date`).
```
SarahSoutoul: The "like in" syntax feels a little conversational to me. Could we maybe change it to "for example" or such as?

docs/guides/customizing-clerk/localization.mdx
```diff
@@ -702,3 +704,19 @@ The following example updates the `not_allowed_access` error message. This messa
...
+- `|` — the pipe operator; it sends the value into the modifier to its right.
+- `modifier(...)` — a named formatter with optional arguments (like in the example, the `timeString` formatter with the `'en-US'` locale as an argument).
+
+For example: `"Yesterday at {{ date | timeString('en-US') }}"` would take the `date` and apply the `timeString` formatter with the `'en-US'` locale to get the time string in English, like `'12:00 PM'`.
+
+#### Chaining pipes
+
+You can chain pipes so each step receives the result of the previous one.
+
+For example: `{{ date | weekday('en-US','long') | titleize }}` would take the `date` and apply the `weekday` formatter with the `'en-US'` locale and `'long'` style ('Monday' instead of 'Mon') to get a long English weekday name. Then, `titleize` would ensure the long English weekday name is properly capitalized. Some countries lowercase weekday names while others capitalize them. Overall, a rendered output would be something like `'Monday'` instead of `'Mon'` or `'monday'`.
```
SarahSoutoul:
```suggestion
For example: `{{ date | weekday('en-US','long') | titleize }}` would take the `date` and apply the `weekday` formatter with the `'en-US'` locale and `'long'` style ('Monday' instead of 'Mon') to get a long English weekday name. Then, `titleize` would ensure the long English weekday name is properly capitalized. This is useful because some locales return lowercase weekday names while others capitalize them, so the final output would consistently be something like `'Monday'` rather than `'Mon'` or `'monday'`.
```

Just a suggestion, but think "some locales" instead of "some countries" work a bit better? It feels a bit more precise, since capitalization conventions are generally language/locale-specific.

SarahSoutoul (approved):

PR #3372, opened 2026-05-14
(custom-flows/account-updates/forgot-password) Add signOutOfOtherSessions prop usage

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/aa-docs-11739/guides/development/custom-flows/account-updates/forgot-password

### What does this solve? What changed?

- Adds `signOutOfOtherSessions` prop usage to the custom flow on resetting a user's password. This demonstrates how to sign users out of all other sessions after resetting a password.

### Deadline

- No rush

### Other resources

https://linear.app/clerk/issue/DOCS-11739/document-signoutofothersessions-parameter-in-signinresetpassword

PR #3373, opened 2026-05-14
(custom-flows) Add update password guide; move forgot password guide

alexisintech:
### 🔎 Previews:

-

While working on [(custom-flows/account-updates/forgot-password) Add signOutOfOtherSessions prop usage](https://github.com/clerk/clerk-docs/pull/3372#top), I realized that the forgot password guide was incorrectly placed in the `/account-updates` grouping for the custom flow guides. It's not an account update; it's part of the sign-in flow for when a user is signing in but forgot their password. This PR moves it to the `/authentication` grouping. This PR also adds an update password guide that covers what was supposed to be in `/account-updates`: a user can reset their password **after they're already signed in**.

### Deadline

-


manovotny:
Pushed some changes directly in 86db646a.

- Added the missing `kotlinx.coroutines.flow.update` import in the Android `update-password.mdx` example so it actually compiles.
- Swapped `JSON.stringify(error)` for `error.message` in the Next.js `update-password.mdx` example — `JSON.stringify` on an `Error` instance returns `"{}"`, so the displayed error wouldn't have been useful.
- Renamed the Next.js example's component from `UserProfilePage` to `UpdatePasswordPage` so it matches the page and filename.
- Added a reciprocal TIP in `update-password.mdx` pointing back to the forgot-password guide, and reworded both TIPs to be declarative (e.g. "To allow users to…, see the…") instead of the "Looking for…?" question framing, which doesn't match the pattern used elsewhere in the docs.
- Changed the pre-existing `type="phoneNumber"` to `type="tel"` on the phone input in `forgot-password.mdx` — `phoneNumber` isn't a valid HTML input type and falls back to plain text (no numeric keyboard on mobile). Was in a hunk you'd already touched, figured I'd grab it.
- Simplified the `forgot-password.mdx` title to `Build a custom forgot password flow` to match the description and read more naturally.
- Moved the new "Forgot password" sidenav entry to sit right after "Email & password" in both manifest locations, so the password sign-in flow and the password recovery flow are adjacent.

Feel free to adjust, or let me know if you want me to undo any changes.

docs/manifest.json
```diff
@@ -692,6 +692,10 @@
                                 "title": "Email & password",
                                 "href": "/docs/guides/development/custom-flows/authentication/email-password"
                               },
+                              {
+                                "title": "Forgot password",
```
manovotny, line 696: The active forgot-password guide moves to `authentication/`, but the legacy version stays under `account-updates/legacy/forgot-password`. 

Would it make sense to move the legacy variant to `authentication/legacy/forgot-password` so both variants live under the same parent grouping (or in a follow up PR), or is it good to leave as is?

alexisintech: I was torn between the two, but since you're leaning this way, I agree!! moving forward, it's best to keep the two layouts the same

manovotny:
Pushed one more change in ee4f5a05.

- Updated the legacy guide link in `authentication/forgot-password.mdx` to point at the new `authentication/legacy/forgot-password` path, since the source-level link was still going through the redirect to the old `account-updates/legacy/` location.

manovotny (approved):

PR #3375, opened 2026-05-15
Revamp organizations/add-members/invitations

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/aa-docs-11769/guides/organizations/add-members/invitations

Sending organization invitations requires emails. This PR adds a solution for users that don't want to enable **Email** for their application:
> If you don't want to enable **Email** for your application, you can bypass the invitation flow completely and add users directly to the Organization by using the [`createOrganizationMembership()`](/docs/reference/backend/organization/create-organization-membership) method.

This PR also revamps the page by adding links to the BAPI reference, splitting up information visually using `<Tabs>` for cURL versus clerkClient examples, and overall using more explicit language in the copy.

### Deadline

No rush

### Other resources

https://linear.app/clerk/issue/DOCS-11769/document-workaround-for-org-invites-with-google-only-auth-skip

### DIFF
--- a/.superset/config.json
+++ b/.superset/config.json
@@ -1,4 +1,6 @@
 {
-  "setup": ["./scripts/ai/setup.sh"],
+  "setup": [
+    "./scripts/ai/setup.sh"
+  ],
   "teardown": []
 }
--- a/.superset/config.json
+++ b/.superset/config.json
@@ -1,6 +1,4 @@
 {
-  "setup": [
-    "./scripts/ai/setup.sh"
-  ],
+  "setup": ["./scripts/ai/setup.sh"],
   "teardown": []
 }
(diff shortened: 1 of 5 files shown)


manovotny:
Pushed some changes directly in 6a163d64b.

- Qualified the `createOrganizationMembership()` workaround as adding **existing** users — the method takes a `user_id`, so the invitee must already have a Clerk account (relevant for the Google-only-auth case this PR documents).
- Reworded the redirect-URL and metadata `clerkClient` notes so the bulk method is correct: `redirectUrl`/`publicMetadata` go on each object in the `params` array, not as top-level params.
- Restored the `429` / `Retry-After` guidance that was dropped along with the rate-limit callout.
- Added `{{ filename: 'terminal' }}` to the `redirect_url` and metadata cURL examples for consistency with the others.
- Fixed `createOrganizationInvitationBulk()`'s return type to `Promise<OrganizationInvitation[]>` (it returns an array — verified against the BAPI handler and the JS SDK type).
- Documented `privateMetadata` and `expiresInDays` on both `create-organization-invitation` and `create-organization-invitation-bulk` so the sibling pages stay in sync.

Left `inviterUserId` as `string | null` (the OpenAPI spec marks it `nullable: true`) and the "max 10 per bulk request" figure as-is (confirmed it's the default in the backend).

Let me know how the changes look to you, but pre-approving.

manovotny (approved):

PR #3389, opened 2026-05-22
fix(totp): Remove Duo callout

dmoerner:

Following a Product Change discussion, we have expanded our allowed TOTP window to 30 seconds before and after the current 30 second window. This means that Duo Authenticator will work without trouble with Clerk, so we can delete the docs callouts about Duo.

I'm inclined to think it's not necessary for us to go into more detail about our current behavior, we can just clean up the Duo callouts.

### Deadline

No rush

### Other resources

Product change discussion: https://clerkinc.slack.com/archives/C0ADXF5HJBF/p1779292181963009
Backend PR: https://github.com/clerk/clerk_go/pull/19056


alexisintech (approved):
yup! perfect :) thank youuu

PR #3390, opened 2026-05-23
Document troubleshooting for Clerk JWT Template

pnhoang:
Added troubleshooting section for Clerk's JWT Template integration with Convex, clarifying the differences and setup steps.

### 🔎 Previews:

-

-

### Deadline

-

### DIFF
--- a/docs/guides/sessions/jwt-templates.mdx
+++ b/docs/guides/sessions/jwt-templates.mdx
@@ -263,3 +263,25 @@ The generated token would look like this:
   "sub": "user_1deJLArSTiWiF1YdsEWysnhJLLY"
 }
 ```
+
+# Troubleshooting
+
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


SarahSoutoul:
Hi Hoang, 

Thanks again for taking the time to put this PR together - we really appreciate the contribution.

We're going to close this one for now, not because the information is wrong, but because we want to keep the Convex docs centered on the new Clerk + Convex integration path. With the updated integration, most developers no longer need to create or manage a custom Clerk JWT template, so we're trying to guide users toward that default experience as clearly as possible.

There are still some more advanced or niche cases where a custom template can make sense, but we think adding that path into these docs right now would be more likely to create confusion for the majority of users than to help them.

Thanks again for the contribution - we'll keep this in mind if the opportunity arises for this addition.

PR #3396, opened 2026-05-27
Revamp test-emails-and-phones doc

alexisintech:
### 🔎 Previews:

- https://clerk.com/docs/pr/aa-docs-11775/guides/development/testing/test-emails-and-phones

### What does this solve?

[User feedback](https://linear.app/clerk/issue/DOCS-11775/clarify-email-link-testing-behavior-for-test-accounts) found that email links are in fact sent to a test email even though our docs currently state that when using a test email, no emails are sent. I tested this, and they are correct.

I also noticed the code examples were using the legacy custom flow API (`prepareFirstFactor`, `attemptFirstFactor`, `preparePhoneNumberVerification`, `attemptPhoneNumberVerification`), which is out of date. We have since updated our APIs with the Core 3 release.

### What changed?

- Replaced the outdated code examples with the latest Core 3 API
- Added pointers from each example to the matching custom flow guide (email/phone OTP) for readers who want the full UI flow, and to the custom flow guide overview for readers who want other flows.
- Reworked the intro so it mentions email verification links alongside OTPs (the doc already covers both, but the opener only referenced OTPs).

### Deadline

- No rush

### Other resources

- [Linear ticket](https://linear.app/clerk/issue/DOCS-11775/clarify-email-link-testing-behavior-for-test-accounts)

### DIFF
--- a/docs/guides/development/testing/test-emails-and-phones.mdx
+++ b/docs/guides/development/testing/test-emails-and-phones.mdx
@@ -1,6 +1,6 @@
 ---
 title: Test emails and phones
-description: Write end to end tests by simulating OTP verifications.
+description: Write end-to-end tests for email and phone verification flows, including OTPs and email verification links.
 ---
 
 Many of Clerk's sign-up, sign-in, and account update flows require verifying ownership of an email address or phone number, typically via a [one-time password (OTP)](!otp) or an email verification link. To confirm that your integration works correctly without sending real emails or SMS messages, you can simulate these verifications by using reserved test values in **test mode**.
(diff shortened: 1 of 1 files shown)


docs/guides/development/testing/test-emails-and-phones.mdx
```diff
@@ -3,9 +3,7 @@ title: Test emails and phones
 description: Write end to end tests by simulating OTP verifications.
```
SarahSoutoul: This is super minor but feels like the frontmatter could be updated to:

`Write end-to-end tests for email and phone verification flows, including OTPs and email verification links.`

Or can be shorter. But sth to represent the addition of email verification links, not just OTPs? @alexisintech

SarahSoutoul (approved):
Left a non blocking comment, so approving this!

docs/guides/development/testing/test-emails-and-phones.mdx
alexisintech: good point! i've updated it :)

PR #3399, opened 2026-05-28
docs(telemetry): add 'How you are notified' section

jacekradko:
clerk/javascript#8549 replaced the postinstall telemetry notice in `@clerk/shared` with a runtime notice gated to Node, non-CI, dev instances. Browser-only setups (`@clerk/clerk-js` direct, Vite SPAs on `@clerk/clerk-react` with no server, Chrome extensions) no longer get any in-band disclosure, so this page is now the authoritative one for them.

Adds a short "How you are notified" section after the intro that spells out the server vs. browser split and restates the dev-instances-only fact. No changes to the `telemetry` prop or opt-out instructions; existing partials and cross-links remain accurate as-is.


SarahSoutoul:
Closing this as changes migrated to new PR on clerk/clerk: https://github.com/clerk/clerk/pull/2973

PR #3404, opened 2026-05-29
Deprecate middleware-based auth gating: shared notice + multi-SDK protect-pages

jacekradko:
First in a stacked series moving Clerk away from middleware-based auth gating (`createRouteMatcher` and auth checks inside `clerkMiddleware`) toward protecting close to the resource. Background: [Deprecate middleware-based auth checks](https://www.notion.so/clerkdev/Deprecate-middleware-based-auth-checks-createRouteMatcher-35f2b9ab44fe80cdbd3ec54460b99cae).

Foundation only: a shared `_partials/middleware-auth-deprecation.mdx` that holds the deprecation wording in one editable place (it's included everywhere downstream), and `protect-pages` reworked into a multi-SDK guide. The new Next.js base covers page, route handler, Server Action, and layout protection, and flags that a layout check isn't a security boundary because layouts don't re-render on soft navigation. The existing Nuxt content moved verbatim to `protect-pages.nuxt.mdx`; it gets rewritten in the Nuxt PR.

One thing worth a look: the partial uses full-strength "deprecated" wording even though the SDK hasn't shipped `@deprecated` yet. It's isolated to that one file so softening it later is a one-line change.

PR #3405, opened 2026-05-29
Deprecate middleware auth gating across Next.js reference and guides

jacekradko:
Stacked on #3404. Applies the deprecation across the Next.js reference and the guides that taught middleware gating.

The reference page is kept-but-wrapped: each auth-gating section keeps its existing anchor but gains the deprecation notice, token-type gating now points to `auth({ acceptsToken })` at the route, and the Combine-Middleware example loses its `auth.protect()` line. In the guides, `basic-rbac` now leads with the page-level check, `session-tasks` elevates `<RedirectToTasks />`, `geo-blocking` and `satellite-domains` keep their middleware but drop the auth tail, the custom sign-in/up steps are framed as legacy, the CSP examples drop the auth gating that was never relevant to CSP, and the MCP guide stops special-casing `.well-known`.

Load-bearing bit to scrutinize: every surviving `createRouteMatcher` example moves from `/foo(.*)` to `/foo/:path*`. The `(.*)` form misses encoded newlines (a bypass) and over-matches `/foobar`; `/:path*` fixes both. Verified against the SDK's path-to-regexp.

### DIFF
--- a/docs/guides/ai/mcp/build-mcp-server.mdx
+++ b/docs/guides/ai/mcp/build-mcp-server.mdx
@@ -285,23 +285,7 @@ sdk: nextjs, expressjs
          export { handler as GET, corsHandler as OPTIONS }
          ```
        </CodeBlockTabs>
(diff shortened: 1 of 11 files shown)


jacekradko:

jacekradko:
Superseded by #3426, now the canonical Next.js `createRouteMatcher` deprecation PR. The migration guide and `@clerk/eslint-plugin` reference (from #3452) are merged in, and the `createRouteMatcher()` deprecation notices are in place. Closing in favor of #3426 (head `468fc4df4`), shipping coordinated with clerk/javascript#8994.

PR #3406, opened 2026-05-29
Deprecate createRouteMatcher in Nuxt docs in favor of built-in route protection

jacekradko:
Stacked on #3405. Moves the Nuxt docs off `createRouteMatcher` to Nuxt's built-in route protection.

`protect-pages.nuxt` now uses named route middleware (`definePageMeta` with file-system inheritance) and route groups (`route.meta.groups`, Nuxt 4.3+) instead of matching paths by hand, and the API-route examples in the middleware reference use `event.path` prefix checks. While rewriting the authorization example I also destructured `has()` from `event.context.auth()`, which was previously referenced but never defined.

### DIFF
--- a/docs/guides/secure/protect-pages.nuxt.mdx
+++ b/docs/guides/secure/protect-pages.nuxt.mdx
@@ -17,28 +17,55 @@ The `useAuth()` composable provides access to the current user's authentication 
 
 ## Protect multiple pages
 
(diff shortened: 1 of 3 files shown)


jacekradko:

wobsoriano (approved):

jacekradko:
Superseded by #3426. The Nuxt deprecation (`createRouteMatcher` to native `event.path` matching in the middleware reference, built-in route middleware for pages) is folded into #3426 (head `468fc4df4`). Thanks for the approval, @wobsoriano. Closing in favor of #3426.

PR #3407, opened 2026-05-29
Deprecate middleware auth gating in Astro docs

jacekradko:
Stacked on #3406. Wraps the Astro middleware reference's `createRouteMatcher` and protect-all sections with the deprecation notice and switches the examples to `:path*`.

Adds an Astro variant of `protect-pages` showing per-page `Astro.locals.auth()` guards. Astro has no batch route-protection mechanism, so the guide is upfront that guarding each page individually is the trade-off rather than pretending there's a one-liner.


alexisintech:
I've posted a docs review in a single commit, to make my suggestions more palatable to view (in a single clean diff)
The review folds the Astro middleware-auth deprecation cleanup further by removing the standalone `protect-pages.astro.mdx` guide that you created, and consolidating its useful bits into the existing `users/reading.astro.mdx` page.

Changes:

- Deleted `docs/guides/secure/protect-pages.astro.mdx`. Its content overlapped with `users/reading.astro.mdx`, which already shows the same isAuthenticated + redirect pattern.
- Moved the `requireAuth` helper into `docs/guides/users/reading.astro.mdx` as a new "Reduce repetition" subsection — the one piece worth preserving from the deleted page.
- Updated the deprecation partial (`_partials/middleware-auth-deprecation.mdx`) to point readers at `/docs/guides/users/reading` instead of the now-deleted `/docs/guides/secure/protect-pages`, and softened "affected" → "discouraged" for consistency.
- Trimmed `reference/astro/clerk-middleware.mdx` — removed the "Protect routes" and "Protect all routes" sections (auth-based, authorization-based, and invert-for-public-routes patterns) since those examples model the very anti-pattern the deprecation warns against. Reworded the remaining createRouteMatcher() example to frame it as matching routes that "need redirects" rather than "protected routes."
- Tiny cleanups: dropped email from the example response in `reading.astro.mdx` (since not all users will collect emails, but every single user has an ID), removed a stray target: '_blank' on an internal link, and changed "authentication and authorization" → "authentication" in the overview's middleware blurb.

Let me know what you think before I go reviewing and changing up the other PR's! @jacekradko

jacekradko:

jacekradko:
Superseded by #3426. The Astro deprecation, including your fold of `requireAuth` into `reading.astro` and the trimmed `clerk-middleware` reference, is consolidated into #3426 (head `468fc4df4`), with `createRouteMatcher()` marked deprecated now that `@clerk/astro` v3.4.8 ships the runtime warning. Thanks for the review, @alexisintech. Closing in favor of #3426.

PR #3409, opened 2026-05-29
feat(search): add active-SDK record field and codify the docs index settings

NWylynko:
### 🔎 Previews:

- https://clerk-git-nick-boost-active-sdk-search-results.clerkstage.dev/docs/pr/nick-include-guides-sdk-in-search

Docs search had no awareness of the reader's SDK, ranked body-content matches above title matches, and had no synonyms. This PR fixes that on the index side; the client-side boost is clerk/clerk#2661.

- **`sdk` record field** — tags each per-SDK variant of a page with the SDK it was built for, so the client's `optionalFilters: sdk:<active>` boost can pick the right variant under `distinct`.
- **Codified index settings** — `scripts/update-algolia-records.ts` is now the source of truth for the index's relevance levers (`searchableAttributes`, `attributesForFaceting` incl. `filterOnly(sdk)`, `ranking` with `attribute`/`exact` above `proximity`, `customRanking`, `attributeForDistinct`/`distinct`, synonyms). Declared and overwritten every run, so dashboard edits revert and every branch's records stay consistent.
- **Hybrid synonyms** — acronyms auto-derived from `docs/_tooltips/*` + a curated phrasing list (fixes `magic link`, `i18n`, `DKIM`, `login`, …).
- TypeScript-enforced setting keys (a typo or renamed field becomes a compile error), unit tests for synonym extraction, and fail-loud on unexpected tooltip read errors.

Detail lives in the thread rather than being repeated here:

- 📋 What's codified and why — [writeup](https://github.com/clerk/clerk-docs/pull/3409#issuecomment-4585030616) · [declare-and-overwrite follow-up](https://github.com/clerk/clerk-docs/pull/3409#issuecomment-4604115256)
- 📊 Why `filters` stays above `attribute`/`exact` — [ranking A/B, with before-after tables](https://github.com/clerk/clerk-docs/pull/3409#discussion_r3357753257)
- 🧭 Records are branch-scoped, settings are index-wide — [the tradeoff](https://github.com/clerk/clerk-docs/pull/3409#issuecomment-4606496657)

### Deadline

- 

### Other resources

- https://github.com/clerk/clerk/pull/2661


manovotny:
Beyond the per-variant `sdk` field, this branch now also **codifies the docs index's faceting, ranking, and synonyms** in `scripts/update-algolia-records.ts`, so they're enforced on every index run instead of living as mutable dashboard state.

What's in this PR specifically:

- `sdk` added to `attributesForFaceting` (`filterOnly`) — what makes the `optionalFilters: sdk:<active>` boost on the client side work at all.
- **Ranking reorder** enforced in the indexer — `attribute`/`exact` above `proximity`, so title/heading matches beat body-content matches.
- **Hybrid synonyms** — acronyms auto-derived from the `_tooltips` glossary + a curated phrasing list (fixes `magic link`, `i18n`, `DKIM`, `login`, …).
- `AGENTS.md`: don't tune ranking/synonyms/faceting in the Algolia dashboard — they're codified here and revert on every reindex.

Paired with the client-side active-SDK boost in clerk/clerk#2661. The full data-backed writeup — why boost `sdk` over `availableSDKs`, the ranking reorder, the synonym set, and every rejected alternative with measurements — lives there: https://github.com/clerk/clerk/pull/2661#issuecomment-4585030268

manovotny (approved):

scripts/update-algolia-records.ts
```diff
@@ -747,6 +830,61 @@ async function main() {
...
+    await algolia.setSettings({
+      indexName: ALGOLIA_INDEX_NAME,
+      indexSettings: {
+        attributesForFaceting: [
+          ...attributesForFaceting,
+          ...missingFacets.map((attribute) => `filterOnly(${attribute})`),
+        ],
+        ranking,
+      },
+    })
```
NWylynko:
```suggestion
    await algolia.setSettings({
      indexName: ALGOLIA_INDEX_NAME,
      indexSettings: {
        attributesForFaceting: [
          ...attributesForFaceting,
          ...missingFacets.map((attribute) => `filterOnly(${attribute})`),
        ],
        ranking,
      },
      forwardToReplicas: true,
    })
```

Image

@manovotny should the setSettings also forward the change to replicas. Granted I don't think we are using replicas, but should be consistent I think.

manovotny:
Follow-up: the indexer now codifies the docs index's relevance settings — declares them and overwrites on every run. d7d395576, 9d61ccc8f

Acting on review feedback: instead of read current settings → diff → additively merge each run, the indexer now declares the settings it owns and overwrites them every run, so any dashboard edit reverts on the next reindex. The additive facet merge actually worked against the goal — it *preserved* whatever was set in the dashboard rather than overwriting it, so indexes could (and did) drift apart.

```ts
const searchableAttributes = ['unordered(hierarchy.lvl0)', /* …lvl1–6 */ 'content', 'unordered(keywords)']
const attributesForFaceting = ['branch', 'record_batch', 'sdk'].map((a) => `filterOnly(${a})`)
const ranking = ['typo', 'geo', 'words', 'filters', 'attribute', 'exact', 'proximity', 'custom']
const customRanking = ['desc(weight.pageRank)', 'desc(weight.level)', 'asc(weight.position)']
await algolia.setSettings({ indexName: ALGOLIA_INDEX_NAME, indexSettings: { searchableAttributes, attributesForFaceting, ranking, customRanking } })
```

**Why these four and not the whole settings object** (i.e. we don't codify defaults — correct): `setSettings` is a top-level partial merge, so we declare only the relevance levers we deliberately own and leave everything else (typo tolerance, pagination, highlighting, and Algolia's server-managed defaults) untouched. Snapshotting the full object would freeze those defaults at snapshot time and silently clobber any future default change or intentional tweak — diff noise and maintenance for zero benefit. Synonyms stay generated from `_tooltips` (already a full replace).

**The two added in `9d61ccc8f`:**
- **`searchableAttributes`** — the corpus + attribute priority the `attribute` ranking criterion rides on (it's what makes a heading match beat a body match, so the reorder depends on it). Identical across dev/prod today, so this locks in the value the ranking work was tested against rather than freezing an unknown state.
- **`customRanking`** — the final tiebreaker, normalized to the weights the indexer actually writes (`pageRank`/`level`/`position`). Drops the dead `desc(weight.popularity)` entry lingering on `dev_docs` — it's never written to records, so it ranked nothing.

**Why `availableSDKs` is intentionally not faceted** — it powers the per-result SDK icons, but the client only *retrieves* it (`Search.tsx` → `SDKsIcon`); it never filters or counts on it, and retrieval is independent of faceting. So the `availableSDKs` facet that existed on `dev_docs` was inert — dropping it doesn't touch the icons. `branch`/`record_batch` are `filterOnly` because the stale-record cleanup filters via `facetFilters`, never facet counts.

**Effect on the next index run:** adds `sdk` faceting everywhere, normalizes `branch`/`record_batch` to `filterOnly`, drops the stray `availableSDKs` facet and the dead `customRanking` entry on `dev_docs`, and locks `searchableAttributes`/`ranking` to the tested values — leaving prod/dev/test with an identical, code-defined config.

A "warn if the dashboard was edited" guard was considered and deferred: since the indexer rewrites these every run, drift self-heals on the next index, so an explicit alert is more than this needs right now.

scripts/update-algolia-records.ts
manovotny: Good Q — went back and forth on this, landed on: no, `setSettings` shouldn't forward.

We're codifying these settings (declare + overwrite every run), which is itself the mechanism for keeping indexes consistent. `forwardToReplicas` is the *other* model — "tune the primary, propagate it" — so under codification it's redundant: the script configures whatever index it targets directly.

It's also the riskier default here specifically because this `setSettings` bundles `ranking`/`customRanking`. A standard replica usually exists to hold a *different* sort, and `forwardToReplicas: true` would overwrite that on every index run. So if we ever add replicas, the right move is to declare them in the script (which can express per-replica settings), not blanket-forward.

The synonyms call *does* forward, and that's deliberate — synonyms should always be identical across replicas, so it's a safe no-op today and the correct behavior if a replica ever shows up. Added a comment in efe7c945e spelling out the asymmetry so it doesn't read like an oversight.

Two notes: there are no replicas on `dev_docs`/`prod_docs` today, so this is all forward-looking; and heads up the suggestion is anchored to the pre-refactor block (the read/merge/`missingFacets` code is gone as of d7d395576 / 9d61ccc8f), so it can't be committed directly.

NWylynko:
@manovotny 

With adding the settings update to the script it creates a bit of split understanding of the script. Before the settings updates was added, the script would entirely scope itself to the runners current branch, including the branch name in the records that where being added, allowing the search client to scope itself to that branch and know that its only getting that set of results.

But when updating the index settings its across all records in that index, not just the records that are scoped to the branch that is being updated. IMO this introduces a split in how you'd expect the script to run, either not caring about the git branch and just having a single 'dev_docs' index, or that changes made only effect the search items being updated and would expect settings would only change for that branch too.

This stems from a limitation of algolia ultimately and a desire on my part to not spin up a whole new index on every search development branch. So if this is a trade-off we are happy with then I'm happy to move ahead, but I haven't seen this addresses so just want to get your opinion.

manovotny:
Totally fair catch — it's the sharpest point on the thread, and you're right that it was undocumented.

The way I think about it: records and settings are genuinely different _kinds_ of thing, so the two scopes are each correct rather than inconsistent. 

* Records are data — rightly per-branch (the `branch:` tag + client filter is what lets us share one index).
* Settings/synonyms are the index's relevance _contract_ — a property of the index, not of a branch's content. 

Algolia has no per-branch settings, and conceptually there shouldn't be: "how this index ranks" isn't branch-specific.

Codifying also shrinks the downside you're pointing at. 

Before, a run could push arbitrary hand-tuned settings index-wide; now every run converges the index to the same declared values, and any change to them is a reviewed diff. So a content branch's indexer run just re-asserts canonical settings — nothing leaks or drifts. 

The only gotcha is a _search-relevance_ branch (rare — like this one) that's actively changing settings, and the escape hatch already exists (like I used here): point `ALGOLIA_INDEX_NAME` at a personal throwaway index. You pay for an extra index only while experimenting with settings, not per branch.

Full per-branch isolation is the "pure" answer — like we discussed before, if usage/cost weren't a factor, we'd spin up an index per branch and call it done. But that's a lot of indexes (and money) for a rare need, so shared `dev_docs` + branch-scoped records + a personal index for settings experiments is the middle ground.

So — accepted trade-off? Agreed. And to your "haven't seen this addressed": it is now, in f85cd1257 — a scope note in the settings block, plus a new "Search index (Algolia)" section in `AGENTS.md` spelling out the split, the codified settings, and the personal-index escape hatch.

scripts/update-algolia-records.ts
```diff
@@ -747,6 +880,101 @@ async function main() {
...
+  // SDK icons (Search.tsx `SDKsIcon`), never filters or counts on it, and retrieval is independent
+  // of faceting.
+  const attributesForFaceting = [filterOnly('branch'), filterOnly('record_batch'), filterOnly('sdk')]
+  //
+  // Ranking: `attribute`/`exact` are moved above `proximity` (vs Algolia's default order) so a
+  // query that matches a page's title/heading outranks one that only matches the same words in
+  // body content. e.g. for "sign in page" the "Build your own sign-in page" guide and the
+  // Quickstart surface above reference pages whose body merely mentions the phrase, while
+  // reference-name queries like `auth()` still win on their exact title match. See the design doc.
+  const ranking = ['typo', 'geo', 'words', 'filters', 'attribute', 'exact', 'proximity', 'custom']
```
NWylynko, line 938:
```suggestion
  const ranking = ['typo', 'geo', 'words', 'attribute', 'exact', 'filters', 'proximity', 'custom']
```

@manovotny 

I think we want to push 'filters' down the ranking so that title/heading matches (the `attribute` and `exact` criteria) take priority over it. filters is normally high because it's usually a boolean "is/isn't in the filter" signal — but here we're feeding it the active-SDK optionalFilters, which is more of a soft suggestion than a hard constraint. Demoting it also lets generic guides rank on their own relevance alongside SDK-scoped guides, instead of the SDK boost outranking a generic guide's title match.

manovotny: Tested both orders on a throwaway copy of `dev_docs` (active SDK = Next.js, `distinct`, on a branch with `sdk` populated). Full sweep — `filters↑` is current, `filters↓` is your suggestion:

| query | filters↑ (current) | filters↓ (suggested) |
|---|---|---|
| `auth()` / `clerkMiddleware` / `useUser` | nextjs #1, all nextjs | identical |
| `sign in page` | nextjs (SignIn, Quickstart, build-your-own) | identical |
| `custom sign up flow` | all nextjs | identical |
| `currentUser` | nextjs through #6 | #3–6 → **astro** |
| `protect a route` | nextjs through #6 | #4–6 → **go / nuxt** |
| `UserButton` | #1–6 nextjs `<UserButton/>` | #1 **ios**, #2–6 **android** |
| `OrganizationSwitcher` | #1–4 nextjs | #5–6 → **js-frontend** |
| `webhooks` | nextjs billing | #5–6 → overview / sync (slightly better) |

Every change is the same shape: demoting `filters` lets a stronger literal title match from another SDK outrank the active SDK's page. `UserButton` is the clearest:

| rank | filters↑ (current) | filters↓ (suggested) |
|---|---|---|
| 1 | `[nextjs]` `<UserButton />` component | `[ios]` UserButton |
| 2 | `[nextjs]` `<UserButton />` › Example | `[android]` UserButton › Parameters |
| 3 | `[nextjs]` `<UserButton />` › Properties | `[android]` UserButton › Usage |

The iOS/Android pages are titled exactly `UserButton` vs our `<UserButton /> component`, so once `filters` sits below `exact` they win — a Next.js user searching "UserButton" gets iOS/Android docs.

The queries we actually care about (`sign in page`, `custom sign up flow`, the reference names) are identical either way, and references stay #1 in both — so there's no offsetting upside. You're right that `optionalFilters` is a soft signal, but on this corpus keeping `filters` high is what prevents the cross-SDK bleed. The one spot your order edges ahead (universal pages like "webhooks overview") is narrow enough I'd chase it via content/synonyms instead. 

Leaning toward keeping `filters` where it is.

NWylynko: Thank you for the thorough testing!

NWylynko:
I'm ready to merge this @manovotny, since I've made changes do you wanna approve it again / reply that your happy with it?

manovotny (approved):
I'm ready to merge too. 🤝

PR #3422, opened 2026-06-04
Update Android native view dismiss prop docs

swolfand:
## What changed

- Updates Android `AuthView` docs to include the current dismiss-related Kotlin parameters from `clerk/clerk-android#710`, including `isDismissible` and `onDismiss`.
- Updates Android `UserProfileView` docs for optional `onDismiss` and `isDismissible` behavior.
- Adds an Expo native `<AuthView />` usage example for `isDismissable`, alongside the existing `mode` examples.

## Notes

Current `main` no longer contains the standalone Organization view reference pages, so this clean branch does not reintroduce those deleted pages.

## Validation

- `pnpm run build:tsx`
- `pnpm run lint`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/docs/reference/expo/native-components/auth-view.mdx
+++ b/docs/reference/expo/native-components/auth-view.mdx
@@ -53,6 +53,8 @@ export default function SignInScreen() {
 
 ### Dismissible auth view
 
+The following example enables the native dismiss button with the `isDismissable` prop.
+
 ```tsx
 <AuthView isDismissable />
 ```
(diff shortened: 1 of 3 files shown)


manovotny:
Pushed some changes directly in 03b189135.

- Reworded the Android `AuthView` `modifier` description — it said the modifier is applied to the top-level `NavDisplay`, but in source the modifier is composed into the outer container (`NavDisplay` gets a fixed `fillMaxSize()`), so I changed it to "the top-level container of the `AuthView`."
- Added a one-line intro before the Expo "Dismissible auth view" example to match the styleguide's "explain before the code block" rule.
- Added the same intro to the Android "Custom dismiss handling" example (its sibling "Non-dismissible auth" already had one).
- Added an intro to the "Non-dismissible profile view" example, matching the Android `AuthView` phrasing.

manovotny (approved):

PR #3426, opened 2026-06-08
Recommend protecting auth close to the resource

jacekradko:
Stops recommending middleware as the place to gate auth and points protection at the resource that reads the data (the page, Route Handler, Server Action, or endpoint). The copyable `createRouteMatcher()` + `auth.protect()` / `!isPublicRoute` allowlist examples come out of the guides and references; `clerkMiddleware()` stays documented for its non-auth uses (redirects, headers, geo, route matching).

This began as Next.js-only and now also folds in the Astro, Nuxt, and Core-2 upgrade-guide versions that were open as separate PRs (#3404-3408, now closed). The old pattern survives on purpose in the renamed canonical guide `protect-content` (was `protect-pages`), which gains a "migrating from middleware-based auth checks" walkthrough: before/after, the `requireUser()` / `requireAuth()` helper, and the add-first-remove-last ordering for inverting `!isPublicRoute`. That guide and the per-SDK `reading.*` pages that host the helper are the content worth real review.

The deprecation wording is deliberately uneven and worth checking: Next.js and Astro call `createRouteMatcher()` "deprecated" because their SDKs log a runtime warning (Astro shipped it in v3.4.8; Next.js ships it in clerk/javascript#8994), while Nuxt only says "no longer recommended" since its SDK hasn't deprecated the helper. Because of the Next.js side, this has to merge together with clerk/javascript#8994, or the docs assert a deprecation that isn't live yet.

One separable rider remains if you'd rather split it: the `use-auth` reference/partial consolidation.

- protect-content: https://clerk.com/docs/pr/3426/guides/secure/protect-content
- Migration guide: https://clerk.com/docs/pr/3426/guides/development/upgrading/upgrade-guides/migrating-from-create-route-matcher
- Next.js clerk-middleware: https://clerk.com/docs/pr/3426/reference/nextjs/clerk-middleware
- Astro clerk-middleware: https://clerk.com/docs/pr/3426/reference/astro/clerk-middleware
- Nuxt clerk-middleware: https://clerk.com/docs/pr/3426/reference/nuxt/clerk-middleware


docs/guides/ai/mcp/build-mcp-server.mdx
```diff
@@ -285,23 +285,7 @@ sdk: nextjs, expressjs
...
-         '/.well-known/oauth-authorization-server(.*)',
-         '/.well-known/oauth-protected-resource(.*)',
-       ])
-
-       export default clerkMiddleware(async (auth, req) => {
-         if (isPublicRoute(req)) return // Allow public access to .well-known endpoints
-         await auth.protect() // Protect all other routes
-       })
-       ```
+    1. Your `.well-known` endpoints must be **publicly accessible** for MCP clients to discover OAuth metadata.Ensure this route is not protected anywhere.
```
alexisintech:
```suggestion
    1. Your `.well-known` endpoints must be **publicly accessible** for MCP clients to discover OAuth metadata. Ensure this route is not protected anywhere.
```

alexisintech:
I've posted my docs review. I actually went with keeping the "Protect pages" guide but changed it to "Protect content"
The existing "Protect content and read data" (`/users/reading`) was slightly adjusted to focus more on reading data. I've renamed the guide "Read data". This guide now points to the "Protect content" guide if users want more detailed examples.

The commit is huge, if you have any questions, we can hop on a call or chat async!

docs/guides/configure/session-tasks.mdx
```diff
@@ -63,79 +69,41 @@ export default function Page() {
...
-  If you'd like to simply redirect users to the sign-in page if they are signed-out, you can [use `auth.protect()`](#using-auth-protect). If you'd like to have more control over what your app does based on user authentication status, you can [use the `isAuthenticated` property](#using-is-authenticated).
-</If>
-
-<If sdk="nextjs">
-  #### Using `auth.protect()`
+<If sdk={["nextjs", "astro", "nuxt", "react-router", "tanstack-react-start"]}>
+  ### Using middleware-based redirects
 
-  When using `auth.protect()` in middleware to protect routes, it will redirect users to the sign-in page if they are signed-out. In the following example, `pending` users will be redirected to the sign-in page, where the `<SignIn />` component will prompt them to fulfill the session tasks. Once finished, their session will move from `pending` to an `active` (signed-in) state.
+  The [`createRouteMatcher()`](/docs/reference/nextjs/clerk-middleware#create-route-matcher) helper function can be used in your `clerkMiddleware()` to protect specific routes or route groups from users who have pending session tasks.
```
Ephem: This whole usecase for `createRouteMatcher` is going away when we deprecate right? So we need to rewrite this to point to protecting close to the resource here as well when we do?

docs/guides/development/geo-blocking.mdx
```diff
@@ -17,9 +17,8 @@ While Clerk does not provide a geo blocking feature, deployment platforms, such
     import { geolocation } from '@vercel/functions'
     import { NextResponse } from 'next/server'
 
-    // Define your protected and blocked routes
-    const isProtectedRoute = createRouteMatcher(['/dashboard(.*)'])
-    const isBlockRoute = createRouteMatcher(['/block(.*)'])
+    // Define the blocked routes
+    const isBlockRoute = createRouteMatcher(['/block/:path*'])
```
Ephem: This is also going to go away.

I wonder what we should do for any code like this that _has_ to stay in the middleware? 🤔

Others just document e.g. `request.nextUrl.pathname.startsWith('/block/')` which I guess makes sense?

docs/guides/secure/basic-rbac.mdx
Ephem, line 89: Shouldn't we migrate this to use `auth.protect()` instead?

```suggestion
// Protect the page from users who are not admins
await auth.protect({ role: 'admin' });
```

If we want to show custom handling, other examples could have `const { has } = await auth.protect();` and then check with `has({ role: 'admin' })`?

Am I missing some reason we need the `Create a reusable function to check Roles` section?

The reason I'm asking is that it's hard to lint custom functions like `checkRole`, so moving to using `auth.protect()` instead would satisfy the lint rule without us having to document anything awkward.

docs/guides/secure/protect-content.nuxt.mdx
```diff
@@ -1,18 +1,16 @@
...
 > [!TIP]
 > To learn how to protect API routes, see the [dedicated guide](/docs/reference/nuxt/clerk-middleware#protect-api-routes).
 
 ## Protect a single page
 
-The `useAuth()` composable provides access to the current user's authentication state and methods to manage the active session. It also includes the `isSignedIn` property to check if the active user is signed in, which is helpful for protecting a page.
-
 <Include src="_partials/vue-nuxt/use-auth" />
 
 ## Protect multiple pages
```
Ephem, line 16: Since there are other ways to protect multiple pages in Nuxt, we could update this section ahead of deprecation. That seems like a separate PR though.

docs/guides/users/reading.mdx
```diff
@@ -1,28 +1,31 @@
...
+
+> [!TIP]
+> This guide briefly touches on protecting content from unauthenticated users. For more detailed examples, see the [Protect content from unauthenticated users](/docs/guides/secure/protect-content) guide.
 
 ## Server-side
 
 ### App Router
 
-[`auth()`](/docs/reference/nextjs/app-router/auth) and [`currentUser()`](/docs/reference/nextjs/app-router/current-user) are App Router-specific helpers that you can use inside of your Route Handlers, Middleware, Server Components, and Server Actions.
+[`auth()`](/docs/reference/nextjs/app-router/auth) and [`currentUser()`](/docs/reference/nextjs/app-router/current-user) are App Router-specific helpers that you can use inside of your Route Handlers, Server Components, and Server Actions.
```
Ephem, line 18: NIT: I don't think we need to remove "Middleware" from this list, since they can still technically be used there, we just don't recommend it for auth gating.

docs/guides/users/reading.mdx
```diff
@@ -31,12 +34,14 @@ The following example uses the [`auth()`](/docs/reference/nextjs/app-router/auth
...
       const { isAuthenticated } = await auth()
 
-      // Protect the route by checking if the user is signed in
+      // Protect the page from unauthenticated users
       if (!isAuthenticated) {
-        return <div>Sign in to view this page</div>
+        // Add logic to handle the unauthenticated user
+        // This example renders a UI but you could also redirect to the sign-in page instead
+        return <div>You must be signed in to view this page</div>
       }
```
Ephem: I wonder when we should show this long form version of protecting, vs just documenting `await auth.protect()`?

I think maybe `await auth.protect()` is a good default, it usually does the right thing and let's the reader focus on the important parts, but how do we also show/link to how you take more control over the experience?

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,268 +36,30 @@ export const config = {
...
-
-- Use [`auth.protect()`](/docs/reference/nextjs/app-router/auth#auth-protect) if you want to redirect unauthenticated users to the sign-in route automatically.
-- Use [`auth().isAuthenticated`](/docs/reference/nextjs/app-router/auth#protect-pages-and-routes) if you want more control over what your app does based on user authentication status.
-
-<CodeBlockTabs options={["auth.protect()", "auth().isAuthenticated()"]}>
-  ```tsx {{ filename: 'proxy.ts' }}
-  import { clerkMiddleware, createRouteMatcher } from '@clerk/nextjs/server'
-
-  const isProtectedRoute = createRouteMatcher(['/dashboard(.*)', '/forum(.*)'])
+Use the returned matcher inside `clerkMiddleware()` for non-auth request handling, such as applying a redirect, header, or geo rule to a group of routes. For a worked example, see [Geo blocking](/docs/guides/development/geo-blocking).
```
Ephem:
```suggestion
Use the returned matcher inside `clerkMiddleware()` for non-auth request handling, such as applying a redirect, header, or geo rule to a group of routes. For an example, see [Geo blocking](/docs/guides/development/geo-blocking).
```

docs/reference/nuxt/clerk-middleware.mdx
```diff
@@ -7,7 +7,7 @@ sdk: nuxt
 The `clerkMiddleware()` helper allows you to protect your Nuxt application **on the server-side**. It can be used to validate a user's authentication status or authorization status.
 
 > [!WARNING]
-> `clerkMiddleware()` should be used to protect API routes only. It's not recommended to use `clerkMiddleware()` to protect pages as it will **only work on initial page reload**. On subsequent navigations, it won't be triggered because client-side navigation will bypass the middleware. To protect pages, see the [dedicated guide](/docs/guides/secure/protect-pages).
+> `clerkMiddleware()` should be used to protect API routes only. It's not recommended to use `clerkMiddleware()` to protect pages as it will **only work on initial page reload**. On subsequent navigations, it won't be triggered because client-side navigation will bypass the middleware. To protect pages, see the [guide on protecting content from unauthenticated users](/docs/guides/secure/protect-content).
```
Ephem: Note: We are going to need a doc that outlines how to protect API routes without `createRouteMatcher` for Nuxt.

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+      return <h1>Hello world</h1>
+    }
+    ```
+  </Tab>
+</Tabs>
+
+### Protect a layout
+
+It's not recommended to perform authentication checks in a layout. Layouts don't re-render on client-side navigation, so the session isn't re-checked on every route change. For more information, see the [Next.js docs](https://nextjs.org/docs/app/building-your-application/authentication#layouts-and-auth-checks).
```
Ephem: This is wrong. Since you can access privileged resources in a layout as well (see prefetching in the dashboard for example), you _also_ have to protect layouts. This is correct in that you can't _only_ protect layouts though.

I think I'd merge layouts+pages and make a note that it's not enough to protect only the layouts.

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+      // Add your Server Action logic using the `userId`
+    }
+    ```
+  </Tab>
+</Tabs>
+
+## Client-side
+
+You can use the `isSignedIn` property from the [`useAuth()`](/docs/reference/hooks/use-auth) or [`useUser()`](/docs/reference/hooks/use-user) hook to protect your Client Components from unauthenticated users.
```
Ephem: I think we need to make it clear that this is not actually _protecting_ anything. It's potentially already rendering on the client when you hit this code, so everything is going to be available to an attacker anyway. This is more of a UX thing really.

The way to protect the client is to protect:

* The resources that are passed down to client components as props
* The external resources the client calls, like API endpoints

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,260 @@
...
+title: Protect content from unauthenticated users
+description: Learn how to protect pages, Route Handlers, and Server Actions from unauthenticated users in your Clerk + Next.js application.
+sdk: nextjs
+---
+
+This guide demonstrates how to protect pages, Route Handlers, and Server Actions from unauthenticated users in your Next.js application.
+
+> [!TIP]
+> To check whether a user is [**authorized** (has a specific Role, Permission, Feature, or Plan)](!authorization-check) in addition to being signed in, see [Authorization checks](/docs/guides/secure/authorization-checks). This guide focuses on **authentication** (whether a user is signed in).
+
```
Ephem, line 10: I think this page should start with a checklist. To protect your application, you need to protect the following resources:

* Server Components
  * Layouts, Pages etc
* Route handlers
  * Internal Next API endpoints
* Server Actions
  * We need to explain these are also API endpoints
* External API endpoints
  * This one is missing in a huge way here

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,260 @@
...
+      // This example renders a UI but you could also redirect to the sign-in page instead
+      return <h1>You must be signed in to view this page</h1>
+    }
+
+    return <h1>Hello world</h1>
+  }
+  ```
+</CodeBlockTabs>
+
+## Migrating from middleware-based auth checks
```
Ephem: This section plus removing everything from the `clerkMiddleware` page is essentially the same as deprecating it now, just without the actual deprecation.

I think we should save these parts for later, this migration guide is going to take a lot of work to get right, and while this is a good foundation, I think we need some better support like the lint rule and codemod for this to be truly complete.

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,268 +36,30 @@ export const config = {
...
 
-  export default clerkMiddleware(async (auth, req) => {
-    if (isProtectedRoute(req)) await auth.protect()
-  })
+## Protecting routes
```
Ephem, line 39: I think we should keep this section for now, maybe with a "we recommend guarding resources close to where they are used, see this guide for how" at the top?

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+**Protect access as close to the resource as possible**, where you read or mutate the data. This way, the same code that touches the data also enforces who can reach it.
+
+If you're currently protecting routes in `clerkMiddleware()`, move the auth check into the resource itself and remove it from `clerkMiddleware()`. **You still need to keep `clerkMiddleware()` as it's required for Clerk configuration and can still be used for things like redirects, headers, and other request handling.** Cross-cutting redirects (incomplete [session tasks](/docs/guides/configure/session-tasks), [onboarding](/docs/guides/development/add-onboarding-flow)) can stay in middleware as a redirect convenience, as long as the real enforcement also runs at the resource. For [checking authorization](!authorization-check) (Roles, Permissions, Features, or Plans), see the [Authorization checks](/docs/guides/secure/authorization-checks) guide.
```
Ephem: > Cross-cutting redirects (incomplete [session tasks](/docs/guides/configure/session-tasks), [onboarding](/docs/guides/development/add-onboarding-flow)) can stay in middleware as a redirect convenience, as long as the real enforcement also runs at the resource. 

Can it though? If those can stay (that are essentially auth guards too), why can't other auth guards stay? This is something we need a better grasp on for sure, I hadn't considered these cases tbh, let's discuss!

Ephem (commented):
I left a bunch of specific comments. I didn't go through everything in detail, as my biggest piece of feedback is that I think this PR is trying to do too much for a first PR that we can release now. It's currently kind of soft-deprecating the `createRouteMatcher` without enough support to actually migrate away from it.

I think ~~this~~ the first PRs should:

* Separate PR - Change all relevant `/(.*)` to `/:path*`
* Add content and guidance on how to protect close to resources
* Downplay `createRouteMatcher` and middleware gating in a few places and link to that new content instead
* Not _remove_ anything around `createRouteMatcher` or try to guide people towards migrating away from it at this point

I think a lot of the content is great, just that some of it is a bit early to release just yet?

docs/guides/secure/basic-rbac.mdx
jacekradko: Yes! I think that's the pattern we actually want people to use

docs/guides/development/geo-blocking.mdx
jacekradko: **TANGENT** >> Geo-blocking is actually the kind of case I'd say createRouteMatcher fits well. It's path-based branching, not auth, so it doesn't move to the resource. The line we're drawing is auth vs non-auth, not matcher vs no matcher: deciding who can access something moves to the resource; doing different things on different paths (geo, redirects, headers) stays in the proxy.

Ideally you'd compose proxies, separate handlers for separate paths so there would be no need for a duplicate `createRouteMatcher`. Unfortunately that's a want-to-have, not what the Next gives us today.

With all of that said, I think we just move to native Request so there is no ambiguity here

Ephem: Yeah, one could argue bypassing geo-_blocking_ is also bad, but it is less bad than auth.

If we had only resource-level protections today, so all apps were built that way, I don't think introducing some kind of path matching as an optimization would be a terrible idea, though still one _Next_ should implement, not us. 😅

> I think we just move to native Request so there is no ambiguity here

🤝

docs/guides/secure/basic-rbac.mdx
jacekradko: Good call on dropping the helper, removed it. One catch: the replacement can't be `auth.protect({ role })` / `has({ role })`. basic-rbac keeps the role in `publicMetadata` (no-Organizations workaround) and `has` checks `orgRole`, not metadata, so those would deny everyone here. So each spot now reads the claim inline (`sessionClaims.metadata.role`) instead.

Ephem: Ah, TIL, thank you! 🙏

docs/guides/users/reading.mdx
jacekradko: Re-added it

docs/guides/configure/session-tasks.mdx
alexisintech: not necessarily. you can still use it to group routes. 
for example, in this section, the middleware is redirecting users who have pending session tasks to a page where they can complete those session tasks. the createRouteMatcher() helper is used to set the routes where a user with pending session tasks (essentially an incomplete/signed out user) should not be able to visit.
it's used in other places in the docs as well

Ephem: I'm not quite sure I follow. Are you arguing we keep the `createRouteMatcher` around for these kind of use cases? Because if so that's something we've already decided against since it's not a strong enough push to push people of these vulnerable patterns.

> should not be able to visit

Isn't this also a form of auth gating, and therefor something something devs need to do close to the resources as well?

alexisintech: for other readers, this discussion was resolved in slack but essentially the decision was made to remove createRouteMatcher() altogether. "if we keep it around in any fashion people wont migrate away from it" and its best to minimize the chances for user error!

alexisintech:
Just posted another docs review, here's some guidance into the reasons for my changes:

- Removed `_partials/middlewareare-auth-deprecation` as it wasn't being used anywhere.
- The "Migrating from middleware-based auth checks" walkthrough now lives in the clerk middleware reference instead of in the `protect-content` guide. The reasoning: anyone looking to protect content in clerk middleware is reading the middleware reference, so the deprecation note and the migration away from protecting content in middleware belong together there. the `protect-content` guide then stays focused on the resource-level patterns it's named after.
- Custom sign-in / sign-up pages: replaced the trailing [!NOTE] callout with a `## Make the route public step`, pointing at the relocated migration section in the clerk middleware reference. Reads as a normal setup step rather than a footnote in the "Visit your new page" section.
- `protect-content` guide intro: moved bullet list of what needs protection (Server Components, Route Handlers, Server Actions, external endpoints) before the authorization-vs-authentication tip, so the scope lands first.
- `app-router/auth reference`: broke auth.protect()'s "Example" and auth()'s "Verify machine requests" sections into bullets covering signed-in, authz, and machine tokens (OAuth, M2M, API keys), so the reference enumerates the cases **instead of one-lining to a single guide that didn't cover all of the cases**.
- `session-tasks` example: renamed requiresSessionTasks → routeRequiresSessionTasks so it's clear the boolean is route-scoped.

docs/guides/users/reading.mdx
alexisintech: good idea for simplification! done here [34fbc74](https://github.com/clerk/clerk-docs/pull/3426/commits/34fbc7406b3b9d7bcb8c913f9e12efffd1cab4ed)

docs/guides/secure/protect-content.mdx
alexisintech: ahhh interesting!! TIL.
fixed here by jacek! [15d7992](https://github.com/clerk/clerk-docs/pull/3426/commits/15d79927f7c5750d994994466b4ba5ee8b0298d9)

docs/guides/secure/protect-content.mdx
alexisintech: fixed here by jacek! [15d7992](https://github.com/clerk/clerk-docs/pull/3426/commits/15d79927f7c5750d994994466b4ba5ee8b0298d9)

docs/guides/secure/protect-content.mdx
alexisintech: fixed here by jacek! [15d7992](https://github.com/clerk/clerk-docs/pull/3426/commits/15d79927f7c5750d994994466b4ba5ee8b0298d9)

docs/reference/nextjs/clerk-middleware.mdx
alexisintech: worked on this section here [6f44324](https://github.com/clerk/clerk-docs/pull/3426/commits/6f44324be9624b2726999c8584d7db804dffcf86)
let me know how you feel about it!

docs/guides/secure/protect-content.mdx
Ephem: The Next.js model is very tricky to get a 100% right, and it's hard to catch when you do it wrong which makes our job harder. 😬

That's one reason we want to encode it in a lint rule too.

alexisintech:
@jacekradko did that get changed? it used to be 404

Screenshot 2026-06-11 at 10 35 28

if it got changed I can sweep the docs and make the update! just wanna check (I can also test too to be sure)

alexisintech:
actually, I just tested a server action that uses auth.protect(), was unauthenticated and it redirected me to the sign in page! didn't return a 404 or 401!

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+      // Add your Server Action logic using the `userId`
+    }
+    ```
+  </Tab>
+</Tabs>
+
+## Client-side
+
+On the client, you can read the `isSignedIn` property from the [`useAuth()`](/docs/reference/hooks/use-auth) or [`useUser()`](/docs/reference/hooks/use-user) hook to conditionally render UI based on whether the user is signed in.
```
Ephem, line 151: I wonder if we should also mention the `<Show>` component here?

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+> [!WARNING]
+> A client-side check is a UX affordance, not a security boundary. By the time this code runs, the component has already been sent to the browser, so anything it contains is available to the user regardless of the check. Protect the data instead: guard the resources you pass to Client Components as props, and the Route Handlers or external API endpoints the client calls (see [Server-side](#server-side)).
```
Ephem: I know this is pretty verbatim from what I said, but reading this part:

> Protect the data instead: guard the resources you pass to Client Components as props

I realize it's a bit fuzzy, it's not very clear guidance. This is essentially already what you are doing by protecting the server components.

Maybe something like this?

```suggestion
> A client-side check is a UX affordance, not a security boundary. By the time this code runs, the component has already been sent to the browser, so anything it contains is available to the user regardless of the check. 
>
> Protect the data instead: guard Route Handlers, Server Functions and external API endpoints the client calls, and by protecting the Server Components, you also guard any resources passed down as props to the client (see [Server-side](#server-side)).
```

docs/guides/secure/protect-content.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+      return <h1>Hello world</h1>
+    }
+    ```
+  </Tab>
+</Tabs>
+
+### Protect a layout
+
+You can run an auth check in a layout the same way as in a page, and you should if the layout itself reads privileged data (for example, when prefetching). But a layout check alone isn't enough: layouts don't re-render on client-side navigation, so the session isn't re-checked on every route change. Protect each page (and every other resource that reads or mutates data) directly, and treat a layout check as an addition, not a replacement. For more information, see the [Next.js docs](https://nextjs.org/docs/app/building-your-application/authentication#layouts-and-auth-checks).
```
Ephem, line 67: No need to change anything here, but this formulation made me think we should sync on something:

> and you should if the layout itself reads privileged data

Say we have a `/dashboard/help` path or something. You can only visit the dashboard signed in, but this specific path only has static content. Maybe that content isn't even sensitive, it's kinda _fine_ if someone not signed in sees it.

I still believe that path should use `await auth.protect()` though. It's easier to reason about if all layouts/pages has that check, but it's also better UX.

So a tricky thing about `await auth.protect()` is that it's _also_ a UX affordance. Another example is, if we protect all API endpoints, server functions etc, there's technically no _security need_ for `await auth.protect()`, resources are protected, but we still want to include it so we:

* Redirect correctly instead of the app showing an error
* Avoid making those unnecessary API calls when we know they'll fail
* (Defence in depth)

I wanted to bring this up just so we are aligned that it has dual purposes.

Ephem (commented):
Re-review of `protect-content`. I think this page is pretty sweet now! 💯 Short and to the point, but covers a lot. It also matches the lint rule pretty closely which is nice.

Will review `clerk-middleware` next.

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,269 +36,119 @@ export const config = {
 }
 ```
 
-By default, `clerkMiddleware` will not protect any routes. All routes are public and you must opt-in to protection for routes.
-
 ## `createRouteMatcher()`
```
Ephem: How about we move this to the very bottom of this page to deemphasize it?

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,269 +36,119 @@ export const config = {
...
 
-`createRouteMatcher()` is a Clerk helper function that allows you to protect multiple routes. `createRouteMatcher()` accepts an array of routes and checks if the route the user is trying to visit matches one of the routes passed to it. The paths provided to this helper can be in the same format as the paths provided to the Next Middleware matcher.
+`createRouteMatcher()` is a Clerk helper function that accepts an array of routes and checks if the route the user is trying to visit matches one of the routes passed to it. The paths provided to this helper can be in the same format as the paths provided to the Next.js Middleware matcher.
 
 The `createRouteMatcher()` helper returns a function that, if called with the `req` object from the Middleware, will return `true` if the user is trying to access a route that matches one of the routes passed to `createRouteMatcher()`.
 
-In the following example, `createRouteMatcher()` sets all `/dashboard` and `/forum` routes as protected routes.
+> [!NOTE]
+> When matching a subtree of routes, prefer the `:path*` form: it maps directly to how the router matches path segments. For example, use `/dashboard/:path*` instead of `/dashboard(.*)`.
```
Ephem: Feels weird to recommend this when the pattern in the middleware matcher right above it is `(.*)`.

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,269 +36,119 @@ export const config = {
...
-- Use [`auth().isAuthenticated`](/docs/reference/nextjs/app-router/auth#protect-pages-and-routes) if you want more control over what your app does based on user authentication status.
-
-<CodeBlockTabs options={["auth.protect()", "auth().isAuthenticated()"]}>
-  ```tsx {{ filename: 'proxy.ts' }}
-  import { clerkMiddleware, createRouteMatcher } from '@clerk/nextjs/server'
+const isDashboardRoute = createRouteMatcher(['/dashboard/:path*'])
+```
 
-  const isProtectedRoute = createRouteMatcher(['/dashboard(.*)', '/forum(.*)'])
+For non-auth path logic in your proxy, use the framework's native matching (`config.matcher` and `req.nextUrl.pathname`). For an example, see [Geo blocking](/docs/guides/development/geo-blocking).
```
Ephem: It's a bit weird to recommend this before we have deprecated it. If we don't recommend it for auth, and don't recommend it for non-auth..?

I think we could just remove this section for now?

docs/reference/nextjs/clerk-middleware.mdx
Ephem: I've reconsidered and now think it's a good call to remove most of this section, the `if (isProtectedRoute(req)) await auth.protect()` example etc right away after all.

I kinda think this section is a bit weird now though. We don't recommend protecting routes in the middleware anymore, so why do we have a pretty hefty section on it?

I think this is perfect to keep here:

Image

I agree with moving the Migration guide out of the `protect-content` page, but I'm not sure this is the perfect place for it either? We should _definitely_ point to the migration guide from here though, but including the full thing feels like it distracts a lot from the other content which is the actual _reference_ content?

Other options are somewhere in Guides->Development->Upgrading Clerk, or maybe we make `clerkMiddleware()` in the side menu an accordion with the migration guide as a separate entry?

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,269 +36,119 @@ export const config = {
...
 
-// Create route matchers to identify which token type each route should require
-const isOAuthAccessible = createRouteMatcher(['/oauth(.*)'])
-const isApiKeyAccessible = createRouteMatcher(['/api(.*)'])
-const isMachineTokenAccessible = createRouteMatcher(['/m2m(.*)'])
-const isUserAccessible = createRouteMatcher(['/user(.*)'])
-const isAccessibleToAnyValidToken = createRouteMatcher(['/any(.*)'])
+  return userId
+}
+```
```
Ephem: Why this wrapper? It's just `const { userId } = await auth.protect()`?

This also wont work with the lint rule. 😄

docs/reference/nextjs/clerk-middleware.mdx
```diff
@@ -36,269 +36,119 @@ export const config = {
...
-      '/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)',
-      // Always run for API routes
-      '/(api|trpc)(.*)',
-      // Always run for Clerk-specific frontend API routes
-      '/__clerk/(.*)',
-    ],
-  }
-  ```
-</CodeBlockTabs>
+### Migrating from middleware-based auth checks
```
Ephem: This is a good start, but it feels a bit unfinished/sparse which is not weird at all since the story isn't that fleshed out.

Rather than leaving a bunch of feedback on this, I'll take a stab at rewriting it in a separate PR tomorrow, including the lint rule now that it's ready for release.

Ephem (commented):
Targeted review at `clerk-middleware.tsx`

---

I think we should update this section in [/nextjs/getting-started/quickstart](https://clerk.com/docs/pr/jacek-middleware-auth-resource-guidance/nextjs/getting-started/quickstart) to point to the new protect content page instead:

Image

alexisintech:
Pushed some review fixes directly in 3c51752a8.

- **Nuxt combined example**: scoped the `org:invoices:create` check to `/api/invoices` only — the previous shape applied it to `/api/admin` too, which would block admins who lack invoice permission.
- **Nuxt route matching**: added a `matchesRoute(path, route)` helper (segment-exact or `${route}/` prefix) so `/api/admin` doesn't collide with `/api/administer`. Same approach the session-tasks example already uses in this PR.
- **Astro `createRouteMatcher()` snippet**: added the missing `import { createRouteMatcher } from '@clerk/astro/server'`.
- **`reading.astro.mdx`**: switched the earlier `.astro` example from `/login` to `/sign-in` so it matches the new `requireAuth()` helper's redirect target.

alexisintech (approved):
alrightyyy I think this is ready to ship 🙏

PR #3429, opened 2026-06-09
docs(enterprise-connections): document multi-valued SAML custom attributes

NicolasLopes7:
Add a "Multi-valued attributes" section to the custom attribute mapping guide and a step in the create flow covering the Multi-value attribute toggle. Explains the single-string (default) vs array-of-strings shape written to publicMetadata, and how the setting interacts with SCIM.

### DIFF
--- a/docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
+++ b/docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
@@ -32,9 +32,24 @@ For example, if you define a custom attribute with the key `department`, it will
 1. In the **Custom attributes** section, select **Create custom attribute**.
 1. In the **Key** field, enter the `publicMetadata` key where this value will be stored (e.g., `department`).
 1. In the **Display name** field, enter a human-readable name (e.g., `Department`).
+1. If your IdP can send more than one value for this attribute (e.g., a user's `groups` or `roles`), enable **Multi-value attribute**. See [Multi-valued attributes](#multi-valued-attributes) for how this changes the stored value. Leave it disabled for single-value attributes like `department`.
 1. Select **Create attribute**.
 
 Once defined, the attribute appears in the **Attribute mapping** card on both the **SSO** tab and — if Directory Sync is enabled — the **Directory sync** tab, ready to be mapped to IdP-specific claim paths or SCIM schema attributes.
+
+### Multi-valued attributes
+
+By default, a custom attribute holds a single value. When your IdP sends the same attribute more than once in a SAML assertion — common for `groups`, `roles`, or `teams` — Clerk keeps only the first value and stores it as a string. To capture every value, enable **Multi-value attribute** when you define or edit the attribute.
+
+The setting controls the shape of the value Clerk writes to `publicMetadata`:
+
+- **Disabled (default):** Clerk stores the first value as a string. For example, `publicMetadata.department` is `"Engineering"`.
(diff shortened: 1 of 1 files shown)


gabrielmeloc22 (approved):

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
```diff
@@ -32,10 +32,25 @@ For example, if you define a custom attribute with the key `department`, it will
...
 Once defined, the attribute appears in the **Attribute mapping** card on both the **SSO** tab and — if Directory Sync is enabled — the **Directory sync** tab, ready to be mapped to IdP-specific claim paths or SCIM schema attributes.
 
+### Multi-valued attributes
+
+By default, a custom attribute holds a single value. When your IdP sends the same attribute more than once in a SAML assertion — common for `groups`, `roles`, or `teams` — Clerk keeps only the first value and stores it as a string. To capture every value, enable **Multi-value attribute** when you define or edit the attribute.
+
+The setting controls the shape of the value Clerk writes to `publicMetadata`:
+
+- **Disabled (default):** Clerk stores the first value as a string. For example, `publicMetadata.department` is `"Engineering"`.
+- **Enabled:** Clerk collects every value the IdP sends for that attribute and stores them as an array of strings. For example, `publicMetadata.groups` is `["engineering", "platform", "sre"]`. A single value is still stored as a one-element array (`["engineering"]`), and empty values are dropped.
```
im2nguyen: in the changelog, you mention that empty values are `[]`?

im2nguyen (approved):

manovotny:
Pushed some changes directly in fa93b838e.

- Reworked the empty-value line: blank values are dropped *from the array*, but an attribute the IdP sends with no usable values is stored as `[]` rather than omitted — matches `ExtractFromMultiValueMap` in clerk_go and the changelog. This addresses the inline question on line 47.
- Softened "the first value" → "one value" in the disabled/default explanation, since duplicate SAML `Attribute` entries resolve last-wins, not first.

Let me know how you feel about these changes. I'm most uncertain about the "one value" as it can be first or last value.

docs/guides/configure/auth-strategies/enterprise-connections/custom-attribute-mapping.mdx
manovotny: @im2nguyen I caught the same thing. I addressed it in https://github.com/clerk/clerk-docs/commit/fa93b838e69f30e3e5e52371b2dfebe96bef06b2.

manovotny:
Flagging the [Slack conversation regarding the naming/label](https://clerkinc.slack.com/archives/C01QFRQNHSN/p1781014305815089?thread_ts=1781014129.848379&cid=C01QFRQNHSN). We'll want to come to an agreement before merge.

manovotny:
Noting that label changes in dashboard need to merge first. https://github.com/clerk/dashboard/pull/9438

manovotny:
Pushed some changes directly in 8b3f8b2f5.

- Reworded the new terminology sentence to "In SAML and SCIM, these are commonly called multi-valued attributes" — SCIM's spec ([RFC 7643 §2.4](https://datatracker.ietf.org/doc/html/rfc7643#section-2.4)) formally uses the term, but the SAML core spec and X.500/LDAP attribute profile don't, so "the specifications call these" overstated the SAML half.
- Reordered the frontmatter description so "shared across SSO and Directory Sync connections" reads as modifying the metadata rather than the newly inserted "including multi-valued attributes" clause.

manovotny (approved):
I'm good with these changes. Just need the dashboard PR to land.

Proactively approving, but the "do not merge" label will still gate an actual merge while we wait.

PR #3430, opened 2026-06-09
Update Client Trust enable steps to Attack protection page

mwickett:
## Description

The Client Trust control is moving from the **Password** settings tab to the **Attack protection** page in the dashboard (clerk/dashboard#9387). This updates the "Enable Client Trust" steps in the Client Trust guide to point to **Attack protection**, matching the navigation convention already used by the bot-protection, user-lockout, and user-enumeration-protection guides.

This also clarifies two related Client Trust details:

- Client Trust requires password-based sign-in to be enabled.
- The **Client Trust Status** update on the **Updates** page does not enable Client Trust. It only changes older custom-flow responses from the legacy `needs_second_factor` + `client_trust_state` shape to the newer `needs_client_trust` status.

## Merge timing

Held behind the `do not merge` label until clerk/dashboard#9387 ships — merging earlier would document a control location that is not live yet.

## Testing

- `pnpm run build:tsx`
- `pnpm run lint`

### DIFF
--- a/docs/guides/secure/client-trust.mdx
+++ b/docs/guides/secure/client-trust.mdx
@@ -24,8 +24,11 @@ When these conditions are met, Clerk sends a [one-time verification code](!otp) 
 
 Client Trust is automatically enabled for Clerk applications created after November 14, 2025. For applications created before this date, you can enable Client Trust in the Clerk Dashboard:
 
-1. In the Clerk Dashboard, navigate to the [**Updates**](https://dashboard.clerk.com/~/updates) page.
-1. Enable the **Client Trust** update.
+1. In the Clerk Dashboard, navigate to the [**Attack protection**](https://dashboard.clerk.com/~/protect/attack-protection) page.
+1. Enable **Client Trust**.
+
+> [!NOTE]
+> Client Trust requires password-based sign-in to be enabled.
 
 ## How Client Trust differs from traditional MFA
 
--- a/docs/guides/secure/client-trust.mdx
+++ b/docs/guides/secure/client-trust.mdx
@@ -15,7 +15,7 @@ Client Trust automatically requires a [second factor](!second-factor) when **all
 1. The user **hasn't enabled [multi-factor authentication (MFA)](/docs/guides/configure/auth-strategies/sign-up-sign-in-options#multi-factor-authentication)**.
 1. The user is signing in from a **new device**.
 
-When these conditions are met, Clerk sends a [one-time verification code](!otp) to the user's email address or phone number. The user must enter this code to complete the sign-in process.
+When these conditions are met, Clerk uses an email code, SMS code, or email link based on your settings. The user must complete this verification to finish the sign-in process.
 
 > [!NOTE]
 > If the user has already enabled MFA (such as an authenticator app or SMS verification), Client Trust will not trigger - their existing MFA method will be used instead.
@@ -35,7 +35,7 @@ Client Trust is automatically enabled for Clerk applications created after Novem
 | Aspect | Client Trust | Traditional MFA |
 | - | - | - |
 | **Trigger** | Automatic on new devices | User must enable it |
-| **Verification method** | Email/phone code | TOTP, SMS code, backup codes |
+| **Verification method** | Email code, SMS code, or email link | TOTP, SMS code, backup codes |
 | **User setup required** | None | User must configure their preferred method |
 | **When it applies** | Only for password sign-ins | All sign-ins |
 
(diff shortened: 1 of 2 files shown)


alexisintech:
my question is, is there ever an instance where a user needs to go to the updates page to enable client trust?
would it be for apps created before nov 14, 2025 or apps created before this feature release date? or neither?

manovotny:
> my question is, is there ever an instance where a user needs to go to the updates page to enable client trust? would it be for apps created before nov 14, 2025 or apps created before this feature release date? or neither?

Would love for @mwickett to sanity check this, but here's what I see.

Neither — the Updates-page "Client Trust" entry and the dashboard control are two switches wired to the same boolean (`flag_client_trust`). The Attack protection toggle flips it directly for any instance (clerk_go: `api/dapi/v1/instances/service.go`), so pre-Nov-14 apps can enable it there without ever touching the Updates page, and newer apps have it on at creation. Pointing the enable steps only at Attack protection is correct.

One caveat from verifying this: a separate Updates-page item, `needs_client_trust_status`, still matters for older apps with **custom sign-in flows**. It doesn't turn the protection on or off — it controls whether sign-ins return the `needs_client_trust` status our custom-flow docs tell readers to handle, vs plain `needs_second_factor` (`api/shared/sign_in/service.go`). New apps get it at creation; pre-Nov-14 apps that only flip the Attack protection toggle don't. Might be worth a short legacy-apps note in the guide — could also be follow-up scope.

alexisintech:
> > my question is, is there ever an instance where a user needs to go to the updates page to enable client trust? would it be for apps created before nov 14, 2025 or apps created before this feature release date? or neither?
> 
> Would love for @mwickett to sanity check this, but here's what I see.
> 
> Neither — the Updates-page "Client Trust" entry and the dashboard control are two switches wired to the same boolean (`flag_client_trust`). The Attack protection toggle flips it directly for any instance (clerk_go: `api/dapi/v1/instances/service.go`), so pre-Nov-14 apps can enable it there without ever touching the Updates page, and newer apps have it on at creation. Pointing the enable steps only at Attack protection is correct.
> 
> One caveat from verifying this: a separate Updates-page item, `needs_client_trust_status`, still matters for older apps with **custom sign-in flows**. It doesn't turn the protection on or off — it controls whether sign-ins return the `needs_client_trust` status our custom-flow docs tell readers to handle, vs plain `needs_second_factor` (`api/shared/sign_in/service.go`). New apps get it at creation; pre-Nov-14 apps that only flip the Attack protection toggle don't. Might be worth a short legacy-apps note in the guide — could also be follow-up scope.

yeah this is exactly what i was thinking - gonna wait for mike to sanity check and if this is all correct, i have some updates i'll need to apply to this PR

mwickett:
Good call - your take checked out @manovotny - I pushed an update to call out the update being tied to the custom UI response.

manovotny:
Pushed some changes directly in 7116b29a8.

- Reworked the SMS-tab setup step in the custom flow Client Trust guide that said "**Client Trust** is enabled by default" — it now notes that applications created before November 14, 2025 need to [enable Client Trust](https://clerk.com/docs/guides/secure/client-trust#enable-client-trust), instead of implying the control lives on the Password tab (which stops being true once clerk/dashboard#9387 ships).
- Added email link to the intro's list of strategies a custom flow may need to handle, with a note that the guide's examples demonstrate the email code and SMS code strategies.

@mwickett sorry this expanded the scope a bit, but these caught my eye while investigating your other changes. Let me know if I went too far.

manovotny (approved):
With the follow up commits, this looks good to me now. However, @alexisintech should definitely weight in again before merge.

docs/guides/secure/client-trust.mdx
```diff
@@ -24,29 +24,34 @@ When these conditions are met, Clerk sends a [one-time verification code](!otp)
 
 Client Trust is automatically enabled for Clerk applications created after November 14, 2025. For applications created before this date, you can enable Client Trust in the Clerk Dashboard:
 
-1. In the Clerk Dashboard, navigate to the [**Updates**](https://dashboard.clerk.com/~/updates) page.
-1. Enable the **Client Trust** update.
+1. In the Clerk Dashboard, navigate to the [**Attack protection**](https://dashboard.clerk.com/~/protect/attack-protection) page.
+1. Enable **Client Trust**.
+
+> [!NOTE]
+> Client Trust requires password-based sign-in to be enabled.
```
alexisintech:
```suggestion
> Client Trust requires [password-based sign-in](/docs/guides/configure/auth-strategies/sign-up-sign-in-options#user-and-authentication) to be enabled.
```

docs/guides/secure/client-trust.mdx
```diff
@@ -24,29 +24,34 @@ When these conditions are met, Clerk sends a [one-time verification code](!otp)
...
 | **User setup required** | None | User must configure their preferred method |
 | **When it applies** | Only for password sign-ins | All sign-ins |
 
 ## Impact on custom sign-in flows
 
 If you've built a [custom sign-in flow](!custom-flow) that allows password-based sign-ins using the Clerk API or SDKs, you'll need to handle the `needs_client_trust` status that Client Trust can trigger.
 
-When Client Trust requires verification, the sign-in attempt will return a status of `needs_client_trust` with `email_code` or `phone_code` in the `supportedSecondFactors` array. Your flow should:
+> [!NOTE]
+> If an older application has Client Trust enabled but your custom sign-in flow receives the legacy `needs_second_factor` status with `client_trust_state` instead of `needs_client_trust`, opt in to the **Client Trust Status** update on the [**Updates**](https://dashboard.clerk.com/~/updates) page. This update doesn't enable Client Trust. It only changes the custom-flow response from `needs_second_factor` to `needs_client_trust`.
```
alexisintech: what is `client_trust_state`? I've never seen that before.

I know that in a custom flow, the [`signIn.status` ](https://clerk.com/docs/nextjs/reference/objects/sign-in-future#properties:~:text=The%20current%20status%20of%20the%20sign%2Din.) can be `needs_second_factor` for legacy apps, and `needs_client_trust` for newer apps. but I'm not familiar with `client_trust_state` - could you explain/send me a reference

alexisintech (commented):
great updates! just a question here @mwickett

docs/guides/secure/client-trust.mdx
mwickett: For apps that have Client Trust enabled but haven't opted in to the Client Trust Status update, `client_trust_state` is the legacy field that can be returned on the sign-in response. In the Client Trust extra-verification case, that means the app receives `needs_second_factor` plus `client_trust_state`, whose [values ](https://github.com/clerk/javascript/blob/8744728e6610b2229f56dd3b31975c3f57395f02/packages/shared/src/types/json.ts#L107) are `pending | new | known`.

Opting in to Client Trust Status changes that response shape so Client Trust-triggered verification returns `needs_client_trust` instead of `needs_second_factor`.

alexisintech: gotcha - the custom flows don't use this state so imo we can remove this and focus on the change that is relevant

```suggestion
> If an older application has Client Trust enabled but your custom sign-in flow receives the legacy `needs_second_factor` status instead of `needs_client_trust`, opt in to the **Client Trust Status** update on the [**Updates**](https://dashboard.clerk.com/~/updates) page. This update doesn't enable Client Trust. It only changes the custom-flow response from `needs_second_factor` to `needs_client_trust`.
```

alexisintech (approved):
thanks for the collaboration on this team 😸💖

mwickett:
Once the Dashboard PR releases (it's now merged), I'll merge this PR.

PR #3431, opened 2026-06-09
Fix dismissible prop names in view docs

seanperez29:
### 🔎 Previews:

- AuthView
- UserProfileView
- OrganizationListView
- OrganizationProfileView

- Updates the native iOS and Android docs to use `isDismissible` instead of `isDismissable`.
- Updated parameter tables, code examples, and related copy.
- Left Expo docs unchanged since that work is covered separately.


SarahSoutoul (approved):
LGTM!

PR #3434, opened 2026-06-10
chore: NPM --> npm

alexisintech:


manovotny (approved):
I was in disbelief these were the only two instances, but sure enough! Nice work!

PR #3438, opened 2026-06-11
Replace Next.js quickstart prompt with CLI-based setup prompt

djgould:
### 🔎 Previews:

- [Next.js Quickstart](https://clerk.com/docs/nextjs/getting-started/quickstart) (serves `prompts/nextjs-quickstart.md` via its `llm:` frontmatter)

- Replaces the pre-built AI prompt on the Next.js quickstart page with the CLI-driven setup flow (`clerk auth login` → `clerk init` → `clerk doctor`), aligned with the [clerk.com hero prompt](https://github.com/clerk/clerk/blob/3105a4b75dffa6b47a3d8d86647f06a7692557ca/src/lib/hero-setup-prompt.ts#L69) and the [AI-first quickstart experiment prompt](https://github.com/clerk/clerk/blob/main/src/app/docs/experiment_ai-first-quickstart/setup-prompt.ts), but tailored to Next.js:
  - Drops the 12-framework fallback table and other-framework guidance (dead weight on a Next.js-specific page); the docs fallback is a single link to the Next.js quickstart.
  - Restores package-manager detection from the hero prompt instead of the experiment variant's hard-coded `--pm npm`.
  - Replaces the bare JSX snippets with a complete `app/layout.tsx` example showing where the auth controls go.
- Why not verbatim: the full hero prompt generates a 9,747-char Cursor deeplink URL, over the 8,000 limit enforced by `check-prompt-deeplink-length` — the "Open in Cursor" button would stop showing. The tailored prompt is 7,215 chars with headroom to spare.

### Deadline

- No rush

### Other resources

- Source prompts: [hero-setup-prompt.ts](https://github.com/clerk/clerk/blob/3105a4b75dffa6b47a3d8d86647f06a7692557ca/src/lib/hero-setup-prompt.ts#L69), [experiment_ai-first-quickstart/setup-prompt.ts](https://github.com/clerk/clerk/blob/main/src/app/docs/experiment_ai-first-quickstart/setup-prompt.ts)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,6 @@ data/api_errors.json
 clerk-typedoc/
 local-clerk-typedoc/
 pnpm-lock.yaml
+# Prompts are verbatim copies of prompts maintained in clerk/clerk; formatting
+# would diverge them from their source and inflate the Cursor deeplink length.
+prompts/
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,6 +2,3 @@ data/api_errors.json
 clerk-typedoc/
 local-clerk-typedoc/
 pnpm-lock.yaml
-# Prompts are verbatim copies of prompts maintained in clerk/clerk; formatting
-# would diverge them from their source and inflate the Cursor deeplink length.
-prompts/
--- a/prompts/nextjs-quickstart.md
+++ b/prompts/nextjs-quickstart.md
@@ -1,6 +1,6 @@
-# Add Clerk Authentication
+# Add Clerk Authentication to Next.js
 
-Set up Clerk authentication with the Clerk CLI.
+Set up Clerk authentication in this Next.js project with the Clerk CLI.
 
 ## Quick Setup
 
@@ -10,7 +10,7 @@ Before running any commands, present the user with a preliminary setup checklist
 Here's what I'll do to get you set up with Clerk.
 
 1. Install or update the Clerk CLI
-2. Set up Clerk in this project, or scaffold a new app with Clerk if this directory is empty
+2. Set up Clerk in this project, or scaffold a new Next.js app with Clerk if this directory is empty
 3. Start your app with Clerk installed.
 
 Shall I proceed?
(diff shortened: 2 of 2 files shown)


manovotny:
Pushed some review fixes directly in e024c665b.

- Added the Next.js version → middleware filename rule as a Critical rule (`proxy.ts` on 16+, `middleware.ts` on 15 and below — identical contents), referenced from Step 4's manual fallback. This piggybacks [DOCS-11835](https://linear.app/clerk/issue/DOCS-11835/agent-facing-nextjs-quickstart-prompt-mandates-proxyts-unconditionally): `clerk init` names the file correctly on its own (verified in the CLI source), but the manual path had no rule, and `clerk doctor` has no filename check to catch a miss.
- Gave Step 4 inline manual setup steps. The docs fallback URL serves this same prompt to any agent fetching it as markdown (the `llm:` frontmatter swap), so an agent whose `clerk init` failed would loop with no actual setup instructions. The link stays, labeled so agents don't chase it.
- Step 7 now says to install `@clerk/ui` with the project's package manager instead of unconditionally `npm install`.
- The CLI presence check gets `|| echo "Clerk CLI not installed"` so the expected not-installed branch exits 0 — some agent harnesses treat non-zero exits as errors to retry or stop on.
- Dropped the "Next.js 15+:" qualifier from the `await auth()` rule — `auth()` is async in every installable SDK version, and the conditional framing invites an agent on an older Next to skip the await.
- Added "Sign in to Clerk (opens your browser)" to the Quick Setup checklist so the user approves the OAuth pause up front.
- Scoped the lockfile signals to directories without a project, so they no longer read as contradicting the no-`--pm`-for-existing-projects rule.
- Restored `import './globals.css'` in the layout example so it models preserving existing imports.

Deeplink budget after all of this: 7,955 of 8,000 encoded chars (was 7,215) — future additions to this prompt will need to trim first. `build:tsx`, Prettier, and the deeplink length check all pass.

The rest of your changes look good. If you're good with my changes, then I say we roll. 🤝

manovotny (approved):

PR #3446, opened 2026-06-16
Remove SendGrid mentions from docs

SarahSoutoul:
### What does this solve?

These docs mentioned **SendGrid** by name in a couple of places, which is more implementation detail than we want to expose. This update removes those provider-specific references while keeping the delivery and deliverability guidance intact.

### What changed

- Removes the SendGrid reference from the email template settings guide.
- Rewords the email deliverability guide to keep the dedicated-IP explanation without naming the provider
- Updates the deliverability page description to match the new provider-neutral wording


alexisintech (approved):
LGTM!

PR #3473, opened 2026-06-25
Copy edits: createRouteMatcher migration guide + eslint-plugin docs

jacekradko:
Copy pass over the createRouteMatcher migration guide and the eslint-plugin reference, aimed at your branch.

Most of it is mechanical (typos, casing, first-person to "Clerk", a couple of comma splices). Three choices are worth a look: I kept "Server Functions" rather than the corpus's "Server Actions" since that's the term the plugin's `serverFunctions` API uses and it's genuinely broader, with an "(including Server Actions)" bridge at first use; I standardized the prose on "Middleware" / `clerkMiddleware()` instead of "Proxy" so it matches the rest of the Next.js docs and the `clerkMiddleware()` reference (the `proxy.ts` filename stays); and I restored two bits the clerk-middleware rewrite had dropped, the geo-blocking / native-matching pointer and the "line to hold" gate rule.

The eslint-plugin wording (glob dialect, "next executable statement") was checked against the rule source. Build and prettier pass.

### DIFF
--- a/docs/guides/development/custom-sign-in-or-up-page.mdx
+++ b/docs/guides/development/custom-sign-in-or-up-page.mdx
@@ -27,7 +27,7 @@ To set up separate sign-in and sign-up pages, follow this guide, and then follow
 
   Ensure the `/sign-in` route is accessible to all users, including unauthenticated users.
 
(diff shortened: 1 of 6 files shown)


Ephem (approved):

alexisintech:
This is great!! Caught a lot of things I caught as well.
I just posted a docs review that wrote over some of these changes. But there are some things in here I'd like to bring over for sure

alexisintech (approved):
LGTM, thank you 😸💖

PR #3475, opened 2026-06-25
feat(auth-strategies): Document passkeys satisfy second factor

dmoerner:
### 🔎 Previews:

- https://clerk-docs-hecgt82fx.clerkstage.dev/

Do not merge. Needs to be updated with the exact rollout date, once we've merged and applied that in terraform.

Describe that passkeys satisfy second factor, which will be the new default behavior for new instances. Document how to opt in to this functionality for older instances using the Clerk Dashboard. Newer instances are forced to use the new feature. (This is just like how we implemented the PII changes.)

### Deadline

No rush. Still waiting on Dashboard PR review, and everything is behind a flag we will roll out in Terraform.

### Other resources

https://linear.app/clerk/issue/USER-3966/passkey-as-2fa-option

### DIFF
--- a/docs/guides/configure/auth-strategies/sign-up-sign-in-options.mdx
+++ b/docs/guides/configure/auth-strategies/sign-up-sign-in-options.mdx
@@ -162,12 +162,13 @@ If you're building a custom user interface instead of using the [Account Portal]
 
 A [passkey](#passkeys) is itself a multi-factor credential — a single sign-in combines two authentication factors (a physical device and a PIN or biometric) — so a passkey sign-in can satisfy your application's multi-factor requirement on its own. When this behavior is in effect, a user who signs in with a passkey isn't prompted to complete a second factor.
 
-For instances created after **\[ROLLOUT DATE — TODO]**, this is the default behavior and can't be disabled.
+For instances created on or after **\[ROLLOUT DATE — TODO]**, this is the default behavior and can't be disabled.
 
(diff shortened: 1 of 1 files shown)


manovotny:
Pushed some changes directly in 7ab01b869.

- Reworked the opt-in steps to reference **Sign-in with passkey** (the actual Dashboard card label) and added a step to enable it first — the **Passkeys satisfy multi-factor authentication** toggle only renders once passkey sign-in is on, so a legacy instance without passkeys enabled wouldn't have found it.
- Changed "created after" to "created on or after" for the default-on side. The backend cutoff is a strict less-than (`IsBeforeCutoff`), so an instance created exactly at the rollout timestamp already gets the locked-on behavior; the "before" opt-in line is correct as-is.

Otherwise, the changes LGTM. Just need the downstream things to merge and the official date. 👍

dmoerner:
Thanks. If Terraform gets unblocked I'm hoping to toggle this on at midnight UTC tonight.

manovotny:
> Thanks. If Terraform gets unblocked I'm hoping to toggle this on at midnight UTC tonight.

Cool. I'll proactively approve so you can remove the `do not merge` blocker label and merge (after updating the date too). Feel free to DM me if you need anything.

manovotny (approved):

dmoerner:
Moved to main clerk repo https://github.com/clerk/clerk/pull/2889

PR #3476, opened 2026-06-25
feat(reverification): Document new reverification_id shortcode

dmoerner:
### 🔎 Previews:

- https://clerk-docs-2mmgbbscn.clerkstage.dev/

Document the use of the new `session.reverification_id` shortcode and illustrate a typical workflow that a client could use to implement SCA.

(I was surprised to see we do not have a canonical list of shortcodes in the Docs, but it is available in the Dashboard so I guess it's fine and it would just be something else we would need to keep in sync.)

### Deadline

No major deadline. Requested by OneSafe and Finary. ~Depends on https://github.com/clerk/dashboard/pull/9558~ Now released!

### Other resources

https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32018R0389&from=EN#d1e493-23-1

### DIFF
--- a/docs/guides/secure/reverification.mdx
+++ b/docs/guides/secure/reverification.mdx
@@ -200,9 +200,9 @@ Clerk tracks every reverification with a unique ID. You can include this ID in y
 - `reverification_id` provides **uniqueness**: it identifies _which_ reverification the token reflects.
 - `fva` provides **freshness**: it conveys how recently the user's factors were verified.
 
-Because every successful reverification generates a new `reverification_id`, and both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
+Each time a reverification starts, Clerk mints a new `reverification_id`. Because both claims are part of the signed session token, your backend can confirm `fva` reflects a recent verification and use `reverification_id` to correlate that reverification with a specific action. A typical flow:
 
 1. Add `{{session.reverification_id}}` as a custom claim to your session token. See [Customize your session token](/docs/guides/sessions/customize-session-tokens).
 1. When the user initiates a sensitive action, [trigger a reverification](#how-to-require-reverification) so a new reverification ID is minted.
-1. On your backend, verify the signed session token, confirm that `fva` is within your required window, and record the `reverification_id` alongside the action.
-1. Reject any action whose `reverification_id` has already been used, ensuring each reverification is consumed only once.
+1. On your backend, verify the signed session token, confirm that `fva` is within your required window, and record the `reverification_id` alongside the action's specific details, such as the payment amount and payee.
+1. Reject any action whose `reverification_id` has already been used, or whose recorded details don't match the action being performed, ensuring each reverification authorizes exactly one action.
--- a/docs/guides/secure/reverification.mdx
+++ b/docs/guides/secure/reverification.mdx
@@ -200,7 +200,7 @@ Clerk tracks every reverification with a unique ID. You can include this ID in y
 - `reverification_id` provides **uniqueness**: it identifies _which_ reverification the token reflects.
 - `fva` provides **freshness**: it conveys how recently the user's factors were verified.
 
-Because both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
+Because every successful reverification generates a new `reverification_id`, and both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
 
 1. Add `{{session.reverification_id}}` as a custom claim to your session token. See [Customize your session token](/docs/guides/sessions/customize-session-tokens).
 1. When the user initiates a sensitive action, [trigger a reverification](#how-to-require-reverification) so a new reverification ID is minted.
--- a/docs/guides/secure/reverification.mdx
+++ b/docs/guides/secure/reverification.mdx
@@ -190,3 +190,19 @@ After setting up reverification on the server-side, you must handle reverificati
   }
   ```
 </If>
+
+## Correlate a reverification with a specific action
+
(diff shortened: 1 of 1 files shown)


dmoerner:
@beerose I would be great to also get your eyes here on this doc that discusses SCA.

docs/guides/secure/reverification.mdx
```diff
@@ -190,3 +190,19 @@ After setting up reverification on the server-side, you must handle reverificati
...
+## Correlate a reverification with a specific action
+
+Some use cases require proving that a single reverification corresponds to exactly one sensitive action — for example, [Strong Customer Authentication (SCA)](https://en.wikipedia.org/wiki/Strong_customer_authentication) for payments, where each payment must be backed by its own verification and a verification can't be replayed across payments.
+
+Clerk tracks every reverification with a unique ID. You can include this ID in your session token by adding the `{{session.reverification_id}}` shortcode as a custom claim. Combined with the default [`fva` (factor verification age)](/docs/guides/sessions/session-tokens) claim, the token gives you two guarantees:
+
+- `reverification_id` provides **uniqueness**: it identifies _which_ reverification the token reflects.
+- `fva` provides **freshness**: it conveys how recently the user's factors were verified.
+
+Because both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
```
beerose: Nitpick: just making it more explicit 
```suggestion
Because every successful reverification generates a new `reverification_id`, and both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
```

docs/guides/secure/reverification.mdx
```diff
@@ -190,3 +190,19 @@ After setting up reverification on the server-side, you must handle reverificati
   }
   ```
 </If>
+
+## Correlate a reverification with a specific action
+
+Some use cases require proving that a single reverification corresponds to exactly one sensitive action — for example, [Strong Customer Authentication (SCA)](https://en.wikipedia.org/wiki/Strong_customer_authentication) for payments, where each payment must be backed by its own verification and a verification can't be replayed across payments.
```
beerose: Nitpick. No strong opinions here, but maybe "dynamic linking" might be worth mentioning. 

```suggestion
Some use cases require proving that a single reverification corresponds to exactly one sensitive action — for example, [Strong Customer Authentication (SCA)](https://en.wikipedia.org/wiki/Strong_customer_authentication) for payments, where each payment must be backed by its own verification and a verification can't be replayed across payments. This pattern is useful in scenarios that require dynamic linking.
```

beerose (approved):
LGTM!

docs/guides/secure/reverification.mdx
```diff
@@ -190,3 +190,19 @@ After setting up reverification on the server-side, you must handle reverificati
...
+
+Some use cases require proving that a single reverification corresponds to exactly one sensitive action — for example, [Strong Customer Authentication (SCA)](https://en.wikipedia.org/wiki/Strong_customer_authentication) for payments, where each payment must be backed by its own verification and a verification can't be replayed across payments. This pattern is useful in scenarios that require dynamic linking.
+
+Clerk tracks every reverification with a unique ID. You can include this ID in your session token by adding the `{{session.reverification_id}}` shortcode as a custom claim. Combined with the default [`fva` (factor verification age)](/docs/guides/sessions/session-tokens) claim, the token gives you two guarantees:
+
+- `reverification_id` provides **uniqueness**: it identifies _which_ reverification the token reflects.
+- `fva` provides **freshness**: it conveys how recently the user's factors were verified.
+
+Because every successful reverification generates a new `reverification_id`, and both claims are part of the signed session token, your backend can verify them and correlate a reverification with a specific action. A typical flow:
```
manovotny: `every successful reverification generates a new reverification_id` is a bit stronger than the shortcode actually guarantees. It resolves to `session.ActiveStepUpID`, which is set when a reverification *starts* and rotates again on the next one — so its presence means one was started, not that it completed. 

`fva` is what conveys completion/freshness. 

Worth separating the two roles so a fresh ID alone isn't read as proof of a completed step-up?

```suggestion
Each time a reverification starts, Clerk mints a new `reverification_id`. Because both claims are part of the signed session token, your backend can confirm `fva` reflects a recent verification and use `reverification_id` to correlate that reverification with a specific action. A typical flow:
```

docs/guides/secure/reverification.mdx
```diff
@@ -190,3 +190,19 @@ After setting up reverification on the server-side, you must handle reverificati
...
+
+1. Add `{{session.reverification_id}}` as a custom claim to your session token. See [Customize your session token](/docs/guides/sessions/customize-session-tokens).
+1. When the user initiates a sensitive action, [trigger a reverification](#how-to-require-reverification) so a new reverification ID is minted.
+1. On your backend, verify the signed session token, confirm that `fva` is within your required window, and record the `reverification_id` alongside the action.
+1. Reject any action whose `reverification_id` has already been used, ensuring each reverification is consumed only once.
```
manovotny: Recording the `reverification_id` and rejecting reuse gives you replay prevention, but the intro frames this as *dynamic linking*, which specifically means binding the verification to the transaction's details (amount, payee), not just consuming it once. 

Could the backend steps record and check those details alongside the ID, so the flow delivers what dynamic linking implies?

```suggestion
1. On your backend, verify the signed session token, confirm that `fva` is within your required window, and record the `reverification_id` alongside the action's specific details, such as the payment amount and payee.
1. Reject any action whose `reverification_id` has already been used, or whose recorded details don't match the action being performed, ensuring each reverification authorizes exactly one action.
```

docs/guides/secure/reverification.mdx
dmoerner: Correct, thank you for the clarification suggestion.

docs/guides/secure/reverification.mdx
dmoerner: Yes, that's more precise, thank you. As long as we make clear it's their backend, which we do here.

manovotny (approved):

dmoerner:
Thank you @manovotny for the review! Going to be demoing this tomorrow as well :)

PR #3477, opened 2026-06-25
fix(auth-strategies): Move immutable attribute setting

dmoerner:
### 🔎 Previews:

- https://clerk-docs-j9ty5zhad.clerkstage.dev/

The immutable attribute setting is moving from each attribute's page (where it was a child of "Used for Sign Up") and instead to the User model page. The reason for this is that some instances have an attribute disabled for sign up but used for sign in, but they still want it to be set immutable.

~Depends on https://github.com/clerk/dashboard/pull/9493~ This is now released.

### Deadline

No rush

### DIFF
--- a/docs/guides/configure/auth-strategies/sign-up-sign-in-options.mdx
+++ b/docs/guides/configure/auth-strategies/sign-up-sign-in-options.mdx
@@ -102,6 +102,7 @@ In this section, you can:
 
 - Allow users to set their first and last name.
 - Allow users to delete their accounts.
+- Control whether users can change their email address, phone number, or username after sign-up. See [Restrict changes](#restrict-changes).
 
 ## Restrict changes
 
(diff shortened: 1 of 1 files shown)


manovotny:
Pushed a couple of changes directly in 4ad330232.

- Reworded the toggle-visibility note: "only appears for identifiers you've enabled" → "only appears for identifiers that are used for sign-up or sign-in." The dashboard renders that toggle whenever the identifier is in use (`isAttributeInUse` = `enabled || required || used_for_first_factor`), so the old wording would have told the sign-in-only users this change is meant for that the toggle isn't there for them.
- Deep-linked the **User model** reference to the `?user_auth_tab=user-profile` tab so readers land on the right tab instead of the default Email tab.

Specifically let me know if I got the identifiers part wrong.

manovotny (approved):

dmoerner:
Thanks @manovotny, the fixes look clearer!

PR #3515, opened 2026-07-22
docs(sessions): document the oiat session token header

nikosdouvlis:
### 🔎 Previews:

- /docs/guides/sessions/session-tokens#header-parameters (will add the Vercel link once the preview builds)

Session tokens carry an `oiat` parameter in their JWT header and we've never documented it. The guide only covers the payload claims, so when a customer's parser tripped over a header parameter it didn't know about, there was nothing for them to look up.

Adds a "Header parameters" section to the session tokens guide with `oiat`, what it means and why it can be older than `iat`. Plus a note that we may add header parameters over time, so parsers are better off ignoring unknown ones than validating against a fixed list.

Leaving the `m` parameter out on purpose, that info is moving to a response header so it isnt something people should be reading off the token.

### Deadline

No rush, but sooner is better since support is fielding questions on this one

### Other resources

- Slack thread that prompted this: https://clerkinc.slack.com/archives/C0AHE7YT3BN/p1784657146864089

### DIFF
--- a/docs/guides/sessions/session-tokens.mdx
+++ b/docs/guides/sessions/session-tokens.mdx
@@ -125,6 +125,17 @@ If you would like to add custom claims to your session token, you can [customize
 
 You can also create custom tokens using a [JWT template](/docs/guides/sessions/jwt-templates).
 
+## Header parameters
+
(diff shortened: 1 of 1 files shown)


nikosdouvlis:
Docs moved to clerk/clerk, reopened this over there: https://github.com/clerk/clerk/pull/3003

PR #3517, opened 2026-07-23
docs: require the original client IP in X-Forwarded-For for CDN-fronted FAPI proxies

mwickett:
## What

Two small doc clarifications for Frontend API proxying:

- **`proxy-fapi.mdx`** — adds a warning (and an inline nginx comment) that `X-Forwarded-For` must carry the **original end user's** IP as the leftmost value, and explains the CDN/load-balancer case: if the proxy origin is itself behind a CDN, the immediate peer (nginx's `$remote_addr`, Cloudflare's `CF-Connecting-IP`) is the *CDN's* IP, not the user's — so you must take the real client IP from the CDN's viewer header (e.g. `CloudFront-Viewer-Address`) instead.
- **`behind-a-proxy.mdx`** — adds a scope note distinguishing "deploy your app behind a reverse proxy" from "proxy the Frontend API through your domain," cross-linking the correct guide. The two pages list different required headers because they're different setups; nothing said so.

## Why

The proxy-fapi guide required `X-Forwarded-For` but its own nginx/Workers examples set it to the immediate peer. For an origin fronted by a CDN, that forwards a datacenter IP as the client on **every** request. Clerk reads the leftmost `X-Forwarded-For` value as the client IP for verified proxies, so a fronted origin that follows the examples literally ships an IP-blind proxy that passes every functional test while every real user looks like datacenter traffic — degrading IP-based security and rate-limiting. Observed in production on a real proxied instance; the real user IP was intact in `CloudFront-Viewer-Address` but never in `X-Forwarded-For`.

## Notes

- Wording is intentionally generic (no product internals): "security and rate-limiting features."
- `pnpm build:tsx` and `pnpm lint` both pass; the new internal cross-link resolves.
- Client-IP behavior verified against `clerk_go` (`api/fapi/v1/domain/http.go`, `pkg/handlers/ip.go` — leftmost `X-Forwarded-For`).

### DIFF
--- a/docs/guides/dashboard/dns-domains/proxy-fapi.mdx
+++ b/docs/guides/dashboard/dns-domains/proxy-fapi.mdx
@@ -40,6 +40,11 @@ When using a proxy, all requests to the Frontend API will be made through your d
   - `Clerk-Secret-Key`: Your Clerk [Secret Key](!secret-key).
   - `X-Forwarded-For`: The IP address of the original client making the request.
 
+  > [!WARNING]
+  > `X-Forwarded-For` must carry the **original end user's** IP address as the first (leftmost) value. Clerk uses this IP for security and rate-limiting features, so an incorrect value can cause legitimate users to be treated as automated traffic.
+  >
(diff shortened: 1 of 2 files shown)


mwickett:
Closing — clerk-docs has merged into the clerk/clerk monorepo, so this change belongs there now. Re-filed as clerk/clerk#3012 (identical edits under clerk-docs/).

PR #3522, opened 2026-07-30
docs: add planPeriod to CheckoutButton custom button example

somcpp:
### 🔎 Previews

- N/A (documentation change)

---

This PR updates the custom `CheckoutButton` example to include the `planPeriod` prop and adds a brief note clarifying that the same required props should be provided when using a custom button.

These changes make the custom button example consistent with the other `CheckoutButton` examples on the page and help ensure developers have a consistent reference regardless of which example they follow.

---

### Deadline

- No rush.

---

### Other resources

- None.

### DIFF
--- a/docs/reference/components/billing/checkout-button.mdx
+++ b/docs/reference/components/billing/checkout-button.mdx
@@ -36,10 +36,13 @@ The `<CheckoutButton />` component renders a button that opens the checkout draw
 </>
 ```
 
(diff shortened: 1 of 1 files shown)


manovotny:
Thanks for this — the custom-button example was the only one on the page missing `planPeriod`.

This is the first external contribution since we moved our docs into Clerk's internal monorepo. Your change is merged with your commits and authorship preserved, and it'll be live on https://clerk.com/docs as soon as it's reviewed.

Closing this out since the change ships through the monorepo. Thanks again for making the docs better.

PR #3525, opened 2026-08-09
docs: add Next.js 16 silent redirect abort troubleshooting

kishan59:
## Description
This PR adds a troubleshooting section regarding a common, silent routing failure in Next.js 16 App Router when handling post-login redirects with Clerk. 

This is my first open-source contribution—excited to help out! Let me know if any formatting changes are needed.

### 🔎 Previews:

- `/docs/guides/development/troubleshooting/overview`

**The Issue:**
When a user successfully authenticates and Clerk pushes them to a route that immediately executes a server-side `redirect()`, Next.js 16 silently aborts the React transition. The `<ClerkProvider>` global state update clashes with a blocking transition, causing the client router to permanently freeze on "Rendering..." with no console errors.

**The Fix:**
Documented that developers must introduce a Suspense boundary (via a `loading.tsx` file) at the redirect origin. This makes the transition concurrent, resolving the deadlock and ensuring smooth handoffs to protected database checks.

### Deadline

- No rush.

### Other resources

- N/A

### DIFF
--- a/docs/guides/development/troubleshooting/overview.mdx
+++ b/docs/guides/development/troubleshooting/overview.mdx
@@ -16,3 +16,26 @@ Want to connect with other developers? Join the [Discord community](https://cler
 ## Support
 
 Can't find what you need? [Contact support](/contact/support) to receive answers to your questions and learn more about Clerk.
+
+## Next.js 16: Client Router Hangs on Sign-In Redirect
+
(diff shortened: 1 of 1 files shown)


manovotny:
@kishan59 you've turned up an underlying SDK bug rather than a docs gap — Clerk's post-auth navigation into a route that server-redirects hangs the router, whether or not that route calls `auth()`. Filed as clerk/javascript#9405. Worth knowing for your own app: the `loading.tsx` workaround didn't work for us when I tested it.

So I'm going to close this in favor of fixing it in the SDK. Appreciate the detail in the write-up — it's what made this reproducible.

PR #3537, opened 2026-09-14
Remove leftover old-version install strings, fix Next.js minimum

eatmorespinach:
A few docs pages still have copyable install strings pinned to `@clerk/nextjs` v4/v5/v6. Readers can paste them, and since these pages are in `llms-full.txt`, LLMs read them too — so they end up seeding new apps on old majors. This clears them out and fixes one version claim that was just wrong.

1. **Core 3 upgrade guide, prep step** — `npm install @clerk/nextjs@6` was a copyable install line on a page people hit mid-upgrade. Reworded to prose so there's nothing to paste.

2. **Core 3 upgrade guide, Next.js minimum** — the page said 15.2.3 in four places, but `peerDependencies.next` is `^15.2.8 || ^15.3.8 || ^15.4.10 || ^15.5.9 || ^15.6.0-0 || ^16.0.10 || ^16.1.0-0` (same across 7.0.0 → 7.9.2). So we were telling people a version that fails install. Now 15.2.8.

   Deliberately not phrased as `next@>=15.2.8` — that's wrong too, since 15.3.0 is above 15.2.8 but outside the range. Each line has its own patch floor, so the prose lists them.

3. **`<Waitlist />` version notes** — "available in `@clerk/nextjs@6.2.0` and above" removed from `restricting-access.mdx`, `waitlist.mdx`, and the component reference. Every supported version has it, so the note just advertises an old version number. Two were wrapped in `<If sdk="nextjs">`; the empty wrappers went too.

4. **Reverification callout** — the partial was nothing but a Core 2 minimum-version list (`@clerk/nextjs@6.12.7`, `@clerk/react@5.25.1`, `@clerk/clerk-js@5.57.1`, `@clerk/clerk-sdk-ruby@3.3.0`), and every supported version now clears all four. Deleted the partial and both `<Include>` sites rather than leave an empty warning.

5. **Cognito migration guide** — `npm i -E @clerk/backend@1.4.3` → `npm i @clerk/backend`. The pin was installing a package two majors behind. The script only uses `createClerkClient` and `users.createUser({ emailAddress, passwordDigest, passwordHasher })`, both unchanged in v3, so it still works unpinned.

6. **`usage-with-older-versions.mdx`** — installs `@clerk/nextjs@^4.0.0` with no warning above the fold about who it's for. Added one: Next.js 12 and older only, new apps go to the quickstart.

**Not here:** getting the Core 2 upgrade guides out of `llms-full.txt`. There's no per-page exclusion mechanism today (`listOutputDocsFiles` takes every titled page), and the alternative moves live URLs — split to GROW-1238 so it doesn't hold up these copy edits. The blog-banner item is site chrome, tracked separately.

### Other resources

- GROW-1222 — https://linear.app/clerk/issue/GROW-1222
- GROW-1238 (the split-out Core 2 / llms-full work) — https://linear.app/clerk/issue/GROW-1238
- Parent: GROW-1215 — https://linear.app/clerk/issue/GROW-1215

Verified locally: `pnpm run build:tsx` clean (879 docs, no warnings), `pnpm run lint` green across all 15 checks.


eatmorespinach:
This branch and PR were migrated to clerk/clerk. Migration branches/PRs created from this PR (most recent last):

- `drewsing/grow-1222-docs-remove-leftover-old-version-install-strings-docs-migration` → https://github.com/clerk/clerk/pull/3379

